PR #15, opened 2016-04-01
Fix compatibility with latest thriftrw-go

prashantv:
Latest thriftrw-go uses a list for compile.FieldGroup instead of a map


abhinav:
:+1:

PR #17, opened 2016-04-04
Add support for "--health" as an alias for the Meta::health method

prashantv:
@abhinav @akshayjshah

### DIFF
--- a/meta_test.go
+++ b/meta_test.go
@@ -1,0 +1,40 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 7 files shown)


meta_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"io/ioutil"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetMetaFile(t *testing.T) {
+	f, err := getMetaFile()
+	require.NoError(t, err, "getMetaFile failed")
```
abhinav: You probably want to delete the file afterwards

options.go
```diff
@@ -31,10 +31,11 @@ type Options struct {
 
 // RequestOptions are request related options
 type RequestOptions struct {
-	ThriftFile  string `short:"t" long:"thrift" description:"Path of the .thrift file" required:"true"`
+	ThriftFile  string `short:"t" long:"thrift" description:"Path of the .thrift file"`
 	MethodName  string `short:"m" long:"method" description:"The full Thrift method name (Svc::Method) to invoke"`
 	RequestJSON string `short:"r" long:"request" description:"The request body, in JSON format"`
 	RequestFile string `short:"f" long:"file" description:"Path of a file containing the request body in JSON"`
+	Health      bool   `short:"h" long:"health" description:"Hit the health endpoint, Meta::health"`
```
abhinav: Will `-h` not conflict with the help flag? In fact most people will expect `-h` to be help.

abhinav:
LGTM besides minor things

meta_test.go
prashantv: :+1:

options.go
prashantv: Good point, I'll remove `-h` and only leave the long option for now.

PR #18, opened 2016-04-04
Fix yab not finding inherited methods on services

prashantv:
We create a list of all functions in the inheritance chain rather than using recursion since we want the error message to contain all the available functions.

Before:
```
$ ./yab -t $MOE_IDL -p localhost:9787 moe Stooges::unknown
Failed while parsing input: could not find method "unknown" in "Stooges", available methods:
        nyuck
```

After:
```
$ ./yab -t $MOE_IDL -p localhost:9787 moe Stooges::unknown
Failed while parsing input: could not find method "unknown" in "Stooges", available methods:
        errorClock
        nyuck
        ok20p
        ok40p
        ok60p
        ok80p
```

@abhinav @akshayjshah

### DIFF
--- a/request.go
+++ b/request.go
@@ -84,11 +84,23 @@ func findService(parsed *compile.Module, svcName string) (*compile.ServiceSpec, 
 }
 
 func findMethod(service *compile.ServiceSpec, methodName string) (*compile.FunctionSpec, error) {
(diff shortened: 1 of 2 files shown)


abhinav:
:+1:

akshayjshah:
:shipit:

PR #19, opened 2016-04-05
Allow meta endpoints to be targeted without writing out a temporary meta.thrift

prashantv:
Use `Compile` option added in https://github.com/thriftrw/thriftrw-go/pull/108 to avoid creating a temporary `meta.thrift` when making requests to the health endpoint.


abhinav:
:+1:

PR #21, opened 2016-04-12
Add Travis + Coveralls support

prashantv:
@abhinav @akshayjshah 

I modified the `cover.sh` from `thriftrw-go` -- I added support for ignoring vendor directories.


blampe:
Eww...

abhinav:
the only scripting language you need is bash

:+1:

PR #22, opened 2016-04-13
Ignore case and underscores when look for struct field matches

prashantv:
Since we're essentially making Thrift case-insensitive, we can
do fuzzy matching to find the right field. This makes it easier
to do things like copy the request struct from Go and use it in
`yab` even though the Go fields use title case.

@abhinav @akshayjshah

### DIFF
--- a/thrift/fields.go
+++ b/thrift/fields.go
@@ -29,35 +29,53 @@ import (
 	"github.com/thriftrw/thriftrw-go/wire"
 )
 
-func fieldMap(fields compile.FieldGroup) map[string]*compile.FieldSpec {
-	m := make(map[string]*compile.FieldSpec)
+// fieldMap returns maps from string to the field spec.
+// The first map is an exact map, which uses the name as specified in the Thrift file.
+// The second map is a fuzzy map, which will ignore case, and ignore
+// any non-alphanumberic characters.
+func fieldMap(fields compile.FieldGroup) (exact, fuzzy map[string]*compile.FieldSpec) {
+	exact = make(map[string]*compile.FieldSpec)
+	fuzzy = make(map[string]*compile.FieldSpec)
 	for _, f := range fields {
-		m[f.ThriftName()] = f
+		exact[f.ThriftName()] = f
+		fuzzy[fuzz(f.ThriftName())] = f
 	}
-	return m
+
+	// If there's any field clashes in the fuzzy map, skip fuzzy matching.
+	if len(exact) != len(fuzzy) {
+		fuzzy = nil
+	}
+
+	return exact, fuzzy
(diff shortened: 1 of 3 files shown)


thrift/fields.go
```diff
@@ -29,35 +29,53 @@ import (
...
 	)
 
 	for k, v := range request {
-		_, ok := fields[k]
+		_, ok := exactFields[k]
 		if !ok {
-			err.addNotFound(k)
-			continue
+			// Try a fuzzy match
+			field, ok := fuzzyFields[fuzz(k)]
```
abhinav: You probably need to handle the `fuzzyFields == nil` case.

thrift/fields.go
```diff
@@ -103,6 +121,25 @@ func fieldsMapToValue(fields map[string]*compile.FieldSpec, userFields map[strin
...
+// It lowercases the string, and removes non-alphanumeric characters.
+func fuzz(fieldName string) string {
+	newField := make([]byte, 0, len(fieldName))
+	for i := 0; i < len(fieldName); i++ {
+		c := fieldName[i]
+		switch {
+		case c >= 'A' && c <= 'Z':
+			newField = append(newField, c+upperToLower)
+		case c >= 'a' && c <= 'z', c >= '0' && c <= '9':
+			newField = append(newField, c)
```
abhinav: Why not just use `unicode.{IsLetter,IsDigit,ToLower}` instead?

thrift/parse_test.go
```diff
@@ -64,6 +64,7 @@ func TestParseRequest(t *testing.T) {
     struct S {
       1: required string f1,
       2: optional NS ns,
+			3: optional NS ns_,
```
abhinav: Indentation mismatch

prashantv: 👍

thrift/fields.go
prashantv: `nil` maps are actually valid empty maps (for reading). Since these are immutable maps, it works fine:
http://play.golang.org/p/0L0CzDAk_Y

thrift/fields.go
prashantv: Couple of reasons:
 - Since I'm only dealing with a very small set of known characters (e.g. based on the Thrift spec), we only really care about ASCII.
 - Unicode deals with runes which can't be easily appended to a `[]byte` without converting to a string. In general, there's more overhead

abhinav: :+1:

thrift/fields.go
abhinav: oh true

abhinav:
:+1:

PR #23, opened 2016-04-13
Improved binary support: allow base64 or files to be used for binary/string input

prashantv:
Previously, binary only supported strings,
```json
"binaryField": "value"
```

Now, we allow for the object to be specifies as a list of numbers or characters:
```json
"binaryField": [1,2,3]
```
```json
"binaryField": ["a", "b", "c"]
```

We also allow using objects to specify custom binary deserialization, such as base64 or reading a file:
```json
"binaryField": {
  "base64": "YWI="
}
```
```json
"binaryField": {
  "file": "/my/file"
}
```

@abhinav @akshayjshah

### DIFF
--- a/thrift/types.go
+++ b/thrift/types.go
@@ -21,11 +21,15 @@
 package thrift
 
 import (
+	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
 )
+
+var errBinaryObjectOptions = errors.New("object input for binary/string must have one of the following keys: base64")
 
 func parseBoolNumber(v json.Number) (bool, error) {
 	i64, err := v.Int64()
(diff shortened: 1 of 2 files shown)


thrift/types.go
```diff
@@ -21,12 +21,17 @@
...
+	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io/ioutil"
 	"strconv"
 	"strings"
 )
 
+var errBinaryObjectOptions = errors.New("object input for binary/string must have one of the following keys: base64")
```
abhinav: missing "file"

prashantv: 👍 fixed

thrift/types.go
```diff
@@ -100,15 +106,67 @@ func parseDouble(value interface{}) (float64, error) {
...
+			vInt, err := strconv.ParseInt(v.String(), 10, 8)
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse list of bytes: %v", err)
+			}
+			bs = append(bs, byte(vInt))
+		case string:
+			if len(v) != 1 {
+				return nil, fmt.Errorf("not a valid character in list of characters: %q", v)
+			}
+			bs = append(bs, v[0])
```
abhinav: If you remove this restriction and append `append(bs, v...)`, the following should just work, right?

```
{"somefield": [1, "a", 5, "foo"]}
```

Any reason to not support that?

prashantv: Good point, no reason not to. Should we support 0 length strings as well then?

abhinav: Illogical use case but supporting it doesn't break anything, so sure.

prashantv: Agreed, updated.

abhinav:
:+1:

PR #24, opened 2016-04-13
Allow struct fields to be specified by field ID

prashantv:
This makes consuming output from tools such as tcap.

tcap output looks like:
```
arg3 as thrift
{ '1':
   { '1': 'string',
     '2': 'go' } }
```

Unfortunately json doesn't allow `'` to be used, but once you replace the single quotes with double quotes, it can be passed to `yab` and makes a successful request.

@abhinav

### DIFF
--- a/thrift/fields.go
+++ b/thrift/fields.go
@@ -22,6 +22,7 @@ package thrift
 
 import (
 	"sort"
+	"strconv"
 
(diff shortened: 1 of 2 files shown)


abhinav:
:+1: 

nice

PR #25, opened 2016-04-15
Add support for request headers and output response headers

prashantv:
Note: we have a bit of a hack for raw headers, where we use a `_raw_` key.
This is because I want the internal structure for Headers to be
map[string]string since that is the most common use, but want to support
sending the headers as is for the raw encoding over TChannel.

@abhinav

### DIFF
--- a/transport/tchannel.go
+++ b/transport/tchannel.go
@@ -21,15 +21,24 @@
 package transport
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
+	"io"
+	"io/ioutil"
 
 	"github.com/uber/tchannel-go"
 	"github.com/uber/tchannel-go/thrift"
 	"golang.org/x/net/context"
 )
 
+// rawHeadersKey is a hack to specify a raw payload via the headers map.
+// If this key is used, then the headers are sent as is.
+const rawHeadersKey = "_raw_"
+
 type tchan struct {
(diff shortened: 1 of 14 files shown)


transport/tchannel.go
```diff
@@ -118,15 +152,28 @@ func (t *tchan) readResponse(call *tchannel.OutboundCall) (*Response, error) {
...
+func (t *tchan) writeArgs(call *tchannel.OutboundCall, r *Request) error {
 	if err := writeHelper(call.Arg2Writer, func(writer tchannel.ArgWriter) error {
-		return thrift.WriteHeaders(writer, headers)
+		switch t.callOptions.Format {
+		case tchannel.JSON:
+			encoder := json.NewEncoder(writer)
+			return encoder.Encode(r.Headers)
+		case tchannel.Raw:
+			if v, ok := r.Headers[rawHeadersKey]; ok {
+				_, err := io.WriteString(writer, v)
```
abhinav: Don't you want to delete the `_raw_` header?

abhinav:
Hacky but :ok:

transport/tchannel.go
prashantv: Requests is immutable, and if the request was used with a benchmark, the same raw header has to be sent multiple times, so we have to keep it

PR #26, opened 2016-04-15
Add customizable timeout when making requests

prashantv:
Allow the user to specify a timeout using the standard duration format (e.g. `1s` or `100ms`) while also allowing units to be specified as a number of milliseconds (for compatibility with `tcurl`)

@abhinav


abhinav:
:shipit:

PR #27, opened 2016-04-15
Setting timeout causes data races in benchmarks

prashantv:
Multiple goroutines may call makeRequest on the same Request object
as it's supposed to be immutable. Have all callers ensure the Request
has a timeout before passing it in to makeRequest.

@abhinav


abhinav:
:+1:

PR #28, opened 2016-04-15
Add support for overriding Caller and setting other TChannel transport options

prashantv:
go-flags allows setting maps via flags, the usage for setting multiple flags would be:
```
yab -P hosts.json moe --health --topt as:raw --topt sk:shard-key --topt rd:router
``` 

We also allow overriding the caller using `--caller`, but this is disabled when making benchmarks (for safety).

For TChannel, we add a more descriptive process name, so even if the user uses a custom caller, we can identify where the request came.

@abhinav

### DIFF
--- a/options.go
+++ b/options.go
@@ -47,9 +47,10 @@ type RequestOptions struct {
 
 // TransportOptions are transport related options.
 type TransportOptions struct {
-	ServiceName  string   `short:"s" long:"service" description:"The TChannel/Hyperbahn service name"`
-	HostPorts    []string `short:"p" long:"peer" description:"The host:port of the service to call"`
-	HostPortFile string   `short:"P" long:"peer-list" description:"Path of a JSON file containing a list of host:ports"`
+	ServiceName      string            `short:"s" long:"service" description:"The TChannel/Hyperbahn service name"`
+	HostPorts        []string          `short:"p" long:"peer" description:"The host:port of the service to call"`
+	HostPortFile     string            `short:"P" long:"peer-list" description:"Path of a JSON file containing a list of host:ports"`
+	TransportOptions map[string]string `long:"topt" description:"Custom options for the specific transport being used"`
 }
 
 // BenchmarkOptions are benchmark-specific options
--- a/options.go
+++ b/options.go
@@ -50,7 +50,11 @@ type TransportOptions struct {
 	ServiceName      string            `short:"s" long:"service" description:"The TChannel/Hyperbahn service name"`
 	HostPorts        []string          `short:"p" long:"peer" description:"The host:port of the service to call"`
 	HostPortFile     string            `short:"P" long:"peer-list" description:"Path of a JSON file containing a list of host:ports"`
+	CallerOverride   string            `long:"caller" description:"Caller will override the default caller name (which is yab-$USER)."`
 	TransportOptions map[string]string `long:"topt" description:"Custom options for the specific transport being used"`
+
+	// benchmarking is a private flag set when a transport is required for benchmarking.
+	benchmarking bool
 }
 
 // BenchmarkOptions are benchmark-specific options
--- a/transport/tchannel.go
+++ b/transport/tchannel.go
@@ -26,6 +26,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 
 	"github.com/uber/tchannel-go"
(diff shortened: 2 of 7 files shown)


options.go
```diff
@@ -47,9 +47,14 @@ type RequestOptions struct {
...
 // TransportOptions are transport related options.
 type TransportOptions struct {
-	ServiceName  string   `short:"s" long:"service" description:"The TChannel/Hyperbahn service name"`
-	HostPorts    []string `short:"p" long:"peer" description:"The host:port of the service to call"`
-	HostPortFile string   `short:"P" long:"peer-list" description:"Path of a JSON file containing a list of host:ports"`
+	ServiceName      string            `short:"s" long:"service" description:"The TChannel/Hyperbahn service name"`
+	HostPorts        []string          `short:"p" long:"peer" description:"The host:port of the service to call"`
+	HostPortFile     string            `short:"P" long:"peer-list" description:"Path of a JSON file containing a list of host:ports"`
+	CallerOverride   string            `long:"caller" description:"Caller will override the default caller name (which is yab-$USER)."`
+	TransportOptions map[string]string `long:"topt" description:"Custom options for the specific transport being used"`
```
abhinav: Should we add an example in the help for transport options? It's not quite obvious how to use it from the help alone. Or is the plan to discourage usage by obfuscation?

options.go
```diff
@@ -47,9 +47,14 @@ type RequestOptions struct {
 
 // TransportOptions are transport related options.
 type TransportOptions struct {
-	ServiceName  string   `short:"s" long:"service" description:"The TChannel/Hyperbahn service name"`
-	HostPorts    []string `short:"p" long:"peer" description:"The host:port of the service to call"`
-	HostPortFile string   `short:"P" long:"peer-list" description:"Path of a JSON file containing a list of host:ports"`
+	ServiceName      string            `short:"s" long:"service" description:"The TChannel/Hyperbahn service name"`
+	HostPorts        []string          `short:"p" long:"peer" description:"The host:port of the service to call"`
+	HostPortFile     string            `short:"P" long:"peer-list" description:"Path of a JSON file containing a list of host:ports"`
+	CallerOverride   string            `long:"caller" description:"Caller will override the default caller name (which is yab-$USER)."`
```
abhinav: Mention that this is ignored if benchmarking? Or mention in benchmark options somewhere that "using this disables --caller".

abhinav: nvm ignore. you have an error

transport/tchannel.go
```diff
@@ -66,10 +71,22 @@ func TChannel(opts TChannelOptions) (Transport, error) {
 		level = *opts.LogLevel
 	}
 
+	callerName := opts.SourceService
+	if cn, ok := opts.TransportOpts["cn"]; ok {
```
abhinav: should we error on unknown transport options?

transport/tchannel.go
```diff
@@ -182,6 +200,18 @@ func (t *tchan) writeArgs(call *tchannel.OutboundCall, r *Request) error {
...
+func applyTChanOptions(callOpts *tchannel.CallOptions, opts map[string]string) {
+	if format, ok := opts["as"]; ok {
+		callOpts.Format = tchannel.Format(format)
+	}
+	if rd, ok := opts["rd"]; ok {
+		callOpts.RoutingDelegate = rd
+	}
+	if sk, ok := opts["sk"]; ok {
+		callOpts.ShardKey = sk
+	}
```
abhinav: should we error on unknown transport options?

options.go
prashantv: Discourage, since you can skip some of the protections (like no overriding caller when benchmarking) using these. I want the ability to override these low level params but don't necessarily want them used in the common cases.

transport/tchannel.go
prashantv: I would like to, but I'm not sure if it's worth the code. I want to:
 - not mutate the options (since you can create multiple transports with the same options)
 - not have to repeat myself for what options we've checked

I'm also wondering whether unknown options should be propagated as custom transport headers, not sure yet.

options.go
blampe: +1 don't encourage these.

transport/tchannel.go
abhinav: Cool. We can add an error here later if necessary.

abhinav:
LGTM

PR #32, opened 2016-05-03
Use YAML for parsing input

prashantv:
YAML is backwards compatible with JSON, but makes it easier to deal with things like maps with integer keys.

This change should be backwards compatible with all valid inputs, and the output is still JSON. It just allows more flexibility when parsing input.

@abhinav @akshayjshah


blampe:
Nice! :shipit:

abhinav:
👍

PR #33, opened 2016-05-05
Refactor encoding code into separate package

prashantv:
The main package is huge -- refactoring all encoding related code into a separate `encoding` package.

There are no changes to the logic, only refactoring + more unit tests.

@abhinav


blampe:
Go for it.

breerly:
Nice improvement 👍

PR #34, opened 2016-05-23
Add aliases -2 and -3 for headers/body to match tcurl

prashantv:
Instead of adding another `string` flag, I created a special `stringAlias` type. This ensures that we prefer the last argument specified, rather than preferring `-r` over `-3` or the other way around.

Added test to ensure that the last option specified is preferred, even when the alias is used.

### DIFF
--- a/main.go
+++ b/main.go
@@ -60,35 +60,36 @@ func main() {
 	parseAndRun(consoleOutput{os.Stdout})
 }
 
-// parseAndRun is like main, but uses the given output.
-func parseAndRun(out output) {
-	var opts Options
-	parser := flags.NewParser(&opts, flags.HelpFlag|flags.PassDoubleDash)
+var errExit = errors.New("sentinel error used to exit cleanly")
+
+func getOptions(args []string, out output) (*Options, error) {
+	opts := newOptions()
+	parser := flags.NewParser(opts, flags.HelpFlag|flags.PassDoubleDash)
 	parser.Usage = "[<service> <method> <body>] [OPTIONS]"
 	findGroup(parser, "transport").ShortDescription = "Transport Options"
 	findGroup(parser, "request").ShortDescription = "Request Options"
 	findGroup(parser, "benchmark").ShortDescription = "Benchmark Options"
 
 	// If there are no arguments specified, write the help.
-	if len(os.Args) <= 1 {
+	if len(args) == 0 {
 		parser.WriteHelp(out)
-		return
+		return opts, errExit
 	}
 
-	remaining, err := parser.Parse()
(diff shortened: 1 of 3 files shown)


kriskowal:
So, it turns out nobody likes these flags. What’s the rationale? I suspect that this might be a hobgoblin.

prashantv:
The `-2` and `-3` will be "hidden" flags -- they'll work but they won't show up in documentation.

A few engineers were using `tcurl` last week and used `-3` for specifying the body. This made it harder to use the same command in yab, since they had to change `-3` to `-r` (or `--request`). Since I doubt we'll want to use `-2` and `-3` for anything else, thought it'd be nice to support them without recommending them.

kriskowal:
I hope you’re entertaining the idea of a SHON flag parser :P

prashantv:
I'm hoping you'll be implementing that!

main_test.go
```diff
@@ -238,3 +239,32 @@ func TestVersion(t *testing.T) {
...
+		{"-3", "7"},
+		{"--arg3", "8"},
+	}
+
+	var flags []string
+	for _, tt := range tests {
+		flags = append(flags, tt.flagName, tt.flagValue)
+
+		opts, err := getOptions(flags, out)
+		require.NoError(t, err, "getOptions(%v) failed", "lol")
```
abhinav: "lol"?

main.go
```diff
@@ -103,7 +104,19 @@ func parseAndRun(out output) {
...
+func parseAndRun(out output) {
+	opts, err := getOptions(os.Args[1:], out)
+	if err != nil {
+		if err == errExit {
+			return
+		}
+		out.Fatalf("Failed to parse options: %v", err)
+	}
+	runWithOptions(*opts, out)
 	return
```
abhinav: do you need an explicit return here? it's the end of the function.

options.go
```diff
@@ -76,6 +81,13 @@ type BenchmarkOptions struct {
 	StatsdHostPort string `long:"statsd" description:"Optional host:port of a StatsD server to report metrics"`
 }
 
+func newOptions() *Options {
+	var opts Options
+	opts.ROpts.Arg3.dest = &opts.ROpts.RequestJSON
+	opts.ROpts.Arg2.dest = &opts.ROpts.HeadersJSON
```
abhinav: This is clever

main.go
prashantv: removed

main_test.go
prashantv: lol at myself. fixed

prashantv:
Updated

abhinav:
👍

PR #36, opened 2016-05-25
Use in-memory filesystem for parsing test thrift files

prashantv:
Test refactoring with no functional changes.

cc @abhinav


breerly:
dummy FS, cool :shipit:

PR #37, opened 2016-05-25
thrift: Improve handling of enums

prashantv:
When an enum value is received over the wire, it is formatted using the
name. If a name is not found, it is formatted as EnumType(%d)

Enums can be specified as:
- int32
- Case-insentivie name
- EnumType(%d)

The last format is so that formatted unknown enums on the result path
can be used as input.

### DIFF
--- a/thrift/from_wire.go
+++ b/thrift/from_wire.go
@@ -123,6 +123,15 @@ func valueFromWireMap(spec *compile.MapSpec, w wire.Map) (map[string]interface{}
 	return result, nil
 }
 
+func mapEnumValueToName(enumSpec *compile.EnumSpec, result int32) interface{} {
+	for _, item := range enumSpec.Items {
+		if item.Value == result {
+			return item.Name
+		}
+	}
+	return fmt.Sprintf("%v(%v)", enumSpec.Name, result)
+}
+
 // valueFromWire converts the wire.Value to the specific type it represents.
 func valueFromWire(spec compile.TypeSpec, w wire.Value) (interface{}, error) {
 	if spec.TypeCode() != w.Type() {
@@ -140,7 +149,11 @@ func valueFromWire(spec compile.TypeSpec, w wire.Value) (interface{}, error) {
 	case wire.TI16:
 		result = w.GetI16()
 	case wire.TI32:
-		result = w.GetI32()
+		if enumSpec, ok := spec.(*compile.EnumSpec); ok {
+			result = mapEnumValueToName(enumSpec, w.GetI32())
+		} else {
+			result = w.GetI32()
+		}
 	case wire.TI64:
 		result = w.GetI64()
 	case wire.TDouble:
--- a/thrift/from_wire_test.go
+++ b/thrift/from_wire_test.go
@@ -230,6 +230,45 @@ func TestValueFromWireSuccess(t *testing.T) {
 				"s": "foo",
 			},
 		},
+		{
+			// Enum with recognized value.
+			w: wire.NewValueI32(1),
+			spec: &compile.EnumSpec{
+				Name: "Op",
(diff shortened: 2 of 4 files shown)


thrift/parse_test.go
```diff
@@ -437,6 +475,20 @@ func TestParseRequest(t *testing.T) {
...
+			// Unknown enum
+			request: map[string]interface{}{
+				"op": "Divide",
+			},
+			errMsg: `unrecognized enum "Divide"`,
+		},
+		{
+			// Unknown enum
+			request: map[string]interface{}{
+				"op": "Op(NaN)",
```
breerly: @kriskowal will be at home :)

breerly:
:shipit: 

(❤️  table test)

thrift/from_wire.go
```diff
@@ -123,6 +123,15 @@ func valueFromWireMap(spec *compile.MapSpec, w wire.Map) (map[string]interface{}
 	return result, nil
 }
 
+func mapEnumValueToName(enumSpec *compile.EnumSpec, result int32) interface{} {
+	for _, item := range enumSpec.Items {
```
blampe: Should we just setup a static mapping so we don't need to iterate like this every time?

prashantv: I think the same issue happens in code generation in `thriftrw` as well -- @abhinav what do you think about setting up a mapping internally in `compile.EnumSpec` and having a method to return the `EnumItem` given a value?

abhinav: We can have a reverse mapping but it will have undefined behavior when multiple enum items have the same enum value.

abhinav: Created https://github.com/thriftrw/thriftrw-go/issues/134

thrift/from_wire_test.go
```diff
@@ -230,6 +230,27 @@ func TestValueFromWireSuccess(t *testing.T) {
...
+			// Enum with recognized value.
+			w: wire.NewValueI32(1),
+			spec: &compile.EnumSpec{
+				Name: "Op",
+				Items: []compile.EnumItem{{
+					Name:  "Add",
+					Value: 1,
+				}},
+			},
+			v: "Add",
```
abhinav: Can you also add a test for an enum with multiple items using the same value?

abhinav: Also for later: You probably want to link the spec with an empty scope before you use it. This is not actionable right now because this isn't doable without some amount of boilerplate (see https://github.com/thriftrw/thriftrw-go/issues/135).

thrift/parse_test.go
```diff
@@ -368,6 +374,38 @@ func TestParseRequest(t *testing.T) {
...
+			request: map[string]interface{}{
+				"op": "Add",
+			},
+			want: []wire.Field{
+				{ID: 18, Value: wire.NewValueI32(1)},
+			},
+		},
+		{
+			request: map[string]interface{}{
+				"op": "Multiply",
```
abhinav: cool

thrift/to_wire.go
```diff
@@ -26,6 +26,10 @@ import (
 
 	"gopkg.in/yaml.v2"
 
+	"strings"
+
+	"strconv"
```
abhinav: import group ordering?

abhinav:
LGTM with minor comments.

thrift/from_wire_test.go
prashantv: Added test for enum with multiple items that have the same value. (we currently just pick the first value).

thrift/to_wire.go
prashantv: fixed, VS Code's "Add Import" apparently just doesn't care about import ordering or grouping

abhinav:
👍

PR #38, opened 2016-05-26
thrift: Resolve typedefs before wire conversion

abhinav:
CC @prashantv

### DIFF
--- a/thrift/from_wire.go
+++ b/thrift/from_wire.go
@@ -141,6 +141,7 @@ func valueFromWire(spec compile.TypeSpec, w wire.Value) (interface{}, error) {
 	var result interface{}
 	var err error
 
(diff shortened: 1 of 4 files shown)


thrift/from_wire.go
```diff
@@ -182,3 +182,14 @@ func valueFromWire(spec compile.TypeSpec, w wire.Value) (interface{}, error) {
 	}
 	return result, nil
 }
+
+func isStringSpec(s compile.TypeSpec) bool {
```
prashantv: Filed https://github.com/thriftrw/thriftrw-go/issues/138 to make this easier in future

abhinav:
Updated to handle typedefs in both directions for more cases

prashantv:
lgtm

PR #41, opened 2016-05-27
Use NonStrict parsing in yab

prashantv:
In future, we might want to parse in strict mode, and if it fails, do non-strict mode, and then print the errors as warnings.


abhinav:
👍

PR #42, opened 2016-05-27
Use compile.RootTypeSpec in latest thriftrw-go

prashantv:


prashantv:
This won't work until we link with an empty scope (https://github.com/thriftrw/thriftrw-go/issues/135)

abhinav:
👍

Issue #46, opened 2016-06-01
Should all HTTP headers be displayed in the output


prashantv:
When making HTTP requests, `yab` currently shows all response HTTP headers in the "headers" section. Should we be stripping common HTTP headers away (unless explicitly left on by a transport option), or should headers always be hidden?

Current output:
```json
{
  "body": {
    "result": 5
  },
  "headers": {
    "Content-Length": "23",
    "Content-Type": "application/x-thrift",
    "Date": "Wed, 01 Jun 2016 15:55:10 GMT"
  }
}
```

blampe:
Stripping some headers would be surprising, maybe either show all or none?

breerly:
I would show all because:

* YARPC defaults to `RPC-` prefix
* Context headers defaults to `Context-`, but one-line inits might update them, eg `YOOBERCTX-`

prashantv:
Alright, let's leave it as showing all the headers.

PR #47, opened 2016-06-01
Add support for making enveloped HTTP Thrift calls

prashantv:
Assume envelopes are default, and only disable them when making Thrift+TChannel calls.

Added an integration test that verifies that the same Thrift handler registered under HTTP and TChannel works consistently with yab.

This is mostly to make HTTP Thrift functional, filed items for polishing up for future (e.g. #46)


encoding/encoding.go
```diff
@@ -93,7 +93,7 @@ func (e Encoding) GetHealth() (Serializer, error) {
 	switch e {
 	case UnspecifiedEncoding, Thrift:
 		method, spec := getHealthSpec()
-		return thriftSerializer{method, spec}, nil
+		return &thriftSerializer{method, spec, defaultOpts}, nil
```
abhinav: Why make this a reference? `thriftSerializer` as a `Serializer` is still immutable since all methods are on the object, not the pointer.

encoding/thrift_request.go
```diff
@@ -109,7 +112,13 @@ func findService(parsed *compile.Module, svcName string) (*compile.ServiceSpec,
 }
 
 func (e thriftSerializer) CheckSuccess(res *transport.Response) error {
-	return thrift.CheckSuccess(e.spec, res.Body)
+	return thrift.CheckSuccess(e.spec, res.Body, e.opts)
+}
+
+func (e thriftSerializer) WithoutEnvelopes() Serializer {
+	withoutEnvelopes := e
+	withoutEnvelopes.opts.UseEnvelopes = false
```
abhinav: You can just modify and return `e`, can't you? It and `opts` are both copies.

encoding/thrift_request_test.go
```diff
@@ -243,3 +243,20 @@ func TestFindMethod(t *testing.T) {
...
+func TestWithoutEnvelopes(t *testing.T) {
+	serializer, err := NewThrift(validThrift, "Simple::foo")
+	require.NoError(t, err, "Failed to create serializer")
+
+	enveloped, err := serializer.Request([]byte("{}"))
+	require.NoError(t, err, "serialize enveloped failed")
+
+	serializer = serializer.(thriftSerializer).WithoutEnvelopes()
+	unenveloped, err := serializer.Request([]byte("{}"))
+	require.NoError(t, err, "serialize unenveloped failed")
```
abhinav: Is it worth verifying the serialized contents?

Unenveloped empty struct is just a single byte: `0x00`. Enveloped will be,

```
0x80, 0x01, 0x00, 0x01, // version | type = 1 | call
0x00, 0x00, 0x00, 0x03, 'f', 'o', 'o', // "foo"
0x00, 0x00, 0x00, 0x00, // whatever seqID you're using
0x00, // empty struct
```

abhinav:
LGTM

encoding/thrift_request.go
```diff
@@ -33,9 +33,12 @@ import (
 	"github.com/thriftrw/thriftrw-go/compile"
 )
 
+var defaultOpts = thrift.Options{UseEnvelopes: true}
```
breerly: nice

main.go
```diff
@@ -198,6 +201,19 @@ func runWithOptions(opts Options, out output) {
...
 
+type noEnveloper interface {
+	WithoutEnvelopes() encoding.Serializer
+}
+
+// withTransportSerializer may modify the serializer for the transport used.
+// E.g. Thrift payloads are not enveloped when used with TChannel.
+func withTransportSerializer(p transport.Protocol, s encoding.Serializer) encoding.Serializer {
+	if p == transport.TChannel && s.Encoding() == encoding.Thrift {
+		s = s.(noEnveloper).WithoutEnvelopes()
```
breerly: Ok, so the owness is on TChannel to turn off envelopes, that makes a lot of sense 💯

breerly: This is likely how we'll do it in YARPC as well.

breerly:
lgtm

encoding/encoding.go
prashantv: 👍 forgot to revert change (i temporarily made `thriftSerializer` mutable)

encoding/thrift_request.go
prashantv: 👍 good point

encoding/thrift_request_test.go
prashantv: True, updated to test the bytes

main.go
prashantv: Yeah I figure if yab ever adds support for Thrift over raw sockets, it'll use envelopes too, so seems like envelopes are the expected way to use Thrift.

PR #49, opened 2016-06-03
Improve man pages + flags documentation

prashantv:
Add rules to generate the man page + HTML version, which we will push to a `gh-pages` branch so that the Github page is updated.

Also updated some benchmarking flags to use hyphens instead of camelCase which is the standard for flags.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,14 @@ install_ci: install
 		go get golang.org/x/tools/cmd/cover
 
 
+.PHONY: update_man
+update_man:
+	go install .
+	yab --man-page > man/yab.1
+	nroff -man man/yab.1 | man2html -title yab > man/yab.html
+	[[ -d ../yab_ghpages ]] && cp man/yab.html ../yab_ghpages/man.html
+	@echo "Please update gh-pages"
+
 .PHONY: test_ci
 test_ci: install_ci build
 	./scripts/cover.sh $(shell go list $(PACKAGES))
--- a/doc.go
+++ b/doc.go
@@ -1,0 +1,107 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// yab is a benchmarking tool for TChannel and HTTP applications.
+// It's primarily intended for Thrift applications but supports other encodings like JSON and binary (raw).
+//
+// It can be used in a curl-like fashion when benchmarking features are disabled.
+//
+// For usage information, check out the man page:
+// http://yarpc.github.io/yab/man.html
+package main
+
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -26,6 +26,14 @@ install_ci: install
 		go get golang.org/x/tools/cmd/cover
 
 
+.PHONY: update_man
+update_man:
+	go install .
+	yab --man-page > man/yab.1
```
blampe: Do you need to `mkdir man`?

Makefile
```diff
@@ -26,6 +26,14 @@ install_ci: install
...
+	nroff man/yab.1 -man -Thtml > man/yab.html
+	[[ -d ../yab_ghpages ]] && cp man/yab.html ../yab_ghpages/man.html
```
blampe: Nice!

doc.go
```diff
@@ -0,0 +1,100 @@
...
+by passing --health.
+
+Thrift requests can be specified as JSON or YAML. For example, for a method
+defined as:
+  void addUser(1: string name, 2: i32 age)
+
+You can pass the request as JSON: {"name": "Prashant", age: 10}
+or as YAML:
+  name: Prashant
+	age: 10
```
abhinav: Indentation mismatch

Makefile
prashantv: This PR adds the `man` dir with the generated man pages so anyone running this command after this change will already have the directory

doc.go
```diff
@@ -0,0 +1,100 @@
...
+
+The request can be specified on the command line using -r or --request:
+  $ yab -p localhost:9787 kv KeyValue::Get -3 '{"key": "hello"}'
```
abhinav: You say `-r` and `--request` but you're using `-3`

doc.go
```diff
@@ -0,0 +1,100 @@
...
+
+Or it can be loaded from a file using -f or --file:
+  $ yab -p localhost:9787 kv KeyValue::Get -f req.yaml
```
abhinav: Also, `-m` was mentioned above and not used

doc.go
```diff
@@ -0,0 +1,100 @@
...
+// It can be used in a curl-like fashion when benchmarking features are disabled.
+//
+// For usage information, check out the man page:
+// http://yarpc.github.io/yab/man.html
+package main
+
+const _reqOptsDesc = `Configures the request data and the encoding.
+
+To make Thrift requests, specify a Thrift file and pass the Thrift
+service and procedure to the method argument (-m) as Service::Method.
```
abhinav: Prefer including the long form (`--method`) as well

doc.go
prashantv: 👍

doc.go
prashantv: wow, old habits die hard. Even when writing documentation I keep using `-3`. Changing to `-r`

doc.go
```diff
@@ -0,0 +1,100 @@
...
+or
+  $ yab -p tchannel://localhost:9787 [options]
+
+For HTTP endpoints, specify the URL as the peer,
+  $ yab -p http://localhost:8080/thrift [options]
+
+The Thrift encoded body will be POSTed to the specified URL.
+
+Multiple peers can be specified using a peer list using -P or --peer-list.
+  $ yab -P hosts.json [options]
```
abhinav: What's the behavior when you specify multiple peers?

doc.go
```diff
@@ -0,0 +1,100 @@
...
+By default, yab will only make a single request. To enable benchmarking, you
+must specify the maximum duration for the benchmark by passing -d or --max-duration.
+
+yab will make requests till either the maximum requests (-n or --max-requests)
+or the maximum duration is reached.
+
+You can control the rate at which yab makes requests using the --rps flag.
+
+An example benchmark command might be:
+  yab -p localhost:9787 moe --health -n 100000 -d 10s --rps 1000
```
abhinav: I would use the long forms of all options in this example (and some of the other ones). It explains exactly what each parameter means.

Makefile
blampe: Ah of course, overlooked that.

doc.go
prashantv: 👍

doc.go
prashantv: Good point, will add section on the behaviour.

doc.go
prashantv: The line becomes really long which actually ends up obscuring the useful info. I think `-d` and `-n` are pretty readable vs adding a lot of noise in `--max-duration` and `--max-requests`. I'd expect almost everyone to use the short form.

doc.go
prashantv: Added example with `-m` and explained how positional args can be used too. Trying to reduce the amount of noise.

doc.go
abhinav: 👍

doc.go
```diff
@@ -0,0 +1,107 @@
...
+Thrift requests can be specified as JSON or YAML. For example, for a method
+defined as:
+  void addUser(1: string name, 2: i32 age)
+
+You can pass the request as JSON: {"name": "Prashant", age: 100}
+or as YAML:
+  name: Prashant
+  age: 100
+
+The request can be specified on the command line using -r or --request:
```
abhinav: nit:

> The request body can [..]

abhinav:
LGTM

doc.go
prashantv: fixed

PR #53, opened 2016-07-01
Add a hidden option for -H

blampe:
@prashantv

### DIFF
--- a/options.go
+++ b/options.go
@@ -66,7 +66,7 @@ type TransportOptions struct {
 	benchmarking bool
 
 	// Alias for tcurl compatibility.
-	H stringAlias `short:"H" hidden:"true"`
+	H stringAlias `short:"H" long:"hostlist" hidden:"true"`
 }
 
 // BenchmarkOptions are benchmark-specific options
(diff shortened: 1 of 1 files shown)


options.go
```diff
@@ -64,6 +64,9 @@ type TransportOptions struct {
 
 	// benchmarking is a private flag set when a transport is required for benchmarking.
 	benchmarking bool
+
+	// Alias for tcurl compatibility.
+	H stringAlias `short:"H" hidden:"true"`
```
prashantv: should we also add `hostfile` or whatever the long version was here?

options.go
```diff
@@ -64,6 +64,9 @@ type TransportOptions struct {
 
 	// benchmarking is a private flag set when a transport is required for benchmarking.
 	benchmarking bool
+
+	// Alias for tcurl compatibility.
```
abhinav: alias of what though?

prashantv:
lgtm

options.go
blampe: sure!

abhinav:
LGTM

PR #54, opened 2016-07-01
Add more tcurl aliases

prashantv:
Added hidden aliases for the request related options from:
https://github.com/uber/tcurl/blob/master/tcurl.usage


blampe:
:)

:shipit:

PR #58, opened 2016-07-01
Print benchmark results even if it's stopped early via ctrl-c

### LINKED ISSUE #56: If a benchmark is stopped early, it should print results so far

prashantv:
Move the run limiter logic into a separate package with a Stop method.

Fixes #56

### DIFF
--- a/benchmark.go
+++ b/benchmark.go
@@ -21,6 +21,8 @@
 package main
 
 import (
+	"os"
+	"os/signal"
 	"runtime"
 	"sync"
 	"time"
@@ -98,6 +100,7 @@ func runBenchmark(out output, allOpts Options, m benchmarkMethod) {
 	}
 
 	run := limiter.New(opts.MaxRequests, opts.RPS, opts.MaxDuration)
+	stopOnInterrupt(out, run)
 
 	start := time.Now()
 	for i, c := range connections {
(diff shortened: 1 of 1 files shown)


benchmark.go
```diff
@@ -163,3 +134,16 @@ func runBenchmark(out output, allOpts Options, m benchmarkMethod) {
 	out.Printf("Total requests:    %v\n", len(overall.latencies))
 	out.Printf("RPS:               %.2f\n", float64(len(overall.latencies))/total.Seconds())
 }
+
+// stopOnInterrupt sets up a signal that will trigger the run to stop.
+func stopOnInterrupt(out output, r *limiter.Run) {
```
breerly: nice

breerly:
* having trouble following why the rate limiter is in this PR
* don't see any tests issuing `os.Interrupt`

abhinav:
+1 on moving the rate limiter changes into a separate PR

prashantv:
No tests for `os.Interrupt` since sending a `SIGINT` stops the process. It's possible to test by spawning a new process, but I don't think it's worth the effort.

The rate limiter changes were the main thing required to Stop (adding the Stop method), but I wanted to add tests and move it out to clean up the code, moved the refactor into a separate PR though: #59 

Will rebase this change once that lands.

prashantv:
Rebased + updated, PTAL

abhinav:
LGTM

PR #59, opened 2016-07-01
Move limiter out to a separate package and add tests

prashantv:
Prefactor out the limiter into a separate package from #58

### DIFF
--- a/limiter/limiter.go
+++ b/limiter/limiter.go
@@ -1,0 +1,47 @@
+package limiter
+
+import (
+	"sync/atomic"
+	"time"
+
+	"github.com/yarpc/yab/ratelimit"
+)
+
+// Run represents a single run that is limited by either
+// a number of requests, or can be stopped halfway.
+type Run struct {
+	// TODO: Migrate to uber-go/atomic.
+	requestsLeft int64
+	limiter      ratelimit.Limiter
+}
+
+// New returns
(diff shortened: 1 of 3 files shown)


limiter/limiter.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	r := &Run{
+		requestsLeft: int64(maxRequests),
+		limiter:      limiter,
+	}
+	time.AfterFunc(maxDuration, func() {
+		r.Stop()
+	})
+
```
abhinav: Nit:

    time.AfterFunc(maxDuration, r.Stop)

limiter/limiter.go
```diff
@@ -0,0 +1,46 @@
...
+	return r
+}
+
+// More returns whether more requests can be started or whether we have
+// reached some limit.
+func (r *Run) More() bool {
```
abhinav: Mention that this blocks until the request can be sent.

abhinav:
LGTM with nits

limiter/limiter.go
prashantv: 👍

limiter/limiter.go
prashantv: 👍 
Also updated `More` so it doesn't `Take` if `requestsLeft` is already 0

PR #61, opened 2016-07-07
benchmark: Round robin peer selection to balance connections

prashantv:
Previously, we relied on randomly picking a peer for each connection
which could lead to an imbalance. E.g., if there are 3 peers and 3
connections, there was no guarantee that all 3 peers would receive
a single connection.

Instead of picking randomly, round-robin the peers that can be used
for each connection (starting at a random offset) for better load
balancing.

### DIFF
--- a/bench_method.go
+++ b/bench_method.go
@@ -21,6 +21,7 @@
 package main
 
 import (
+	"math/rand"
 	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


blampe:
🐑

bench_method.go
```diff
@@ -63,19 +64,34 @@ func (m benchmarkMethod) call(t transport.Transport) (time.Duration, error) {
 	return duration, err
 }
 
+func hostPortBalancer(hostPorts []string) func(i int) string {
+	startOffset := rand.Intn(len(hostPorts))
+	return func(i int) string {
+		offset := (startOffset + i) % len(hostPorts)
```
abhinav: worth pulling `len(hostPorts)` into a variable? It's not going to change between calls.

abhinav:
LGTM

bench_method.go
prashantv: 👍

bench_method.go
```diff
@@ -63,19 +64,35 @@ func (m benchmarkMethod) call(t transport.Transport) (time.Duration, error) {
 	return duration, err
 }
 
+func hostPortBalancer(hostPorts []string) func(i int) string {
```
breerly: Is this really specific to hostports? 

You might consider moving it to a `balancer.go` file and renaming to `Balance(strs []string) func(i int) string`.

prashantv: It's not, but I'd rather keep it private here until we need it elsewhere.

breerly: Ah, I could see that. Thanks!

PR #62, opened 2016-07-08
read defaults from ~/.config/yab/defaults.ini

blampe:
@prashantv


main.go
```diff
@@ -90,6 +103,15 @@ It can be used in a curl-like fashion when benchmarking features are disabled.
 		return opts, errExit
 	}
 
+	// Read defaults from ~/.yab.ini
+	if user, err := user.Current(); err != nil {
```
prashantv: instead of using else, suggest reduce nesting here,
```
user, err := user.Current()
if err != nil {
  // fatal
}

[...]
homeDir := user.HomeDir 
```

main.go
```diff
@@ -90,6 +103,15 @@ It can be used in a curl-like fashion when benchmarking features are disabled.
...
+		out.Fatalf("Failed to find the current user: %v\n", err)
+	} else {
+		homeDir := user.HomeDir
+		iniParser := flags.NewIniParser(parser)
+		iniParser.ParseFile(path.Join(homeDir, ".yab.ini"))
```
prashantv: `ParseFile` can return an error, we might want to return an error to the user in that case

main.go
prashantv: instead of using the user.Homedir, what do you think of using `$HOME`? `user.Current` relies on ldap / glibc which we might not need otherwise, and `$HOME` might be easier to add a simple unit test for

blampe: I was trying to be portable but if this is OK then I'm on board.

prashantv: hmm true, let's leave this as is then, i think there's other places using `$USER`, we should probably move all the user-specific logic into a single place in future.

blampe: Related to @breerly's comment, it looks like casimir/xdg-go would do what we want. Would you be OK with me dropping that package in?

prashantv: If we can find a xdg package that works well across all the OSes, then sure. I think we want:
~/.config/yab for OSX and Linux
%APPDATA% for Windows

You can see what the most imported packages for Go are using godoc:
https://godoc.org/?q=xdg

We'll want to find that that's used in other packages but follows conventions in other operating systems too.

main.go
```diff
@@ -139,6 +162,20 @@ func parseAndRun(out output) {
...
 
+// parseDefaultConfigs reads defaults from ~/.config/yab/yab.ini if they're
+// available.
+func parseDefaultConfigs(parser *flags.Parser) error {
+	app := xdg.App{Name: "yab"}
+	configFile := app.ConfigPath("yab.ini")
+	if _, err := os.Stat(configFile); err == nil {
+		iniParser := flags.NewIniParser(parser)
+		if err := iniParser.ParseFile(configFile); err != nil {
+			return fmt.Errorf("couldn't read %q: %q", configFile, err)
```
abhinav: The error should use `%v`. `%q` is for quoting strings.

main.go
```diff
@@ -139,6 +162,20 @@ func parseAndRun(out output) {
...
 }
 
+// parseDefaultConfigs reads defaults from ~/.config/yab/yab.ini if they're
+// available.
+func parseDefaultConfigs(parser *flags.Parser) error {
+	app := xdg.App{Name: "yab"}
+	configFile := app.ConfigPath("yab.ini")
+	if _, err := os.Stat(configFile); err == nil {
+		iniParser := flags.NewIniParser(parser)
+		if err := iniParser.ParseFile(configFile); err != nil {
```
abhinav: Reduce nesting by inverting the outer `if` condition.

```go
if _, err := os.Stat(configFile); err != nil {
  return nil
}

iniParser := flags.NewIniParser(parser)
if err := iniParser.ParseFile(configFile); err != nil {
  ..
}

return nil
```

main.go
```diff
@@ -139,6 +162,20 @@ func parseAndRun(out output) {
 	runWithOptions(*opts, out)
 }
 
+// parseDefaultConfigs reads defaults from ~/.config/yab/yab.ini if they're
+// available.
+func parseDefaultConfigs(parser *flags.Parser) error {
+	app := xdg.App{Name: "yab"}
```
abhinav: Use a constant?

main_test.go
```diff
@@ -360,3 +361,46 @@ func TestAlises(t *testing.T) {
...
+		},
+		{
+			message:       "absent ini file should be ignored",
+			configPath:    path.Join("testdata", "ini", "missing"),
+			expectedError: "",
+		},
+		{
+			message:       "invalid ini file should raise error",
+			configPath:    path.Join("testdata", "ini", "invalid"),
+			expectedError: "couldn't read \"testdata/ini/invalid/yab/yab.ini\": \"testdata/ini/invalid/yab/yab.ini:2: time: unknown unit foo in duration 3foo\"",
```
abhinav: Instead of escaping quotes in strings, use the backticks string literal:

    `couldn't read "testdata/ini/invalid/yab/yab.ini": "..."`

Also, once you fix the `%q` for the error in the error message to use `%v`, this will be slightly different.

main_test.go
```diff
@@ -360,3 +361,46 @@ func TestAlises(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		os.Setenv(configHomeEnv, tt.configPath)
+
+		parser, _ := newParser()
+
+		err := parseDefaultConfigs(parser)
```
abhinav: Nit: reduce superfluous empty lines

main_test.go
```diff
@@ -360,3 +361,46 @@ func TestAlises(t *testing.T) {
...
+
+		if tt.expectedError == "" {
+			assert.Nil(t, err, tt.message)
```
abhinav: use `assert.NoError` instead

main_test.go
```diff
@@ -360,3 +361,46 @@ func TestAlises(t *testing.T) {
...
+	}{
+		{
+			message:       "valid ini file should parse correctly",
+			configPath:    path.Join("testdata", "ini", "valid"),
+			expectedError: "",
+		},
+		{
+			message:       "absent ini file should be ignored",
+			configPath:    path.Join("testdata", "ini", "missing"),
+			expectedError: "",
```
abhinav: Skip specifying the `expectedError` field when using the zero value (`""`)

main_test.go
```diff
@@ -360,3 +361,46 @@ func TestAlises(t *testing.T) {
...
+		os.Setenv(configHomeEnv, tt.configPath)
+
+		parser, _ := newParser()
+
+		err := parseDefaultConfigs(parser)
+
+		if tt.expectedError == "" {
+			assert.Nil(t, err, tt.message)
+		} else {
+			assert.NotNil(t, err, tt.message)
```
abhinav: Use `assert.Error`

main_test.go
```diff
@@ -360,3 +361,46 @@ func TestAlises(t *testing.T) {
...
+			assert.Equal(t, err.Error(), tt.expectedError, tt.message)
```
abhinav: Make this check happen only if the previous check works, or this will panic.

```go
if assert.Error(t, err, tt.message) {
    assert.Equal(t, err.Error(), tt.expectedError, tt.message)
}
```

main.go
```diff
@@ -69,15 +71,30 @@ func main() {
...
+func getOptions(args []string, out output) (*Options, error) {
+	parser, opts := newParser()
 	parser.Usage = "[<service> <method> <body>] [OPTIONS]"
 	parser.ShortDescription = "yet another benchmarker"
 	parser.LongDescription = `
 yab is a benchmarking tool for TChannel and HTTP applications. It's primarily intended for Thrift applications but supports other encodings like JSON and binary (raw).
 
 It can be used in a curl-like fashion when benchmarking features are disabled.
+
+Default options can be specified in a ~/.config/yab/yab.ini file with contents similar to this:
```
breerly: Suggest `~/.config/yab/defaults.ini` instead of `~/config/yab/yab.ini`.

abhinav:
LGTM

main.go
```diff
@@ -110,6 +127,12 @@ It can be used in a curl-like fashion when benchmarking features are disabled.
 		return opts, errExit
 	}
 
+	// Read defaults if they're available.
+	err := parseDefaultConfigs(parser)
+	if err != nil {
```
prashantv: nit: inline `err :=` into if

main_test.go
```diff
@@ -361,6 +362,43 @@ func TestAlises(t *testing.T) {
...
+			message:    "valid ini file should parse correctly",
+			configPath: path.Join("testdata", "ini", "valid"),
+		},
+		{
+			message:    "absent ini file should be ignored",
+			configPath: path.Join("testdata", "ini", "missing"),
+		},
+		{
+			message:       "invalid ini file should raise error",
+			configPath:    path.Join("testdata", "ini", "invalid"),
```
prashantv: nit: make `configPath` just have the "valid" or "invalid" etc, and have the test runner set the rest of the path

main_test.go
```diff
@@ -361,6 +362,43 @@ func TestAlises(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		os.Setenv(configHomeEnv, tt.configPath)
+		parser, _ := newParser()
+		err := parseDefaultConfigs(parser)
+		if tt.expectedError == "" {
+			assert.NoError(t, err, tt.message)
+		} else {
+			assert.Error(t, err, tt.message)
+			assert.Equal(t, err.Error(), tt.expectedError, tt.message)
```
prashantv: use `assert.EqualError` and put the expected error as the first arg. then you can drop the `assert.Error`, just `assert.EqualError(t, tt.expectedError, err, tt.message)`

prashantv:
Some minor nits, otherwise lgtm

prashantv:
:shipit:

PR #63, opened 2016-07-08
better man-page support for pre-formatted code and bullet lists

blampe:
This fixes some mis-alignment issues I noticed with our example code. Basically now if we start a line with a tab we'll try to re-write it in a more groff-friendly way.

I think the version of `man2html` that I'm using doesn't correctly support UTF-8 because it's messing up the charset.

@prashantv

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ install_ci: install
 update_man:
 	go install .
 	yab --man-page > man/yab.1
-	nroff -Tascii -man man/yab.1 | man2html -title yab > man/yab.html
+	groff -man -T html man/yab.1 > man/yab.html
 	[[ -d ../yab_ghpages ]] && cp man/yab.html ../yab_ghpages/man.html
 	@echo "Please update gh-pages"
 
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ install_ci: install
 update_man:
 	go install .
 	yab --man-page > man/yab.1
-	nroff -man man/yab.1 | man2html -title yab > man/yab.html
+	nroff -Tascii -man man/yab.1 | man2html -title yab > man/yab.html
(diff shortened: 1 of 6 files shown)


prashantv:
Need to figure out this `man2html` thing, before I ran a sed to fix them up (it was mainly the `--` that was affected), but I think figuring out why this is happening and fixing it is probably better, since there's a lot more strange characters now

blampe:
Yep totally agree, I'll look into this. @prashantv what version does `man2html` give you? I see 3.0.1.

blampe:
Take another look, it looks better from my end. I still don't like the line wrapping we're doing on the indented blocks, I might follow up with that later.

blampe:
Scratch that, the weird characters are fixed but `man2html` now seems to be removing some sections that still show up correctly in `.1` (e.g., "Examples"). Hm.

prashantv:
I have version 3.0.1 for `man2html` as well. The ASCII output seems good -- although you're right, some things are just completely gone. E,g. some options (like `--peer` and `--service`) are just missing completely. Let me see if I can reproduce that locally.

prashantv:
Same issue with `man2html` for me. I found another converter called `mandoc`: http://mdocml.bsd.lv/ 

However, the rendered HTML is broken in different ways:
image

I'm starting to think maybe we should generate a text file and just link to that instead.

blampe:
Sure, I'll play around with `mandoc` and `pandoc`, and if I can't get those to work I'll just generate the text file instead.

blampe:
OK, looks much better now.

prashantv:
This looks great! 🚢

prashantv:
Updated the gh_pages branch, the man page looks a lot better now, thanks @blampe 
http://yarpc.github.io/yab/man.html

breerly:
This is great, +1

On Fri, Jul 15, 2016 at 8:20 AM, Prashant Varanasi <notifications@github.com

PR #65, opened 2016-07-16
Add option to disable enveloping

prashantv:
Some HTTP services don't use Thrift envelopes, so add option to disable
envelopes.

### DIFF
--- a/options.go
+++ b/options.go
@@ -39,15 +39,16 @@ type Options struct {
 
 // RequestOptions are request related options
 type RequestOptions struct {
-	Encoding    encoding.Encoding `short:"e" long:"encoding" description:"The encoding of the data, options are: Thrift, JSON, raw. Defaults to Thrift if the method contains '::' or a Thrift file is specified"`
-	ThriftFile  string            `short:"t" long:"thrift" description:"Path of the .thrift file"`
-	MethodName  string            `short:"m" long:"method" description:"The full Thrift method name (Svc::Method) to invoke"`
-	RequestJSON string            `short:"r" long:"request" description:"The request body, in JSON or YAML format"`
-	RequestFile string            `short:"f" long:"file" description:"Path of a file containing the request body in JSON or YAML"`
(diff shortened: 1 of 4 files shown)


blampe:
lgtm, +@abhinav for another pair of eyes.

options.go
```diff
@@ -39,15 +39,16 @@ type Options struct {
...
+	ThriftFile       string            `short:"t" long:"thrift" description:"Path of the .thrift file"`
+	MethodName       string            `short:"m" long:"method" description:"The full Thrift method name (Svc::Method) to invoke"`
+	RequestJSON      string            `short:"r" long:"request" description:"The request body, in JSON or YAML format"`
+	RequestFile      string            `short:"f" long:"file" description:"Path of a file containing the request body in JSON or YAML"`
+	HeadersJSON      string            `long:"headers" description:"The headers in JSON or YAML format"`
+	HeadersFile      string            `long:"headers-file" description:"Path of a file containing the headers in JSON or YAML"`
+	Health           bool              `long:"health" description:"Hit the health endpoint, Meta::health"`
+	Timeout          timeMillisFlag    `long:"timeout" default:"1s" description:"The timeout for each request. E.g., 100ms, 0.5s, 1s. If no unit is specified, milliseconds are assumed."`
+	DisableEnvelopes bool              `long:"disable-envelope" description:"Disables Thrift envelopes (disabled by default for TChannel)"`
```
breerly: Suggest "DisableThriftEnveloping" or "DisableThriftEnvelopes".

breerly: Or perhaps "ThriftDisableEnvelope" to put this setting next to the other thrift one.

breerly:
lgtm

abhinav:
👍

options.go
prashantv: good idea, do you think the option should also become `--disable-thrift-envelope`? Or is that too verbose for an option?

PR #66, opened 2016-07-18
Don't leak groff to help output

blampe:
We were accidentally calling `toGroff` on the parser's long description, which is also shown for `--help`. Additionally long descriptions for groups are only displayed in the man page output, so we don't need to set them until they're needed.

@prashantv


prashantv:
lgtm

PR #67, opened 2016-07-18
Release latest yab changes as 0.5.0

prashantv:
Includes:
 - Round robin peer selection when benchmarking for better load distribution
 - Better man pages
 - Option to disable Thrift envelopes


blampe:
:shipit:

PR #68, opened 2016-07-18
Help: Fix reference to --peer-list

abhinav:
The error message was referring to `--hostfile` which doesn't exist.

@prashantv


blampe:
:shipit:

prashantv:
lgtm, thanks for fixing!

PR #71, opened 2016-07-19
Fix peer list bugs for benchmarking

### LINKED ISSUE #70: Benchmark with --peer-list returns an error
Error saying cannot specify peer using `--peer` and `--peer-list`.

prashantv:
Peer-list was ignoring the last line if it did not end with a "\n"
Benchmark warmup failed if --peer-list was used (introduced in
round-robin peer selection, #61).

Fixes #70

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


main_test.go
```diff
@@ -213,6 +214,33 @@ func TestHealthIntegration(t *testing.T) {
 	main()
 }
 
+func TestBenchmarkIntegration(t *testing.T) {
+	origArgs := os.Args
+	defer func() { os.Args = origArgs }()
+
+	const numServers = 5
```
abhinav: Why make this a constant here? It's not used anywhere else. Just for readability?

prashantv: Yeah just for readability

abhinav:
👍

PR #74, opened 2016-07-19
Update README with latest --help output

blampe:
Nitpicking our README and the output of `--help` -- my last change removed the titles for these groupings.

This doesn't need a release.

### DIFF
--- a/main.go
+++ b/main.go
@@ -48,14 +48,10 @@ func findGroup(parser *flags.Parser, group string) *flags.Group {
 	panic("no group called " + group + " found.")
 }
 
-func setGroupShortDesc(parser *flags.Parser, groupName, desc string) {
+func setGroupDescs(parser *flags.Parser, groupName, shortDesc, longDesc string) {
 	g := findGroup(parser, groupName)
-	g.ShortDescription = desc
-}
-
-func setGroupLongDesc(parser *flags.Parser, groupName, desc string) {
-	g := findGroup(parser, groupName)
-	g.LongDescription = desc
+	g.ShortDescription = shortDesc
+	g.LongDescription = longDesc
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -147,9 +155,9 @@ Default options can be specified in a ~/.config/yab/defaults.ini file with conte
 
 	if opts.ManPage {
 		parser.LongDescription = toGroff(parser.LongDescription)
-		setGroupDesc(parser, "request", "Request Options", toGroff(_reqOptsDesc))
-		setGroupDesc(parser, "transport", "Transport Options", toGroff(_transportOptsDesc))
-		setGroupDesc(parser, "benchmark", "Benchmark Options", toGroff(_benchmarkOptsDesc))
+		setGroupLongDesc(parser, "request", toGroff(_reqOptsDesc))
```
prashantv: The short group description is used to key the group, and since the short group name (like "request") was already changed to "Request Options", trying to find a group called "request" will fail and so I think calling yab with `--man-page` with this change will cause a panic

Can we:
 - add a simple test to `main_test.go` which will call `--man-page` (we should also change the `WriteManPage` to pass `out` as a `Writer` so that we can redirect output rather than flooding the terminal during tests)
 - go back to the old way where set set short + long descriptions for group in a single function so we don't have to key on 2 different strings when setting these attributes. since the long desc isn't shown in help anyway and is only used for the man page, it seems OK to always set the groff version of the desc earlier.

prashantv:
@blampe Rebased and pushed a commit to address the man page issue. Can you review my last commit, thanks!

blampe:
LGTM! Sorry I didn't get to this sooner :)

PR #79, opened 2016-07-22
Support TMultiplexedProtocol for HTTP+Thrift

### LINKED ISSUE #77: HTTP/Enveloping: Support TMultiplexedProtocol
`yab` currently does not support TMultiplexedProtocol for HTTP/Thrift services. The only difference is that instead of using the `$methodName` in the envelope, it will use `$serviceName$delimiter$methodName`, where `$delimiter` is configurable and it defaults to `:`.

prashantv:
Fixes #77

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ install_ci: install
 .PHONY: update_man
 update_man:
 	go install .
-	yab --man-page > man/yab.1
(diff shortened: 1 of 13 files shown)


abhinav:
lgtm

blampe:
LGTM

PR #86, opened 2016-07-28
Allow JSON to be used with non-map requests

### LINKED ISSUE #84: problem parsing response with --encoding=json
```
$ yab --peer-list=hosts.json --service=autobahn --method=hosts_v1 --encoding=json --request='{"serviceName":"foo"}'
Failed while parsing response: failed to parse JSON: json: cannot unmarshal array into Go value of type map[string]interface {}
```

with tcurl the response is

```
$ tcurl -H hosts.json autobahn hosts_v1 -3 '{"serviceName":"foo"}' | jq .
{
  "ok": true,
  "head": null,
  "body": [...],
  "headers": {
    "as": "json"
  },
  "trace": "7075701ec1913375"
}
```

prashantv:
Fixes #84

### DIFF
--- a/encoding/encoding_test.go
+++ b/encoding/encoding_test.go
@@ -143,7 +143,7 @@ func TestJSONEncodingResponse(t *testing.T) {
 
 	tests := []struct {
 		data   string
-		want   map[string]interface{}
+		want   interface{}
 		errMsg string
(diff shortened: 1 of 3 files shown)


blampe:
LGTM

PR #88, opened 2016-07-29
Add "ok" field to output containing TChannel ApplicationError bit

### LINKED ISSUE #81: Expose TChannel ApplicationError
Currently we do not expose the TChannel "ApplicationError" bit. It's possible due to bugs that a server or relay returns the body of a Thrift exception, but does not set the application error bit correctly.

These issues will be much easier to debug if the application error is exposed on the response (similar to tcurl which returns an "ok" field)

prashantv:
Fixes #81

### DIFF
--- a/main.go
+++ b/main.go
@@ -255,8 +255,8 @@ func runWithOptions(opts Options, out output) {
 	if len(response.Headers) > 0 {
 		outSerialized["headers"] = response.Headers
 	}
-	if response.Trace != "" {
-		outSerialized["trace"] = response.Trace
+	for k, v := range response.TransportFields {
+		outSerialized[k] = v
 	}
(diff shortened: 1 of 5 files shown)


blampe:
LGTM

abhinav:
Are there no tests for Thrift app errors that need to be updated?

prashantv:
The tests for application errors just test the response body, not the full response. I added tests in `tchannel_test` to ensure that application errors are set in the `TransportFields`:
https://github.com/yarpc/yab/pull/88/files#diff-1ca4480574725c1829319c7dfa2b6070

And ensured that `TransportFields` are outputted:
https://github.com/yarpc/yab/pull/88/files#diff-53618cae32e23ae47e1c35f10da92654

abhinav:
SGTM :shipit:

PR #92, opened 2016-08-02
Move info about default configs to the manpage and automatically update README.md

blampe:
* Unclutters `--help` by moving config defaults to `yab.1`
* Adds an admittedly gross one-liner to automatically update `README.md` in the same way we update our manpages. (`brew install moreutils` if you're on OS X and missing `sponge`)


prashantv:
That one-line is kind of terrifying, but it works so :shipit:

PR #95, opened 2016-08-03
Update glide and use testImports for test deps

prashantv:
Recreated the `glide.yaml` using `glide init` which automatically put test dependencies in `testImport`. This doesn't work as well for test subpackages (e.g., tchannel-go/testutils).

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,18 +1,18 @@
-hash: 8e5d5d46415cfa0f686537efda563c9d39bc2cad2ae2b9b6c5bece08e710f3d5
-updated: 2016-08-03T09:34:42.07213616-07:00
+hash: 55fd0a7525390f35a5dfcc739bd4aee790f7cc81aa3ce4576de948675bb3fca1
+updated: 2016-08-24T23:17:59.856656069-07:00
 imports:
 - name: github.com/apache/thrift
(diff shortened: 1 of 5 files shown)


abhinav:
👍

prashantv:
This is failing due to a bug in glide: https://github.com/Masterminds/glide/issues/517

mattfarina:
Was this `glide.lock` file generated with `glide update`? What version of Glide was used and how was it installed?

I ask because `github.com/apache/thrift` shows up under both `import` and `testImport` which may be the problem. Trying to assess how that happened. Working on a fix.

prashantv:
@mattfarina Yes, I generated the `glide.lock` using `glide update`. The reason it seems to show up in both:
 - There is a direct dependency from yab's tests on `github.com/apache/thrift`
 - There is no direct dependency from yab's non test code on `github.com/apache/thrift` but there is an indirect dependency through `github.com/uber/tchannel-go/thrift`.

In cases like this, I think `glide up` should remove entries from the `testImport` section in `glide.lock` that are in the `import` section.

prashantv:
This works fine with the latest glide. I updated the dependencies again and fixed the YARPC integration test to pick up the new API change.

PR #96, opened 2016-08-10
Disable flags unquoting for JSON payloads

prashantv:
See: https://github.com/jessevdk/go-flags/issues/185

Previously, if a user wanted to send a string JSON payload, they may
try: `yab -r '"body"'`. However, go-flags was unescaping the body
that the shell may already have unescaped, which ends up sending the
raw string `body` instead of `"body"`.

Disable unquoting using the "unquote" option so it's easier to pass
string JSON payloads to JSON services.


abhinav:
👍

PR #97, opened 2016-08-10
Default to JSON encoding for endpoints without '::'

prashantv:
Currently, the user is forced to specify `-e json` if they are making a
JSON call, which is a little different from tcurl and a little more
verbose.

We only support Thrift, JSON and Raw, and since Thrift can be inferred
based on the method name, and raw is rarely used, defaulting to JSON
should lead to a better user experience.


abhinav:
👍

PR #99, opened 2016-08-21
If ThriftFile is specified, default encoding to Thrift

prashantv:
Event if the method name doesn't contain '::', assume Thrift, since
we will return better error messages (e.g., list all the available
services) rather than giving the error "please specify an encoding".


blampe:
👍

abhinav:
:shipit:

PR #103, opened 2016-08-25
Mention mac config directories in the man page

prashantv:


abhinav:
:shipit: 

Altohugh I would be in favor of overriding that path

prashantv:
Yeah I'm not a fan of that path either, but it does seem to be the recommended location for OSX apps to store their configuration. I think the user experience is worse though, since they have to remember 2 locations and switch based on which OS they're currently on.

@breerly Do you have thoughts on the location for configuration files on OSX?

PR #104, opened 2016-08-25
Override default peer options based on args

prashantv:
If defaults sets the hosts file, but args sets a single peer, the
final options should only contain the single peer. If both are set,
yab returns an error.

This is required if we want to set a default peer list in a global
config that is only used when the user doesn't override the peer in
any way.

### DIFF
--- a/main.go
+++ b/main.go
@@ -193,10 +193,10 @@ func overrideDefaults(defaults *Options, args []string) {
 	argsParser.ParseArgs(args)
 
 	// Clear default peers if the user has specified peer options in args.
-	if len(argsOnly.TOpts.HostPorts) > 0 && defaults.TOpts.HostPortFile != "" {
+	if len(argsOnly.TOpts.HostPorts) > 0 {
 		defaults.TOpts.HostPortFile = ""
 	}
-	if len(argsOnly.TOpts.HostPortFile) > 0 && len(defaults.TOpts.HostPorts) > 0 {
+	if len(argsOnly.TOpts.HostPortFile) > 0 {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -181,6 +183,24 @@ func parseAndRun(out output) {
...
+// E.g., if the defaults has a peer list file, and the user has specifed
+// a peer through the command line, then the final options should only
+// contain the peer specified in the args.
+func overrideDefaults(defaults *Options, args []string) {
+	argsParser, argsOnly := newParser()
+	argsParser.ParseArgs(args)
+
+	// Clear default peers if the user has specified peer options in args.
+	if len(argsOnly.TOpts.HostPorts) > 0 && defaults.TOpts.HostPortFile != "" {
+		defaults.TOpts.HostPortFile = ""
```
abhinav: Do you really need the second check? If you always set `HostPortFile` to `""` when `HostPorts` is non-empty, you get the same result.

main.go
```diff
@@ -181,6 +183,24 @@ func parseAndRun(out output) {
...
+	}
+	if len(argsOnly.TOpts.HostPortFile) > 0 && len(defaults.TOpts.HostPorts) > 0 {
+		defaults.TOpts.HostPorts = nil
```
abhinav: Same as above

main.go
prashantv: Good idea, done

main.go
prashantv: 👍

abhinav:
👍

PR #106, opened 2016-08-26
Add linux build; Ensure build makes a binary

ZymoticB:
Add a build target which will build a linux compatible binary.
Additionally ensure that the build target will actually build a binary
it seems that `go build -i ./a/... ./number/... ./of/... ./pkgs/... .`
only builds the libraries so we strip the local dir out of $PACKAGES
and then explicitly build it.

Also add a DEFAULT_GOAL to be explicit and avoid ever accidentally
breaking the default behaviour by adding a target in the wrong place.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,17 @@
-PACKAGES := $(shell glide novendor)
+PACKAGES := $(filter-out ., $(shell glide novendor))
 
 export GO15VENDOREXPERIMENT=1
+
+.DEFAULT_GOAL:=build
 
 
 .PHONY: build
 build:
 	go build -i $(PACKAGES)
+	go build -i .
 
+build_prod:
+	env GOOS=linux GOARCH=amd64 $(MAKE)
 
 .PHONY: install
 install:
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PACKAGES := $(filter-out ., $(shell glide novendor))
+PACKAGES := $(shell glide novendor)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,12 +1,17 @@
-PACKAGES := $(shell glide novendor)
+PACKAGES := $(filter-out ., $(shell glide novendor))
```
prashantv: We should leave `.` in `PACKAGES` since it's also used for the test command and we want to run tests for the main package too.

Makefile
```diff
@@ -1,12 +1,17 @@
...
+.DEFAULT_GOAL:=build
+
 
 .PHONY: build
 build:
 	go build -i $(PACKAGES)
+	go build -i .
 
+build_prod:
+	env GOOS=linux GOARCH=amd64 $(MAKE)
```
prashantv: Instead of this, I have a temporary branch (just pushed to https://github.com/yarpc/yab/compare/cross_builds?expand=1) that adds a script that will build the binary for a bunch of platforms. I typically use that to build release binaries to upload to the Github release page.

What do you think of that approach? Would prefer not to build a binary called `yab` that is linux/amd64 specific from the makefile directly, since "build_prod" could mean "build with optimizations for production" rather than the architecture.

ZymoticB: That script seems like a more reasonable approach, given that building for linux isn't the usual testing workflow; better off living in a script. I'll remove that piece

prashantv:
lgtm, thanks @ZymoticB

PR #107, opened 2016-08-26
Add Error Rate to benchmark output

ZymoticB:
Example of new output

> Benchmark parameters:
  CPUs:            24
  Connections:     48
  Concurrency:     5
  Max requests:    18000
  Max duration:    1m0s
  Max RPS:         300
Errors:
   969: tchannel error ErrCodeDeclined: Service is not healthy
    27: tchannel error ErrCodeTimeout: request timed out after XXXXms (limit was XXXms)
   284: tchannel error ErrCodeTimeout: request timed out after XXXms (limit was XXXms)
  1005: tchannel error ErrCodeTimeout: timeout
Total errors: 2285
Error rate: 0.1535
Latencies:
  0.5000: 613.118244ms
  0.9000: 907.028082ms
  0.9500: 951.607273ms
  0.9900: 998.944412ms
  0.9990: 1.697313257s
  0.9995: 1.803770667s
  1.0000: 1.971703062s
Elapsed time:      1m1.077s
Total requests:    14883
RPS:               243.68

### DIFF
--- a/bench_state.go
+++ b/bench_state.go
@@ -98,7 +98,7 @@ func (s *benchmarkState) printErrors(out output) {
 		out.Printf("  %4d: %v\n", v, k)
 	}
 	out.Printf("Total errors: %v\n", s.totalErrors)
-	out.Printf("Error rate: %.4f\n", float32(s.totalErrors)/float32(s.totalRequests))
+	out.Printf("Error rate: %.4f%%\n", 100*float32(s.totalErrors)/float32(s.totalRequests))
 }
 
 func (s *benchmarkState) getQuantile(q float64) time.Duration {
--- a/bench_state.go
+++ b/bench_state.go
@@ -30,9 +30,12 @@ import (
 )
 
 type benchmarkState struct {
-	statter   statsd.Client
-	errors    map[string]int
-	latencies []time.Duration
+	statter       statsd.Client
+	errors        map[string]int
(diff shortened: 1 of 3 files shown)


bench_state.go
```diff
@@ -86,6 +86,7 @@ func (s *benchmarkState) printErrors(out output) {
 		total += v
 	}
 	out.Printf("Total errors: %v\n", total)
+	out.Printf("Error rate: %.4f\n", float32(total)/float32(len(s.latencies)))
```
prashantv: nit: latencies only includes requests that succeeded, so it's possible that all requests failed in which case this will probably get a panic for dividing by 0. Can we put this in a `if len(s.latencies) > 0` block

ZymoticB: That also means that some other stuff is reporting incorrect numbers (e.g. Total Requests in benchmark.go). I'll fix those up as well in this diff.

bench_state_test.go
```diff
@@ -49,7 +49,19 @@ func TestBenchmarkStateErrors(t *testing.T) {
 	})
 
 	buf, out := getOutput(t)
+
+	//before merge
```
prashantv: nit: here and below, space after `// `

prashantv:
lgtm, thanks for fixing up the reporting @ZymoticB

PR #110, opened 2016-09-21
Check the system xdg dir for a yab config file

eklitzke:
This tests the user's config dir for a yab config, and if that fails yab
will fallback to checking the system directories. On a Linux machine
this means that the config files will be checked in this order (with the
first found taking precedence):

    - ~/.config/yab/defaults.ini
    - /etc/xdg/yab/defaults.ini

The motivation behind this change is to allow system administrators to
install a system-wide default configuration file.

### DIFF
--- a/main.go
+++ b/main.go
@@ -201,15 +201,31 @@ func overrideDefaults(defaults *Options, args []string) {
 	}
 }
 
+// findBestConfigFile finds the best config file to use. An empty string will be
+// returned if no config file should be used.
+func findBestConfigFile() string {
+	app := xdg.App{Name: "yab"}
+
+	// Find the best config path to use, preferring the user's config path and
+	// falling back to the system config path.
+	configPaths := []string{app.ConfigPath("defaults.ini")}
+	configPaths = append(configPaths, app.SystemConfigPaths("defaults.ini")...)
+	var configFile string
+	for _, path := range configPaths {
+		if _, err := os.Stat(path); err == nil {
+			configFile = path
+			break
+		}
+	}
+	return configFile
+}
+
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -206,10 +206,19 @@ func overrideDefaults(defaults *Options, args []string) {
 func parseDefaultConfigs(parser *flags.Parser) error {
 	app := xdg.App{Name: "yab"}
 
-	configFile := app.ConfigPath("defaults.ini")
-	if _, err := os.Stat(configFile); err != nil {
-		// No defaults file to read
-		return nil
+	// Find the best config path to use, preferring the user's config path and
```
prashantv: Can load multiple files? I imagine if there are multiple files, we'd try the system ones first, then override from the user's configuration?

eklitzke: AFAIK the current parser doesn't allow this (although obviously it's possible to do in principle).

eklitzke:
OK, will add tests.

main.go
prashantv: what happens if you call `iniPaser.ParseFile` multiple times?

eklitzke: Was just thinking about this. The way XDG works for all of the files on my Linux desktop is that it does not merge files, it uses the same precedence logic I've implemented here. The recommended way to customize a system XDG file on Linux is to copy the system one to your home directory and then modify it.

The issue here is that some of the flags we use with yab do something only if set. That means that you can't meaningfully unset a flag by putting it in your home directory and clearing that value.

eklitzke:
Just added test, let me know what you think.

main.go
prashantv: Ok cool, makes sense.

prashantv:
lgtm!

PR #111, opened 2016-09-22
benchmarking: Allow unlimited requests (capped by max duration)

### LINKED ISSUE #105: supprt setting a "nolimit" for num_requests
Using all parameters i am hitting the max requests over 20 minutes with 100000000000000 max requests.

```
-n -1
``` 
would do just fine :D

prashantv:
Currently we always require a MaxRequests. Instead change the default to
have no MaxRequests, and make only MaxRequests or MaxDuration required
for a benchmark.

If RPS and MaxDuration are provided, an implicit MaxRequests is
calculated. Otherwise, the test runs until the max duration is hit.

Fixes #105

### DIFF
--- a/limiter/limiter.go
+++ b/limiter/limiter.go
@@ -10,6 +10,7 @@ import (
 // Run represents a single run that is limited by either
 // a number of requests, or can be stopped halfway.
 type Run struct {
+	unlimited    atomic.Bool
 	requestsLeft atomic.Int64
 	limiter      ratelimit.Limiter
 	cancel       chan struct{}
@@ -24,11 +25,15 @@ func New(maxRequests, rps int, maxDuration time.Duration) *Run {
 	}
 
 	r := &Run{
+		unlimited:    *atomic.NewBool(maxRequests == 0),
 		requestsLeft: *atomic.NewInt64(int64(maxRequests)),
 		limiter:      limiter,
 		cancel:       make(chan struct{}),
 	}
-	time.AfterFunc(maxDuration, r.Stop)
+
+	if maxDuration > 0 {
+		time.AfterFunc(maxDuration, r.Stop)
+	}
 
 	return r
 }
@@ -37,6 +42,11 @@ func New(maxRequests, rps int, maxDuration time.Duration) *Run {
 // reached some limit. If more requests can be made, it blocks until the rate
 // limiter allows another request.
 func (r *Run) More() bool {
+	if r.unlimited.Load() {
+		r.limiter.Take(r.cancel)
+		return true
+	}
+
 	if r.requestsLeft.Load() >= 0 {
 		r.limiter.Take(r.cancel)
 	}
@@ -50,5 +60,6 @@ func (r *Run) Stop() {
 		return
 	}
 	r.requestsLeft.Store(0)
+	r.unlimited.Store(false)
 	close(r.cancel)
 }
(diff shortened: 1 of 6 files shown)


limiter/limiter.go
```diff
@@ -24,11 +25,15 @@ func New(maxRequests, rps int, maxDuration time.Duration) *Run {
 	}
 
 	r := &Run{
+		unlimited:    *atomic.NewBool(maxRequests == 0),
```
eklitzke: Does this actually need to be an atomic? It only changes during stop, and a non-atomic read then should be fine.

options.go
```diff
@@ -82,8 +82,8 @@ type TransportOptions struct {
 
 // BenchmarkOptions are benchmark-specific options
 type BenchmarkOptions struct {
-	MaxRequests int           `short:"n" long:"max-requests" default:"1000000" description:"The maximum number of requests to make"`
-	MaxDuration time.Duration `short:"d" long:"max-duration" default:"0s" description:"The maximum amount of time to run the benchmark for"`
+	MaxRequests int           `short:"n" long:"max-requests" default:"0" description:"The maximum number of requests to make. Values <= 0 imply no limit."`
```
eklitzke: This is not right since the unlimited flag is only set if max-requests is zero, but the comment here implies that -1 would work.

prashantv: Good point, only 0 should be supported. Negative values should probably throw an error.

limiter/limiter.go
prashantv: We access `unlimited` from `Stop` and from `More`. `More` is running on the goroutines making requests, while `Stop` will run from either a timer goroutine, or the main goroutine waiting for user requests to end. So we'll have two accesses to the same variable from different goroutines, where one is a write, so it's a data race.

We can actually test this by making it a non-atomic bool, and running the tests with `-race`, and we get:
```
Read at 0x00c4200161e0 by goroutine 24:
  github.com/yarpc/yab/limiter.(*Run).More()
      /Users/prashant/gocode/src/github.com/yarpc/yab/limiter/limiter.go:45 +0x3f
  github.com/yarpc/yab/limiter.TestUnlimitedRequests()
      /Users/prashant/gocode/src/github.com/yarpc/yab/limiter/limiter_test.go:80 +0x1ce
  testing.tRunner()
      /usr/local/Cellar/go/1.7/libexec/src/testing/testing.go:610 +0xc9

Previous write at 0x00c4200161e0 by goroutine 25:
  github.com/yarpc/yab/limiter.(*Run).Stop()
      /Users/prashant/gocode/src/github.com/yarpc/yab/limiter/limiter.go:63 +0x8b
  github.com/yarpc/yab/limiter.(*Run).Stop-fm()
      /Users/prashant/gocode/src/github.com/yarpc/yab/limiter/limiter.go:35 +0x41
```

eklitzke: Just because the reads and writes race doesn't mean it needs to be atomic. It means that you can get the wrong read value while a write is happening.

If this is non-atomic it means that in a race scenario you could do an extra request which is not a problem afaik.

eklitzke:
this lgtm now

limiter/limiter.go
prashantv: The Go memory model is stricter than other memory models,
https://golang.org/ref/mem

The tldr is
```
Programs that modify data being simultaneously accessed by multiple goroutines must serialize such access.
```

In x86, this could be a volatile, and the stale read would be OK. Go doesn't allow that though.

The data race detector has no false positives in Go, so the fact that it gave a race when I made it non-atomic means it's definitely a data race.

PR #116, opened 2016-09-22
Release latest yab changes as 0.6.2

prashantv:


prashantv:
Will land once Travis is green

Issue #100, opened 2016-08-23
support for varying requests in benchmarks


blampe:
Right now we only take one body and duplicate it for every request within a benchmark. @zymoticb pointed out that it's sometimes useful to benchmark different request bodies concurrently, and I wanted to open up a conversation around whether we could (or should) support something like that. 

Concretely, this means something like "I want half of my benchmark's requests to use a body FOO and the other half to use BAR."

I was thinking we might be able to support a directory argument, and each file therein would be interpreted as a unique request body. With a heuristic like "append .headers to the request's filename" we could also attach per-request headers. This would leave a directory structure like so:

```
requests/
    foo
    foo.headers
    bar

$ yab --requests=./requests ...
```

This would cause yab to round-robin between foo and bar; and foo would override headers provided on the CLI.

The user could then control the distribution of requests with this file structure. For example if I want 2/3 of requests to be FOO and 1/3 to be BAR I would simply drop FOO.1 and FOO.2 into the directory.

Thoughts?

prashantv:
This seems totally reasonable. Currently, before the benchmark, yab does:
 - a single request to the specified endpoint, verifies the result
 - some k warmup requests

If we had multiple requests, would we want a loop over all the requests initially? Or should we just pick a random one, then rotate through the specified requests during warmup and benchmarks only.

Is there any other parameters we'd want to tweak on a per-request basis?

ZymoticB:
Timeouts would need to be tweaked on a per-request basis; at least for the use case I'm currently targeting.

I think a verification request to each endpoint is definitely useful.

How fast is the "k" done? In my current use case I have endpoints which range (normalizing traffic to 100 rps) from 40 rps to 0.3 rps. In the case of the "0.3" rps the warm up requests need to either be very slow or non-existent, but on the other end a "ramp up" is definitely useful. Maybe this value needs to be tuneable per-endpoint.

prashantv:
Hmm, that seems like it will complicate yab a lot to have that level of customization on a per-request basis. Right now, the warmup just makes a set number of requests (customizable via `-k`) and starts the benchmark after that warmup. With multiple requests, it may round robin those k requests.

If there's two different endpoints wich such different behaviours, it seems like the user should run `yab` twice. I imagine multiple requests/headers to be used when the overall load on the service is similar, and they both have similar performance. Otherwise, it doesn't really make sense to aggregate the performance results for very different endpoints.

Thoughts?

ZymoticB:
[...]

PR #117, opened 2016-09-30
Allow yaml templates to specify RPC calls

jsu1212:
This will allow you to specify RPCs as template files on the command line:

`yab --timeout 3s --peer 127.0.0.1:21300 -y time.yaml`

Example contents of `time.yaml`:

```
service: "time"
method: "time::queryV2"
thrift: "path/to/some/files/time.idl"
headers:
    header1: value1
    header2: value2
request:
    location: 
        latitude: 37.7
        longitude: -122.4
        cityId: 1
```

I'm planning a template generator which will come in a later PR.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -839,3 +839,17 @@ func TestNoWarmupBenchmark(t *testing.T) {
 	assert.Contains(t, buf.String(), "Total errors: 100")
 	assert.Contains(t, buf.String(), "Error rate: 100")
 }
+
+func TestTemplates(t *testing.T) {
+	origArgs := os.Args
+	defer func() { os.Args = origArgs }()
+
+	echoAddr := echoServer(t, "", []byte{0})
+	os.Args = []string{
+		"yab",
+		"-y", exampleTemplate,
+		"-p", echoAddr,
+	}
+
+	main()
+}
--- a/template.go
+++ b/template.go
@@ -45,4 +45,6 @@ func readYamlRequest(opts *Options) error {
 	opts.ROpts.HeadersJSON = string(headers)
 	opts.ROpts.RequestJSON = string(body)
 	opts.ROpts.Timeout = timeMillisFlag(t.Timeout)
+
+	return nil
 }
--- a/template.go
+++ b/template.go
@@ -5,6 +5,7 @@ import (
 	"log"
 
 	"gopkg.in/yaml.v2"
+	"time"
 )
 
 type template struct {
@@ -13,6 +14,7 @@ type template struct {
 	Method  string            `yaml:"method"`
 	Headers map[string]string `yaml:"headers"`
 	Request interface{}       `yaml:"request"`
+	Timeout time.Duration     `yaml:"timeout"`
 }
 
 func readYamlRequest(opts *Options) {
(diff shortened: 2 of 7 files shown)


blampe:
#100 looks related

template.go
```diff
@@ -0,0 +1,50 @@
...
+type template struct {
+	Service string            `yaml:"service"`
+	Thrift  string            `yaml:"thrift"`
+	Method  string            `yaml:"method"`
+	Headers map[string]string `yaml:"headers"`
+	Request interface{}       `yaml:"request"`
+	Timeout time.Duration     `yaml:"timeout"`
+}
+
+func readYamlRequest(opts *Options) {
```
prashantv: can we return an `error` and let the caller do `out.Fatalf` instead?

template.go
```diff
@@ -0,0 +1,50 @@
+package main
+
+import (
+	"io/ioutil"
+	"log"
+
+	"gopkg.in/yaml.v2"
+	"time"
```
prashantv: nit: import should be at the top with the other standard library imports

template.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		log.Fatalf("Unable to marshal yaml: %v\n", err)
+	}
+
+	headers, err := yaml.Marshal(t.Headers)
+	if err != nil {
+		log.Fatalf("Unable to marshal yaml: %v\n", err)
+	}
+
+	// Should we overwrite if these clash with options specified on the command line?
```
prashantv: I think the yaml options should be the "base" options, that the user should be able to override via the command line. What do you think?

main_test.go
```diff
@@ -839,3 +839,17 @@ func TestNoWarmupBenchmark(t *testing.T) {
 	assert.Contains(t, buf.String(), "Total errors: 100")
 	assert.Contains(t, buf.String(), "Error rate: 100")
 }
+
```
jsu1212: This test is admittedly a little weak, but should cover the most basic functionality.  @prashantv, open to ideas on how to improve this.

kriskowal:
I’m landing changes to dev that distinguish transport headers from application headers. I’ve also added a request option that is simply a map of headers that overrides the file name and JSON header request options, which may make it simpler to thread them from the template.

template.go
```diff
@@ -0,0 +1,50 @@
...
+	"io/ioutil"
+	"time"
+
+	"gopkg.in/yaml.v2"
+)
+
+type template struct {
+	Service string            `yaml:"service"`
+	Thrift  string            `yaml:"thrift"`
+	Method  string            `yaml:"method"`
```
kriskowal: In YARPC, we are exposing method as "Procedure". This template could also expose "Caller" as an optional override, just as is possible at the CLI. "Baggage" is now an option on the dev branch as well. I think it will be good to punt on transport headers though.

kriskowal:
Let’s make this work with a shebang.

Here’s `timequery2`

```
#/usr/bin/env yab -Y
service: "time"
method: "time::queryV2"
thrift: "path/to/some/files/time.idl"
headers:
    header1: value1
    header2: value2
request:
    location: 
        latitude: 37.7
        longitude: -122.4
        cityId: 1
```

And override from the CLI.

```
$ timequery2 --timeout 3s --peer 127.0.0.1:21300
```

prashantv:
@kriskowal Great idea!

kriskowal:
@jsu1212 I’ve added you as a collaborator and rebased these changes. I’m going to open a pull request to replace this one using a branch in this repository. #134

PR #121, opened 2016-10-18
Support -H, -B, --jaeger for headers and baggage

kriskowal:


transport/http.go
```diff
@@ -30,6 +30,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/opentracing/opentracing-go"
+
```
prashantv: nit: can remove blank since this is a third party import like the below import

bench_method_test.go
```diff
@@ -144,10 +138,11 @@ func TestBenchmarkMethodCall(t *testing.T) {
 	}
 
 	tOpts := TransportOptions{
+		CallerName:  "bar",
 		ServiceName: "foo",
 		HostPorts:   []string{s.hostPort()},
 	}
-	tp, err := getTransport(tOpts, encoding.Thrift)
+	tp, err := getTransport(tOpts, encoding.Thrift, nil)
```
prashantv: What is the difference in behaviour between `nil` here and `NoopTracer` in `bench_method.go`?

bench_method.go
```diff
@@ -38,7 +39,7 @@ type benchmarkMethod struct {
 // up by making some number of requests through it.
 func (m benchmarkMethod) WarmTransport(opts TransportOptions, warmupRequests int) (transport.Transport, error) {
 	opts.benchmarking = true
-	transport, err := getTransport(opts, m.serializer.Encoding())
+	transport, err := getTransport(opts, m.serializer.Encoding(), opentracing.NoopTracer{})
```
prashantv: Just realized, baggage only gets with the Jaeger tracer right? Perhaps we can't do this unless we explicitly disallow baggage with benchmarking? Otherwise, the initial request will use baggage but the benchmarking requests will not.

main.go
```diff
@@ -252,8 +256,30 @@ func runWithOptions(opts Options, out output) {
...
 
+	if opts.TOpts.CallerName != "" {
+		if opts.TOpts.benchmarking {
+			out.Fatalf("Cannot override caller name when running benchmarks\n")
+		}
+	} else {
+		opts.TOpts.CallerName = "yab-" + os.Getenv("USER")
+	}
+
+	var tracer opentracing.Tracer
```
prashantv: can we move the logic of get a tracer into a separate method?

transport/interface.go
```diff
@@ -23,15 +23,21 @@ package transport
...
-	Timeout time.Duration
-	Headers map[string]string
-	Body    []byte
+	Method          string
+	Timeout         time.Duration
+	Headers         map[string]string
+	Baggage         map[string]string
+	RoutingKey      string
+	RoutingDelegate string
+	ShardKey        string
```
prashantv: I don't know if routing key/delegate/shard key belong as top level fields on the request, unless we've made those transport-agnostic concerns in YARPC?

transport/interface.go
```diff
@@ -23,15 +23,21 @@ package transport
...
+	Body            []byte
+	Tracer          opentracing.Tracer
```
prashantv: The `Request` and `Response` objects only contain values that represent the on-wire presentation currently. I'm not sure if Tracer should be here, should we move it into the `Call` interface instead?

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@ Changelog
 =========
 
 * Upgrade to TChannel 1.2.0 and add Open Tracing support.
+* Add support for -H key:value header arguments.
```
breerly: Suggest:

* Add support for headers with argument `-H key:value`.
* Add support for baggage with argument `-B key:value` when `--jaeger` is passed.

bench_method_test.go
kriskowal: If you thread nil, TChannel and YARPC will fall back to the global tracer. I’m going to explicitly thread NoopTracer and mandate a non-nil tracer in getTransport, so there’s no question for yab.

main_test.go
```diff
@@ -384,8 +384,7 @@ func TestAlises(t *testing.T) {
 		{
 			args: []cmdArgs{
 				{"-P", "file"},
-				{"-H", "file"},
```
breerly: So we're stealing the `-H` flag for headers?

kriskowal: yes.

kriskowal:
Added passing tests for baggage. Added a passing test for TChannel headers. Left a broken test for HTTP headers with #125 

Now seeking only a stamp from @prashantv

main.go
```diff
@@ -325,5 +336,42 @@ func makeRequest(t transport.Transport, request *transport.Request) (*transport.
...
+		for k, v := range request.Baggage {
+			span = span.SetBaggageItem(k, v)
+		}
+		ctx = opentracing.ContextWithSpan(ctx, span)
+	}
+
 	return t.Call(ctx, request)
 }
+
+func makeInitialRequest(out output, transport transport.Transport, serializer encoding.Serializer, req *transport.Request) {
```
kriskowal: I am not sure where this new code is coming from. There might be rebase magic here.

transport.go
```diff
@@ -29,23 +29,23 @@ import (
 	"io/ioutil"
 	"net"
 	"net/url"
-	"os"
 	"strings"
 
+	"github.com/opentracing/opentracing-go"
+	"github.com/uber/tchannel-go"
 	"gopkg.in/yaml.v2"
```
kriskowal: third party group should be below first party group

transport/interface.go
```diff
@@ -23,15 +23,19 @@ package transport
 import (
 	"time"
 
+	"github.com/opentracing/opentracing-go"
```
kriskowal: should be in same group as bellow

transport_test.go
```diff
@@ -21,10 +21,13 @@
 package main
 
 import (
+	"fmt"
 	"os"
 	"testing"
 	"time"
 
+	"github.com/opentracing/opentracing-go"
+	opentracing_ext "github.com/opentracing/opentracing-go/ext"
```
kriskowal: should separate third party group below

bench_method.go
```diff
@@ -25,6 +25,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/opentracing/opentracing-go"
```
prashantv: +1, same throughout this change

server_util_test.go
```diff
@@ -78,11 +83,21 @@ func (methodsT) echo() handler {
 
 func (methodsT) traceEnabled() handler {
```
prashantv: can you add a comment that explains the returned result from the method

bench_method_test.go
```diff
@@ -54,24 +55,18 @@ func benchmarkMethodForTest(t *testing.T, methodString string, p transport.Proto
...
-		method         string
-		wantErr        string
+		hostPort string
+		method   string
+		wantErr  string
 	}{
 		{
 			hostPort: s.hostPort(),
 		},
-		{
```
prashantv: this looks unrelated to headers -- should this have been handled in a previous diff?

server_util_test.go
```diff
@@ -37,8 +39,10 @@ type server struct {
 
 type handler func(ctx context.Context, args *raw.Args) (*raw.Res, error)
 
-func newServer(t *testing.T) *server {
-	ch := testutils.NewServer(t, testutils.NewOpts().SetServiceName("foo").DisableLogVerification())
+func newServer(t *testing.T, tracer opentracing.Tracer) *server {
```
prashantv: Rather than take a tracer (and get into the pattern of everyone passing `nil` for default options), we should probably just implement a lightweight options pattern here,


```go
func withTracer(opentracing.Tracer) func(*testutils.ChannelOpts) {
  return func(opts *testutils.ChannelOpts) {
    opts.Tracer = tracer
  }
}

func newServer(t *testing.T, opts... func(*testutils.ChannelOpts)) {
  chOpts := testutils.NewOpts().[...]
  for _, opt := range opts {
    opt(chOpts)
  }
  [...]
}
```

server_util_test.go
```diff
@@ -78,11 +83,21 @@ func (methodsT) echo() handler {
...
+		if span != nil {
+			if spanCtx, ok := span.Context().(jaeger.SpanContext); ok {
+				if spanCtx.IsDebug() {
+					ret = 0x01
+				}
+			} else {
+				ret = 0xfe
+			}
+		} else {
+			ret = 0xff
```
prashantv: no span in context case returning 0 seems like a good default

main.go
prashantv: yeah looks like the diff needs to be rebased off the latest `dev` branch

request_test.go
```diff
@@ -141,7 +141,7 @@ func TestGetHeaders(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		got, err := getHeaders(tt.inline, tt.file)
+		got, err := getHeaders(tt.inline, tt.file, nil)
```
prashantv: can we add unit tests for the `override` here

transport/tchannel_test.go
```diff
@@ -174,7 +174,6 @@ func TestTChannelCallSuccessRaw(t *testing.T) {
 	for _, tt := range tests {
 		var lastSpan uint64
 		testutils.RegisterFunc(svr, "echo", func(ctx context.Context, args *raw.Args) (*raw.Res, error) {
-			// assert.False(t, tchannel.CurrentSpan(ctx).TracingEnabled(), "Tracing should be disabled")
```
prashantv: intentional change? shouldn't this just be uncommenting?

transport/tchannel.go
```diff
@@ -105,20 +112,30 @@ func NewTChannel(opts TChannelOptions) (Transport, error) {
...
+	return t.tracer
+}
+
 func (t *tchan) Protocol() Protocol {
 	return TChannel
 }
 
 func (t *tchan) Call(ctx context.Context, r *Request) (*Response, error) {
-	call, err := t.sc.BeginCall(ctx, r.Method, t.callOptions)
+	req := *r
```
prashantv: can you add a comment for why we do this

main.go
```diff
@@ -309,6 +306,23 @@ type noEnveloper interface {
 	WithoutEnvelopes() encoding.Serializer
 }
 
+func getTracer(opts Options, out output) (opentracing.Tracer, io.Closer) {
+	var tracer opentracing.Tracer
+	var closer io.Closer
+	tracer = opentracing.NoopTracer{}
```
abhinav: nit: use var group and initialize tracer at declaration:

```
var (
  tracer opentracing.Tracer = opentracing.NoopTracer{}
  closer io.Closer
)
```

server_util_test.go
kriskowal: these cases were helpful in isolating test errors during development. I’ll scale it back down.

bench_method_test.go
kriskowal: This is a far flung effect of refactoring tracing. Tracing is no longer a transport level concern, since we no longer use trace sampling on the channel level. Consequently it becomes a request scoped concern and the logic to detect whether the caller was overridden had to be factored upward so I could validate benchmark options.

transport/interface.go
kriskowal: ✔

bench_method_test.go
prashantv: Ahh I see

integration_test.go
```diff
@@ -80,9 +82,64 @@ func argHandler(arg int32) (int32, error) {
...
+		return errors.New("trace context is not from jaeger")
+	}
+	if !spanCtx.IsDebug() {
+		return errors.New("span context is not configured to submit traces")
+	}
+
+	val := span.BaggageItem("baggagekey")
+	if val == "" {
+		return errors.New("missing baggage")
+	} else if val != "baggagevalue" {
```
prashantv: nit: no need for `else` since the above case returns,
```go
val := span.BaggageItem(..)
if val == "" {
  return errors.New("missing baggage")
}
if val != "baggagevalue" {
  return errors.New("unexpected baggage")
}
return nil
```

integration_test.go
```diff
@@ -80,9 +82,64 @@ func argHandler(arg int32) (int32, error) {
...
+		return errors.New("unexpected baggage")
+	}
+	return nil
+}
+
+func verifyThriftHeaders(ctx thrift.Context) error {
+	headers := ctx.Headers()
+	if val, ok := headers["headerkey"]; ok {
```
prashantv: nit: i'd write this the same as the `verifyBaggage` to reduce nesting
```go
val, ok := ctx.Headers()["headerkey"]
if !ok {
  return errors.New("missing header")
}
if val != "headervalue" {
  return errors.New("unexpected header")
}
return nil
```

integration_test.go
```diff
@@ -80,9 +82,64 @@ func argHandler(arg int32) (int32, error) {
...
+		}
+	} else {
+		return errors.New("missing header")
+	}
+
+	return nil
+}
+
+func verifyYARPCHeaders(reqMeta yarpc.ReqMeta) error {
+	headers := reqMeta.Headers()
```
prashantv: nit: can we instead have a single `veriyHeaders(headers map[string]string) error` method, and pass in the `ctx.Headers()` or `reqMeta.Headers()`? That way we can avoid the duplication here

transport/tchannel.go
```diff
@@ -75,7 +81,7 @@ func NewTChannel(opts TChannelOptions) (Transport, error) {
 	}
 
 	callerName := opts.SourceService
-	if cn, ok := opts.TransportOpts["cn"]; ok {
+	if cn, ok := opts.TransportOpts["cn"]; ok && cn != "" {
```
prashantv: nit: should we keep this as is -- it might be we want to test what happens if the `cn` header is empty in a request.

integration_test.go
kriskowal: YARPC does not expose Headers().Items().

prashantv: Ahh, I forgot that yarpc headers are not an actual map, this is fine then.

PR #122, opened 2016-10-19
Upgrade to TChannel 1.2.0

kriskowal:
r @prashantv

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -13,34 +13,21 @@ import:
   subpackages:
   - require
 - package: github.com/thriftrw/thriftrw-go
-  version: master
-  subpackages:
-  - ast
-  - compile
-  - envelope
-  - protocol
-  - wire
+  version: ~0.2.1
 - package: github.com/uber-go/atomic
   version: ^1.0.0
 - package: github.com/uber/tchannel-go
-  version: ^1.0.9
-  subpackages:
-  - thrift
+  version: ^1.2.0
 - package: golang.org/x/net
   version: master
   subpackages:
   - context
 - package: gopkg.in/yaml.v2
   version: master
+- package: github.com/opentracing/opentracing-go
+  version: ^1.0.0
 testImport:
 - package: github.com/apache/thrift
   version: master
-  subpackages:
-  - lib/go/thrift
 - package: github.com/yarpc/yarpc-go
-  version: master
-  subpackages:
-  - encoding/thrift
-  - transport
-  - transport/http
-  - transport/tchannel
+  version: ~0.2.1
--- a/glide.yaml
+++ b/glide.yaml
@@ -2,30 +2,24 @@ package: github.com/yarpc/yab
 import:
 - package: github.com/cactus/go-statsd-client
   version: master
-  subpackages:
-  - statsd
 - package: github.com/casimir/xdg-go
   version: master
 - package: github.com/jessevdk/go-flags
   version: master
 - package: github.com/stretchr/testify
   version: master
-  subpackages:
-  - require
 - package: github.com/thriftrw/thriftrw-go
   version: ~0.2.1
 - package: github.com/uber-go/atomic
-  version: ^1.0.0
+  version: ^1
 - package: github.com/uber/tchannel-go
-  version: ^1.2.0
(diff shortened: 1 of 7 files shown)


kriskowal:
This disables the trace test because there is no API for this in the upgraded TChannel. We can recover the feature and test it when we introduce a dependency on Jaeger.

transport_test.go
```diff
@@ -230,38 +230,38 @@ func TestGetTransportCallerName(t *testing.T) {
...
-
-		tchan, err := getTransport(opts, encoding.Raw)
-		require.NoError(t, err, "getTransport failed")
-		res, err := tchan.Call(ctx, &transport.Request{Method: "test"})
-		require.NoError(t, err, "transport.Call failed")
-
-		assert.Equal(t, tt.traceEnabled, res.Body[0], "TraceEnabled mismatch")
-	}
-}
+// func TestGetTransportTraceEnabled(t *testing.T) {
```
breerly: Is there a better way to disable the test than by commenting it out? For example, something like a:

```
t.Skip("reason this is being skipping")
```

transport/tchannel_test.go
```diff
@@ -174,7 +174,7 @@ func TestTChannelCallSuccessRaw(t *testing.T) {
 	for _, tt := range tests {
 		var lastSpan uint64
 		testutils.RegisterFunc(svr, "echo", func(ctx context.Context, args *raw.Args) (*raw.Res, error) {
-			assert.False(t, tchannel.CurrentSpan(ctx).TracingEnabled(), "Tracing should be disabled")
+			// assert.False(t, tchannel.CurrentSpan(ctx).TracingEnabled(), "Tracing should be disabled")
```
breerly: Seems like this should be removed entirely.

glide.yaml
```diff
@@ -13,34 +13,21 @@ import:
...
   - require
 - package: github.com/thriftrw/thriftrw-go
-  version: master
-  subpackages:
-  - ast
-  - compile
-  - envelope
-  - protocol
-  - wire
+  version: ~0.2.1
```
breerly: We should upgrade yab to the latest thriftrw - let's create a task for that.

glide.yaml
```diff
@@ -13,34 +13,21 @@ import:
...
-  - thrift
+  version: ^1.2.0
 - package: golang.org/x/net
   version: master
   subpackages:
   - context
 - package: gopkg.in/yaml.v2
   version: master
+- package: github.com/opentracing/opentracing-go
+  version: ^1.0.0
```
breerly: Just `^1` is fine.

glide.yaml
```diff
@@ -13,34 +13,21 @@ import:
...
-  - protocol
-  - wire
+  version: ~0.2.1
 - package: github.com/uber-go/atomic
   version: ^1.0.0
 - package: github.com/uber/tchannel-go
-  version: ^1.0.9
-  subpackages:
-  - thrift
+  version: ^1.2.0
```
breerly: Just `^1.2` is fine here.

glide.yaml
```diff
@@ -13,34 +13,21 @@ import:
...
-  subpackages:
-  - lib/go/thrift
 - package: github.com/yarpc/yarpc-go
-  version: master
-  subpackages:
-  - encoding/thrift
-  - transport
-  - transport/http
-  - transport/tchannel
+  version: ~0.2.1
```
breerly: I actually didn't realize yab had a dep on yarpc - let's create a ticket to upgrade to the latest version.

prashantv: Yep only for testing, we do a full integration test with YARPC to make sure yab works against it. We should definitely update to the latest.

transport/tchannel_test.go
prashantv: I think this will come back in the next PR when we add Jaeger?

glide.yaml
breerly: Perhaps we should move to `testImport:`?

prashantv: Yep, it's already there!

glide.yaml
kriskowal: #123

PR #124, opened 2016-10-27
Support set constants

abhinav:
@prashantv @kriskowal


prashantv:
lgtm

abhinav:
Added a failing test case too.

Issue #125, opened 2016-11-01
Prefix application headers over HTTP 


kriskowal:
yab does not currently prefix application headers with Rpc-Header- over HTTP. I propose that we repurpose TOpts --topts for raw HTTP headers and alias that with -T, so -T for transport headers, -H for application headers, -B for baggage headers.

PR #127, opened 2016-11-02
Make headers work consistent with YARPC for HTTP application headers

kriskowal:
YARPC expects application headers to be passed as HTTP headers with the Rpc-Header- prefix. This change makes the -H flag consistently work across TChannel and HTTP applications, and reuses the --topts transport header map (now also -T) for HTTP headers without a prefix. So -T consistently means "transport headers".

r @prashantv @abhinav

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -144,7 +144,7 @@ func getTransport(opts TransportOptions, encoding encoding.Encoding, tracer open
 			TargetService: opts.ServiceName,
 			HostPorts:     opts.HostPorts,
 			Encoding:      encoding.String(),
-			TransportOpts: opts.TransportOptions,
+			TransportOpts: opts.Headers,
 			Tracer:        tracer,
 		}
 		return transport.NewTChannel(topts)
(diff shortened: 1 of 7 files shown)


kriskowal:
Fixes #125

transport/http_test.go
```diff
@@ -190,7 +190,7 @@ func TestHTTPCall(t *testing.T) {
 	require.NoError(t, err, "Failed to create HTTP transport")
 
 	for _, tt := range tests {
-		tt.r.Headers = map[string]string{"hook": tt.hook}
+		tt.r.TransportHeaders = map[string]string{"hook": tt.hook}
```
prashantv: can we add test coverage for the `Headers` here

transport.go
```diff
@@ -144,7 +144,7 @@ func getTransport(opts TransportOptions, encoding encoding.Encoding, tracer open
 			TargetService: opts.ServiceName,
 			HostPorts:     opts.HostPorts,
 			Encoding:      encoding.String(),
-			TransportOpts: opts.TransportOptions,
+			TransportOpts: opts.Headers,
```
prashantv: should this also be renamed to `TransportHeaders`?

transport/http_test.go
kriskowal: ✔

transport.go
kriskowal: TransportOpts on this line is part of the transportutils interface, so not flexible on this end. opts is a TransportOptions object locally, TransportOptions.TransportHeaders would taste redundant.

PR #128, opened 2016-11-02
7.0.0

kriskowal:


prashantv:
Can you update the commit message of the last commit to say "0.7.0" and same for the title?

kriskowal:
@prashantv ✔️

prashantv:
It still shows 7.0.0 everywhere to me

PR #131, opened 2016-11-02
0.7.0

kriskowal:


man/yab.html
```diff
@@ -1,5 +1,5 @@
 <!-- Creator     : groff version 1.19.2 -->
-<!-- CreationDate: Wed Nov  2 15:00:07 2016 -->
+<!-- CreationDate: Wed Nov  2 16:37:24 2016 -->
```
prashantv: nit: if this is the only change, can probably revert this change

PR #134, opened 2016-11-12
Adds support for yab YAML templates

kriskowal:
This will allow you to specify RPCs as template files on the command line:

`yab --timeout 3s --peer 127.0.0.1:21300 -y time.yaml`

Example contents of `time.yaml`:

```
#!/usr/bin/env yab -y
service: "time"
method: "time::queryV2"
thrift: "path/to/some/files/time.idl"
headers:
    header1: value1
    header2: value2
request:
    location: 
        latitude: 37.7
        longitude: -122.4
        cityId: 1
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -243,11 +243,9 @@ func parseDefaultConfigs(parser *flags.Parser) error {
 
 func runWithOptions(opts Options, out output) {
 	if opts.ROpts.YamlTemplate != "" {
-		err := readYamlRequest(&opts)
-		if err != nil {
+		if err := readYamlRequest(&opts); err != nil {
 			out.Fatalf("Failed while reading yaml template: %v\n", err)
 		}
-
 	}
 
 	reqInput, err := getRequestInput(opts.ROpts.RequestJSON, opts.ROpts.RequestFile)
--- a/main.go
+++ b/main.go
@@ -242,6 +242,10 @@ func parseDefaultConfigs(parser *flags.Parser) error {
 }
 
 func runWithOptions(opts Options, out output) {
+	if opts.ROpts.YamlTemplate != "" {
+		readYamlRequest(&opts)
+	}
+
 	reqInput, err := getRequestInput(opts.ROpts.RequestJSON, opts.ROpts.RequestFile)
 	if err != nil {
 		out.Fatalf("Failed while loading body input: %v\n", err)
--- a/main.go
+++ b/main.go
@@ -243,7 +243,11 @@ func parseDefaultConfigs(parser *flags.Parser) error {
 
 func runWithOptions(opts Options, out output) {
 	if opts.ROpts.YamlTemplate != "" {
-		readYamlRequest(&opts)
+		err := readYamlRequest(&opts)
+		if err != nil {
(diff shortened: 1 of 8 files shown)


kriskowal:
This works in broad strokes. I’m having trouble threading some of the flags, like peers (even as hostports:). This may require some yaml tags on the options structs. Disregard my prior review. This works just fine as a shebang e.g., `#!/usr/bin/env yab -y` and should work fine for all new flags since rebase.

main.go
```diff
@@ -242,6 +242,14 @@ func parseDefaultConfigs(parser *flags.Parser) error {
 }
 
 func runWithOptions(opts Options, out output) {
+	if opts.ROpts.YamlTemplate != "" {
+		err := readYamlRequest(&opts)
+		if err != nil {
+			out.Fatalf("Failed while reading yaml template: %v\n", err)
+		}
+
```
prashantv: nit: remove blank line

main.go
```diff
@@ -242,6 +242,14 @@ func parseDefaultConfigs(parser *flags.Parser) error {
 }
 
 func runWithOptions(opts Options, out output) {
+	if opts.ROpts.YamlTemplate != "" {
+		err := readYamlRequest(&opts)
```
prashantv: nit: can inline into `if`

template.go
```diff
@@ -0,0 +1,50 @@
+package main
+
+import (
+	"io/ioutil"
+	"time"
+
+	"gopkg.in/yaml.v2"
+)
+
+type template struct {
```
prashantv: (something to consider for future) I wonder if we could reuse the `RequestOptions` somehow rather than duplicating the options

template.go
```diff
@@ -0,0 +1,50 @@
...
+	Request interface{}       `yaml:"request"`
+	Timeout time.Duration     `yaml:"timeout"`
+}
+
+func readYamlRequest(opts *Options) error {
+	t := template{}
+
+	bytes, err := ioutil.ReadFile(opts.ROpts.YamlTemplate)
+	if err != nil {
+		return err;
```
prashantv: no need for semicolons here and below (surprised our lint didn't catch this)

template.go
```diff
@@ -0,0 +1,50 @@
...
+
+	headers, err := yaml.Marshal(t.Headers)
+	if err != nil {
+		return err;
+	}
+
+	opts.ROpts.ThriftFile = t.Thrift
+	opts.ROpts.MethodName = t.Method
+	opts.TOpts.ServiceName = t.Service
+	opts.ROpts.HeadersJSON = string(headers)
```
prashantv: can we just set `t.Headers` and avoid the remarshal of the headers?

template.go
```diff
@@ -0,0 +1,50 @@
...
+	opts.ROpts.RequestJSON = string(body)
+	opts.ROpts.Timeout = timeMillisFlag(t.Timeout)
+
+	return nil
+}
```
prashantv: can we add unit tests for this file?

jsu1212: Done

PR #135, opened 2016-11-16
Flesh out properties of YAML templates

kriskowal:
This change adds support for further properties of YAML templates:

- peer
- peers
- peerlist (resolved relative to the template file)
- caller (overrides default unless benchmarked)
- shardkey
- routingkey
- routingdelegate
- baggage
- jaeger

It also rethreads headers without remarshaling. Arguments are merged over templated headers. There are some funny precedence issues in combination with HeadersJSON, but nothing that wasn’t already not quite right.

This also renames the template property "method" to "procedure" to be internally consistent with YARPC’s glossary. The inconsistency with yab’s -m is perhaps forgivable since the -m does not need to be specified and would otherwise clash with -p for peer. I could make both work, but my inclination is to just get it right before it’s released.

r @prashantv @jsu1212


template.go
```diff
@@ -22,18 +22,28 @@ package main
...
 type template struct {
-	Service string            `yaml:"service"`
-	Thrift  string            `yaml:"thrift"`
-	Method  string            `yaml:"method"`
-	Headers map[string]string `yaml:"headers"`
-	Request interface{}       `yaml:"request"`
-	Timeout time.Duration     `yaml:"timeout"`
+	Peers           []string          `yaml:"peers"`
+	Peer            string            `yaml:"peer"`
+	PeerList        string            `yaml:"peerlist"`
```
prashantv: should we use lower camel which is what most thrift fields will look like in the YAML file?

template.go
```diff
@@ -54,15 +64,51 @@ func readYamlRequest(opts *Options) error {
...
 
-	headers, err := yaml.Marshal(t.Headers)
-	if err != nil {
-		return err
+	if t.Peer != "" {
+		opts.TOpts.HostPorts = []string{t.Peer}
+	} else if t.Peers != nil {
+		opts.TOpts.HostPorts = t.Peers
+	} else if t.PeerList != "" {
+		opts.TOpts.HostPortFile = path.Join(path.Dir(opts.ROpts.YamlTemplate), t.PeerList)
```
prashantv: So unfortunately the path.Join doesn't work the way I was expecting, so we'll probably need to do is `IsAbs` check. We also shoudl use `filepath` instead of `path`, since `filepath` is OS specific, while `path` just assumes `/` (for those crazy systems where there's other separators)

template.go
```diff
@@ -54,15 +64,51 @@ func readYamlRequest(opts *Options) error {
...
+	}
+
+	// Header arguments have precedence over template headers.
+	if t.Headers != nil {
```
prashantv: nit: prefer length check instead `len(t.Headers) > 0`
same below for `ROpts.Headers`

template.go
```diff
@@ -54,15 +64,51 @@ func readYamlRequest(opts *Options) error {
...
+		if opts.ROpts.Headers == nil {
+			opts.ROpts.Headers = t.Headers
+		} else {
+			for k, v := range t.Headers {
+				if _, exists := opts.ROpts.Headers[k]; !exists {
+					opts.ROpts.Headers[k] = v
+				}
+			}
+		}
+	}
```
prashantv: let's move this into a separate function, and unit test it individually. I think we can simplify this method, but want to get unit test coverage to make sure we don't break anything:
```go
if opts.ROpts.Headers == nil {
  opts.ROpts.Headers = t.Headers
} else {
  for k, v := range t.Headers {
    if _, exists := opts.ROpts.Headers[k]; !exists {
      opts.ROpts.Headers[k] = v
    }   
  }
}

File: template.go [hunk 5]
```diff
@@ -54,15 +64,51 @@ func readYamlRequest(opts *Options) error {
...
+
+	if t.Jaeger {
+		opts.TOpts.Jaeger = true
+	}
+
+	// Baggage arguments have precedence over template headers.
+	if t.Baggage != nil {
```
prashantv: seems like the same logic as for headers

template.go
```diff
@@ -54,15 +64,51 @@ func readYamlRequest(opts *Options) error {
 		return err
 	}
 
-	headers, err := yaml.Marshal(t.Headers)
-	if err != nil {
-		return err
+	if t.Peer != "" {
+		opts.TOpts.HostPorts = []string{t.Peer}
+	} else if t.Peers != nil {
+		opts.TOpts.HostPorts = t.Peers
```
abhinav: You may want to check `len(t.Peers)` instead to make it play nicely with empty
slices in addition to nil.

kriskowal:
@prashantv ready for another look

kriskowal:
It would be great to get eyes on this from @jsu1212, whether these changes place undue burden on tools that may already be producing these templates. In particular, I worry about changing "method" to "procedure", and whether I should add a fallback to allow for either.

template.go
```diff
@@ -22,18 +22,28 @@ package main
...
-	Headers map[string]string `yaml:"headers"`
-	Request interface{}       `yaml:"request"`
-	Timeout time.Duration     `yaml:"timeout"`
+	Peers           []string          `yaml:"peers"`
+	Peer            string            `yaml:"peer"`
+	PeerList        string            `yaml:"peerList"`
+	Caller          string            `yaml:"caller"`
+	Service         string            `yaml:"service"`
+	Thrift          string            `yaml:"thrift"`
+	Procedure       string            `yaml:"procedure"`
```
prashantv: Why the rename? The option is currently named method everywhere else in yab.

If we want to change it to procedure, we should do that everywhere separately, and support `method` as an alias for `procedure` for backwards compatibility in flags.

kriskowal:
This is ready for review once again.

prashantv:
It seems like this change hasn't been rebased off dev correctly, it's still showing up a lot of the method -> procedure rename in the diff.

kriskowal:
Thanks, take another look @prashantv

testdata/templates/peer.yaml
```diff
@@ -0,0 +1,25 @@
...
+peer: 127.0.0.1:8080
+peers:
+- 127.0.0.1:8080
+- 127.0.0.1:8081
+peerList: peers.json
+caller: bar
+shardKey: sk
+routingKey: rk
+routingDelegate: rd
+thrift: testdata/templates/foo.thrift
```
prashantv: similar to `PeerList`, should this location be relative to the location of the yaml template (so this should become `foo.thrift`)

kriskowal: Absolutely. I’ll give that a gander.

PR #150, opened 2017-02-14
Add --no-jaeger flag to disable jaeger

prashantv:
If jaeger is enabled by default in a defaults.ini file, it cannot
currently be disabled due to
https://github.com/jessevdk/go-flags/issues/191

We need a better long term solution for boolean flags, but for now, add
a separate no-jaeger flag that will disable Jaeger.

### DIFF
--- a/main.go
+++ b/main.go
@@ -317,7 +317,7 @@ func getTracer(opts Options, out output) (opentracing.Tracer, io.Closer) {
 		tracer opentracing.Tracer = opentracing.NoopTracer{}
 		closer io.Closer
 	)
-	if opts.TOpts.Jaeger {
+	if opts.TOpts.Jaeger && !opts.TOpts.NoJaeger {
 		tracer, closer = jaeger.NewTracer(opts.TOpts.CallerName, jaeger.NewConstSampler(true), jaeger.NewNullReporter())
 	} else if len(opts.ROpts.Baggage) > 0 {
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -317,7 +317,7 @@ func getTracer(opts Options, out output) (opentracing.Tracer, io.Closer) {
 		tracer opentracing.Tracer = opentracing.NoopTracer{}
 		closer io.Closer
 	)
-	if opts.TOpts.Jaeger {
+	if opts.TOpts.Jaeger && !opts.TOpts.NoJaeger {
```
kriskowal: Consider override order. Flags on the CLI need to supersede flags from the config file. To make this work, we’ll have to push the mush heavier boulder up the hill—read the options once from config, once from flags, and merge them manually. The same applies for headers and baggage.

prashantv: Yeah this is an issue we have for all our flags -- I'm not sure what the right solution is for merging defaults and flags, but I think we should attack that in a separate PR.

For boolean flags specifically, I think if go-flags handled the `--no-FLAG` case, then it would help this quite a lot. The issue right now is that if a YAML file says "disable jaeger" (which is the default, so it would not really be a very useful option), then there's no way to enable Jaeger again.

This is more of a hacky solution so we can at least disable Jaeger, and then work on getting `--no-FLAG` into `go-flags` and that should be backwards compatible when we remove the flag here.

Issue #155, opened 2017-02-21
Support plugins for resolving custom URLs to a list of peers


prashantv:
Today, `-p` specifies a single peer, and for HTTP also happens to specify the URL template for the request including the protocol scheme of choice as well as the path. `-P` specifies a peer list, currently only supporting a file in YAML or JSON or flat text format. We propose to extend `-P` to accept a URI with multiple supported protocols including `file:` (to explicate the current behavior), and add `http:` to support grabbing the content via an HTTP request, and allow registering alternate protocols.

I think the right solution is to use the [plugin support](https://golang.org/doc/go1.8#plugin) in go1.8 to load plugins that can provide custom peer providers.
```go
type PeerProvider interface {
	Schemes() []string
	Resolve(context.Context, *url.URL) ([]string, error)
}
```

Resolve returns an array of alternative flags for `-p`, which over HTTP continues to serve as the URL template and can be used in combination with `-P`. In the absence of a provided `-p`, the selected peer template string is used both for the address and the template.

```
$ yab -P alt://blah/blah:http --health
```

kriskowal:
Landed in #156

PR #156, opened 2017-02-22
Support peer provider plugins

kriskowal:
This change factors out a pluggable peer provider registry and extends the meaning of the `-P` flag to be a URI, currently only supporting the implicit or explicit `file:/` protocol scheme.

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,6 +32,7 @@ import (
 	"time"
 
 	"github.com/yarpc/yab/encoding"
+	"github.com/yarpc/yab/peerprovider"
 	"github.com/yarpc/yab/transport"
 
 	"github.com/casimir/xdg-go"
@@ -241,6 +242,13 @@ func parseDefaultConfigs(parser *flags.Parser) error {
 }
 
 func runWithOptions(opts Options, out output) {
+	if opts.TOpts.HostPortFile == "?" {
+		for _, scheme := range peerprovider.Schemes() {
+			out.Printf("%s\n", scheme)
+		}
+		return
+	}
+
 	if opts.ROpts.YamlTemplate != "" {
 		if err := readYamlRequest(&opts); err != nil {
 			out.Fatalf("Failed while reading yaml template: %v\n", err)
--- a/peerprovider/interface.go
+++ b/peerprovider/interface.go
@@ -10,7 +10,19 @@ import (
 var registry = make(map[string]PeerProvider)
 
 func init() {
+	RegisterPeerProvider("", filePeerProvider{})
 	RegisterPeerProvider("file", filePeerProvider{})
+}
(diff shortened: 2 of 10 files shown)


kriskowal:
Fixes #155

peerprovider/interface.go
```diff
@@ -0,0 +1,46 @@
...
+	"golang.org/x/net/context"
+)
+
+var registry = make(map[string]PeerProvider)
+
+func init() {
+	RegisterPeerProvider("file", filePeerProvider{})
+}
+
+// Schemes returns supported peer provider protocol schemes.
```
prashantv: Can we add unit tests in this package for:
 - registering a new scheme
 - making sure it's returned in `Schemes`
 - ensuring it's used when `Resolve` is called with that scheme

transport.go
```diff
@@ -87,28 +84,39 @@ func ensureSameProtocol(hostPorts []string) (string, error) {
...
 
-		if len(hostPorts) == 0 {
-			return opts, errPeerRequired
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+		defer cancel()
+
+		// Paths are a strict subset of the URL file protocol scheme. We
+		// explicate the file protocol as a key into the peer provider
+		// registry.
+		if u.Scheme == "" {
```
prashantv: Could we register the file provider under the `""` scheme and have this work without this `if`?

main.go
```diff
@@ -241,6 +242,13 @@ func parseDefaultConfigs(parser *flags.Parser) error {
 }
 
 func runWithOptions(opts Options, out output) {
+	if opts.TOpts.HostPortFile == "?" {
```
prashantv: we can unit test this in, `main_test.go` has some integration tests that should have examples of how to do this.

kriskowal: addressed

peerprovider/interface_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	res, err := Resolve(context.Background(), mustParseURL("fake://dekaf"))
+	assert.NoError(t, err, "should resolve without error")
+	assert.Equal(t, res, f.res, "should resolve fake:// name")
+}
+
+func TestResolveError(t *testing.T) {
+	oldRegistry := registry
+	defer func() { registry = oldRegistry }()
+	registry = make(map[string]PeerProvider)
```
prashantv: optionally: make a helper for this,
```go

func stubRegister() func() {
  oldRegistry := registry
  registry = make(map[string]PeerProvider)
  return func() { registry = oldRegistry }()
}

func TestResolveError(t *testing.T) {
  defer stubRegistry()()

  [...]
}
```

PR #157, opened 2017-02-22
Support peer provision over HTTP

kriskowal:
This adds support for fetching peer lists over HTTP.

### DIFF
--- a/peerprovider/http.go
+++ b/peerprovider/http.go
@@ -1,0 +1,31 @@
+package peerprovider
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
+	"golang.org/x/net/context"
+	"golang.org/x/net/context/ctxhttp"
+)
+
+type httpPeerProvider struct{}
+
+func (httpPeerProvider) Resolve(ctx context.Context, url *url.URL) ([]string, error) {
+	resp, err := ctxhttp.Get(ctx, nil, url.String())
+	if err != nil {
+		return nil, fmt.Errorf("failed to read peer list over HTTP: %v", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("failed to read peer list over HTTP, status not OK: %v", http.StatusText(resp.StatusCode))
+	}
+	defer resp.Body.Close()
+
+	contents, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


peerprovider/http.go
```diff
@@ -0,0 +1,26 @@
...
+	"golang.org/x/net/context"
+)
+
+type httpPeerProvider struct{}
+
+func (httpPeerProvider) Resolve(ctx context.Context, url *url.URL) ([]string, error) {
+	resp, err := http.Get(url.String())
+	if err != nil {
+		return nil, fmt.Errorf("failed to read peer list over HTTP: %v", err)
+	}
```
prashantv: we should close the response body once we're done with it `defer resp.Body.Close()`

peerprovider/http.go
```diff
@@ -0,0 +1,26 @@
...
+	"net/http"
+	"net/url"
+
+	"golang.org/x/net/context"
+)
+
+type httpPeerProvider struct{}
+
+func (httpPeerProvider) Resolve(ctx context.Context, url *url.URL) ([]string, error) {
+	resp, err := http.Get(url.String())
```
prashantv: Can we use `ctxhttp` here so that it takes the context into account for timeouts and cancellation

kriskowal:
@prashantv This now has unit tests and is ready for review.

peerprovider/http_test.go
```diff
@@ -0,0 +1,44 @@
...
+		w.WriteHeader(http.StatusNotFound)
+	}))
+	defer svr.Close()
+
+	pp := httpPeerProvider{}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	_, err := pp.Resolve(ctx, mustParseURL(svr.URL))
+	assert.Error(t, err, "error resolving peers")
+}
```
prashantv: can we add another unit test where the `Get` fails? This can be done by picking a network address that will fail `e.g., 192.0.2.1`  which is for tests/examples?
http://smakd.potaroo.net/ietf/idref/rfc5735/index.html

maybe with a short timeout so the test doesn't take very long

peerprovider/http_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
```
prashantv: nit: delete blank between third party imports

PR #161, opened 2017-02-28
Fix parsing of URLs for peer list in YAML templates

prashantv:


template.go
```diff
@@ -57,6 +59,13 @@ func readYAMLRequest(opts *Options) error {
 	}
 
 	base := filepath.Dir(opts.ROpts.YamlTemplate)
+	base, err = filepath.Abs(base)
+	if err != nil {
+		return err
+	}
```
kriskowal: // Adding a final slash so that the base URL refers to a directory, unless the base is exactly "/".

template.go
```diff
@@ -57,6 +59,13 @@ func readYAMLRequest(opts *Options) error {
 	}
 
 	base := filepath.Dir(opts.ROpts.YamlTemplate)
+	base, err = filepath.Abs(base)
```
kriskowal: // Ensuring that the base directory is fully qualified. Otherwise, whether it is fully qualified depends on argv[0]. Must be fully qualified to be expressible as a `file:///` URL. Go’s URL parser does not recognize `file:path` as host-relative, not-CWD relative.

PR #174, opened 2017-03-09
Fix parsing of strings from values that look like ints/bools

prashantv:
Since YAML tries to guess the type of a value, a string argument with a
value that looks like an int or a bool will not be a string type. We
should coerce these values into a string to users are not forced to
explicitly quote strings in their YAML file.

### DIFF
--- a/thrift/types.go
+++ b/thrift/types.go
@@ -169,12 +169,14 @@ func parseBinary(value interface{}) ([]byte, error) {
 		return parseBinaryList(v)
 	case map[interface{}]interface{}:
 		return parseBinaryMap(v)
-	default:
+	case bool, int, int8, int16, int32, int64, uint64, float32, float64:
 		// Since YAML tries to guess the type of a value, if the value
 		// looks like an integer, YAML will use int types even though the
 		// user needs a string. E.g., `msg: true` will treat "true" as a bool,
 		// but `msg: t` will be treat "t"" as a string. So instead of throwing
-		// an error at the user, convert the value we have to a string.
+		// an error at the user, coerce known YAML scalae types to a string.
 		return []byte(fmt.Sprint(v)), nil
+	default:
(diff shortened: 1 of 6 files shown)


thrift/types.go
```diff
@@ -170,6 +170,11 @@ func parseBinary(value interface{}) ([]byte, error) {
 	case map[interface{}]interface{}:
 		return parseBinaryMap(v)
 	default:
-		return nil, fmt.Errorf("cannot parse binary/string from %T: %v", value, v)
+		// Since YAML tries to guess the type of a value, if the value
+		// looks like an integer, YAML will use int types even though the
+		// user needs a string. E.g., `msg: true` will treat "true" as a bool,
+		// but `msg: t` will be treat "t"" as a string. So instead of throwing
+		// an error at the user, convert the value we have to a string.
+		return []byte(fmt.Sprint(v)), nil
```
abhinav: Maybe we should have this behavior for scalars only. Otherwise,

    foo: [1, 2, 3]  # will translate to "{1 2 3}"

prashantv: 👍

thrift/types.go
```diff
@@ -169,6 +169,13 @@ func parseBinary(value interface{}) ([]byte, error) {
 		return parseBinaryList(v)
 	case map[interface{}]interface{}:
 		return parseBinaryMap(v)
+	case bool, int, int8, int16, int32, int64, uint64, float32, float64:
+		// Since YAML tries to guess the type of a value, if the value
+		// looks like an integer, YAML will use int types even though the
+		// user needs a string. E.g., `msg: true` will treat "true" as a bool,
+		// but `msg: t` will be treat "t"" as a string. So instead of throwing
+		// an error at the user, coerce known YAML scalae types to a string.
```
abhinav: s/scalae/scalar/

PR #179, opened 2017-03-13
Add support for template arguments

prashantv:
Template arguments are of the format ${argname:fallback}. Arguments can
be specified on the commandline using `--arg` or `-A`. Arguments are
parsed as YAML values, and allow more complex types (e.g., lists) to be
specified as arguments.

We only use template arguments for request bodies currently.

Argument parsing is based on @abhinav's work for YARPC configuration.


templateargs/argparser/arg.rl
```diff
@@ -0,0 +1,64 @@
...
+    write data;
+}%%
+
+// Var is the parsed envvar string.
+type Var struct {
+  Name  string
+  Value string
+}
+
+// Parse parses strings in the format, ${NAME} and ${NAME:VALUE}.
```
kriskowal: Add a Lit struct and return LitOrVar. Lit should produce a string with `\$ => $` and `\\ => \`.

templateargs/process.go
```diff
@@ -0,0 +1,103 @@
...
+		if err := yaml.Unmarshal([]byte(v), &unmarshalled); err != nil {
+			return nil, fmt.Errorf("failed to process %q with value %q: %v", k, v, err)
+		}
+		argsUnmarshalled[k] = unmarshalled
+	}
+
+	return processMap(req, argsUnmarshalled)
+}
+
+func processString(v string, args map[string]interface{}) (interface{}, error) {
```
kriskowal: This function should unconditionally parse the string in its DSL, including escaping `\$` and `\\`. The grammar should return a simple AST with either a Lit or a Var term, and this function should just switch on that.

This leaves open the possibility of extending the grammar to support string interpolation like `${host}:${port}`. We can forbid this for now.

testdata/templates/bad-arg.yaml
```diff
@@ -0,0 +1,3 @@
+service: foo
+request:
+  name: ${user:foo
```
kriskowal: Needs test for `\${user}`. Would be good to test `\\`, `\$`.

kriskowal: Might be good to test requests with embedded non-string arguments, e.g., as in `yab -A peer-list:$(cat peers.yaml)`

templateargs/process.go
```diff
@@ -0,0 +1,103 @@
...
+	// If a string starts with \, avoid processing the value and strip the prefix.
+	if strings.HasPrefix(v, `\`) {
+		return v[1:], nil
+	}
+
+	// Anything that starts with "${ " should be processed as a template arg.
+	if !strings.HasPrefix(v, "${") {
```
abhinav: and ends with `}`?

testdata/templates/bad-arg.yaml
prashantv: We cover a lot of the edge cases such as `\\` and `\$` in `TestProcessString` and `TestProcessMap` already. but added a couple more to `args.yaml` -- it has a fallback which is not a string, a list replacement, and testing for escapes.

testdata/templates/args.yaml
```diff
@@ -0,0 +1,8 @@
+service: foo
+headers:
+  header1: ${user:foo}
```
prashantv: Note: this is not replaced at the moment, template arguments are used for request bodies.

Before 1.0, we may want to think about processing everything in the YAML file? We would probably decode the template to a `map`, use `templateargs.ProcessMap`, then pass it through `mapdecode`. Thoughts @kriskowal @abhinav

abhinav: Sounds reasonable. Although `mapdecode` expects to decode into a known struct shape, not dynamic Thrift shapes.

prashantv: Yep, we want to use it to load into the `Template` struct (known Go shape). Won't affect the body which is still a map.

options.go
```diff
@@ -52,6 +52,7 @@ type RequestOptions struct {
 	Health       bool              `long:"health" description:"Hit the health endpoint, Meta::health"`
 	Timeout      timeMillisFlag    `long:"timeout" default:"1s" description:"The timeout for each request. E.g., 100ms, 0.5s, 1s. If no unit is specified, milliseconds are assumed."`
 	YamlTemplate string            `short:"y" long:"yaml-template" description:"Send a tchannel request specified by a yaml template"`
+	TemplateArgs map[string]string `short:"A" long:"arg" description:"A list of key-value template arguments"`
```
prashantv: 👍

PR #183, opened 2017-04-05
Allow disabling thrift envelope in YAML template

prashantv:
Currently, Thrift envelopes must be disabled at the command line. Allow
disabling the envelope from the YAML template.

### DIFF
--- a/testdata/templates/foo.yaml
+++ b/testdata/templates/foo.yaml
@@ -6,6 +6,7 @@ routingKey: rk
 routingDelegate: rd
 thrift: foo.thrift
 timeout: 4.5s
+disableThriftEnvelope: false
 headers:
     header1: from template
     header3: overridden by Headers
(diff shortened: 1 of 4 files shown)


testdata/templates/foo.yaml
```diff
@@ -6,6 +6,7 @@ routingKey: rk
 routingDelegate: rd
 thrift: foo.thrift
 timeout: 4.5s
+disableThriftEnvelope: false
```
abhinav: Isn't `disableThriftEnvelope: false` the default? Possibly add a separate YAML with it set to true?

prashantv: I wanted to make sure it was overriding any even when it's false, but decided to just make this `true` and `TestDisableThriftEnvelopeOverride` for testing the overrides.

PR #184, opened 2017-04-07
Add documentation on YAML templates.

prashantv:


README.md
```diff
@@ -54,6 +55,10 @@ Request Options:
       --timeout=                 The timeout for each request. E.g., 100ms,
                                  0.5s, 1s. If no unit is specified,
                                  milliseconds are assumed. (default: 1s)
+  -y, --yaml-template=           Send a tchannel request specified by a yaml
```
abhinav: all caps YAML

doc.go
```diff
@@ -63,6 +63,25 @@ Or it can be loaded from a file using -f or --file:
 
 	$ yab -p localhost:9787 -t kv.thrift kv KeyValue::Get --file req.yaml
 
+Request options can also be specified in a YAML file, e.g., get.yab:
+	#!/usr/bin/env yab -y
```
abhinav: Based on other examples, add empty line before code block.

abhinav: Also, I wouldn't introduce the shebang first. Just the `yab -y foo.yaml` format and then say, `-y` reads input from stdin if a filename wasn't provided so it can be used as a `shebang` for executable YAML files like so. Otherwise the assumption will be that the templates must have the shebang.

doc.go
```diff
@@ -63,6 +63,25 @@ Or it can be loaded from a file using -f or --file:
...
+
+	service: kv
+	peer: localhost:9787
+	method: KeyValue::get
+	thrift: kv.thrift
+	request:
+	  key: "${key:hello}"
+
+The above YAML file represents the same request as the command line, and can
```
abhinav: "as the command line" => "as the command before"?

doc.go
```diff
@@ -63,6 +63,25 @@ Or it can be loaded from a file using -f or --file:
...
 
+Request options can also be specified in a YAML file, e.g., get.yab:
+	#!/usr/bin/env yab -y
+
+	service: kv
+	peer: localhost:9787
+	method: KeyValue::get
+	thrift: kv.thrift
+	request:
+	  key: "${key:hello}"
```
abhinav: nit: perhaps don't introduce this format until you introduce arguments later?

README.md
prashantv: 👍

doc.go
prashantv: 👍

doc.go
prashantv: 👍

doc.go
prashantv: 👍

doc.go
dentafrice: Note, this doesn't seem to work on Jessie.

```
Distributor ID: Debian
Description:    Debian GNU/Linux 8.7 (jessie)
Release:        8.7
Codename:       jessie
```

```
$ ./requests/testEmail/transactional.yab                                                                                                    
/usr/bin/env: yab -y: No such file or directory
```

prashantv: Yeah unfortunately there's no nice solution for Linux, filed https://github.com/yarpc/yab/issues/198

We have a couple of options, neither of which is great (using a separate `yaby` binary which assumes `-y`, or having `yab` detect with the first argument is a yaml template)

PR #185, opened 2017-04-07
Release version 0.10.0

prashantv:
* Support disabling Thrift envelopes from YAML templates.
* Add template argument support, which allows passing parameters
  to YAML templates.


CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+# 0.9.1
+
+* Support disabling Thrift envelopes from YAML templates.
+* Add template argument support, which allows passing parameters
+  to YAML templates.
```
abhinav: These are features. They merit a minor release.

prashantv: Bumped to 0.10.0 (hope we don't need too many more minors before 1.0)

PR #193, opened 2017-04-26
Use go.uber.org/atomic

billf:


kriskowal:
This PR could use a short description for motivation. You however, already have your stamps. Carry on.

template.go
```diff
@@ -166,7 +166,7 @@ func overrideParam(s *string, newS string) {
 	}
 }
 
-func UnmarshalTemplate(bytes []byte) (*template, error) {
```
prashantv: since this is a `main.go` and can't be imported, should be OK since it won't break anything.

PR #195, opened 2017-04-29
Adding map to blacklist caller names, tcurl to start

### LINKED ISSUE #180: Block "tcurl" callers
People use the `--caller` option to emulate `tcurl`'s caller name, but this masks who is actually making the call. Let's prevent it.

jeffbean:
Fixes #180  and allows us to blacklist other caller names as well. 

Thoughts while doing this is we could just append `USER` instead of block. Also we could do a `strings.contains` to get caller names with `tcurl` anywhere in the name. 

@blampe

### DIFF
--- a/main.go
+++ b/main.go
@@ -79,7 +79,7 @@ func fromPositional(args []string, index int, s *string) bool {
 
 func main() {
 	log.SetFlags(0)
-	parseAndRun(consoleOutput{os.Stdout, os.Stderr})
+	parseAndRun(consoleOutput{os.Stdout})
 }
 
 var errExit = errors.New("sentinel error used to exit cleanly")
@@ -286,7 +286,7 @@ func runWithOptions(opts Options, out output) {
 	if opts.TOpts.CallerName != "" {
 		if _, ok := warningCallerNames[opts.TOpts.CallerName]; ok {
 			// TODO: when logger is hooked up this should use the WARN level message
-			out.Warnf("WARNING: Deprecated caller name: [%v]\nPlease change the caller name as it will be blocked in the next release.", opts.TOpts.CallerName)
+			out.Warnf("WARNING: Deprecated caller name: %q Please change the caller name as it will be blocked in the next release.\n", opts.TOpts.CallerName)
 		}
 		if _, ok := blockedCallerNames[opts.TOpts.CallerName]; ok {
 			out.Fatalf("Disallowed caller name: %v", opts.TOpts.CallerName)
--- a/main.go
+++ b/main.go
@@ -45,8 +45,10 @@ import (
 
 var (
 	errHealthAndProcedure = errors.New("cannot specify procedure and use --health")
+
 	// map of caller names we do not want to be used.
-	disallowedCallerNames = map[string]bool{"tcurl": true}
+	warningCallerNames = map[string]struct{}{"tcurl": struct{}{}}
+	blockedCallerNames = map[string]struct{}{}
 )
 
 func findGroup(parser *flags.Parser, group string) *flags.Group {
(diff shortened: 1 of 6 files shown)


blampe:
This will need to be a delicate roll out -- I suggest a path that lets us emit a warning (possibly customizable with a pointer to internal docs?), give people some time to adjust, and then elevate it to an error.

main.go
```diff
@@ -43,7 +43,11 @@ import (
 	"github.com/uber/tchannel-go"
 )
 
-var errHealthAndProcedure = errors.New("cannot specify procedure and use --health")
+var (
+	errHealthAndProcedure = errors.New("cannot specify procedure and use --health")
+	// map of caller names we do not want to be used.
+	disallowedCallerNames = map[string]bool{"tcurl": true}
```
abhinav: optional: prefer `map[T]struct{}` for sets rather than `map[T]bool` to avoid accidentally using the value inside the map rather than the `ok`.

(It's easy to accidentally do, `if v := disallowedCallerNames[name]; v { .. }`.)

main.go
```diff
@@ -278,6 +282,9 @@ func runWithOptions(opts Options, out output) {
 	}
 
 	if opts.TOpts.CallerName != "" {
+		if val, ok := disallowedCallerNames[opts.TOpts.CallerName]; ok {
+			out.Fatalf("Disallowed caller name: %v", val)
```
abhinav: This will print "Disallowed caller name: true". You meant to use `opts.TOpts.CallerName`.

main_test.go
```diff
@@ -145,6 +145,18 @@ func TestRunWithOptions(t *testing.T) {
...
+			desc: "Fail to allow caller names from a map",
+			opts: Options{
+				ROpts: validRequestOpts,
+				TOpts: TransportOptions{
+					ServiceName: "foo",
+					Peers:       []string{echoServer(t, fooMethod, nil)},
+					CallerName:  "tcurl",
+				},
+			},
+			errMsg: "Disallowed caller",
```
abhinav: Change to

    errMsg: "Disallowed caller name: tcurl"

jeffbean:
@blampe Do we have docs you want to point to? I am setting up two lists for warning and blocked and can print all you want.

main_test.go
```diff
@@ -60,12 +60,17 @@ func TestRunWithOptions(t *testing.T) {
 		Procedure:  fooMethod,
 	}
 
+	blockedCallerNames = map[string]struct{}{
```
prashantv: should we be resetting this back to the original value?
options:
- Don't set this value at all, and assume "tcurl" is in that list
- Set this value early in the code (maybe in `init` in the `main_test.go` so we know all tests use this value)
- Reset back to original.

I'm learning to one of the earlier options.

main_test.go
```diff
@@ -176,6 +213,10 @@ func TestRunWithOptions(t *testing.T) {
 			continue
 		}
 
+		if tt.warnMsg != "" {
+			assert.Contains(t, warnBuf.String(), tt.warnMsg)
+		}
```
prashantv: else, ensure that the `warnBuf` is empty

output.go
```diff
@@ -32,16 +32,22 @@ type output interface {
...
 	log.Fatalf(format, args...)
 }
 
-func (consoleOutput) Printf(format string, args ...interface{}) {
+func (c consoleOutput) Printf(format string, args ...interface{}) {
 	fmt.Printf(format, args...)
 }
+
+func (c consoleOutput) Warnf(format string, args ...interface{}) {
+	fmt.Fprintf(c.errFile, format, args...)
```
prashantv: realized we're not using the actual values in the structs, console output always prints to console, so remove the fields, just print to `os.Stderr`

main_test.go
jeffbean: Good point, i added to `init` since these will change in practice.

PR #196, opened 2017-04-29
Progress bar for benchmark tests

### LINKED ISSUE #57: Have some sort of progress indicator or status update as the benchmark is running

jeffbean:
Fixes #57 

Prints a Progress bar to the console based on request count. Using library that does atomic increments. Other notes is it uses the terminal width and we are not setting it manually.

Output:

`yab moe --health --rps 100 -d 10s `
```
{
  "body": {
    "result": {
      "ok": true
    }
  },
  "ok": true,
  "trace": "4802f563e101ee42"
}

Benchmark parameters:
  CPUs:            24
  Connections:     48
  Concurrency:     1
  Max requests:    1000
  Max duration:    10s
  Max RPS:         100
 1000 / 1000 [============================================================================================================================] 100.00% 10s
Benchmark finished
Latencies:
  0.5000: 785.466µs
  0.9000: 1.006528ms
  0.9500: 1.189203ms
  0.9900: 9.582816ms
  0.9990: 57.667579ms
  0.9995: 59.614299ms
  1.0000: 61.56102ms
Elapsed time:      10s
Total requests:    1000
RPS:               100.00
```

Test without finishing total requests `yab moe --health --rps 20000 -d 10s`
```
{
  "body": {
    "result": {
      "ok": true
    }
  },
  "ok": true,
  "trace": "78915b6b9d43168c"
}

Benchmark parameters:
  CPUs:            24
  Connections:     48
  Concurrency:     1
  Max requests:    200000
  Max duration:    10s
  Max RPS:         20000
 98210 / 200000 [===========================================================>-------------------------------------------------------------]  49.10% 10s
Benchmark finished
Latencies:
  0.5000: 565.675µs
  0.9000: 1.575297ms
  0.9500: 13.889164ms
  0.9900: 87.541209ms
  0.9990: 91.826491ms
  0.9995: 93.020593ms
  1.0000: 101.400753ms
Elapsed time:      10.045s
Total requests:    98210
RPS:               9776.14

```

### DIFF
--- a/benchmark.go
+++ b/benchmark.go
@@ -31,6 +31,8 @@ import (
 	"github.com/yarpc/yab/limiter"
 	"github.com/yarpc/yab/statsd"
 	"github.com/yarpc/yab/transport"
+
+	pb "gopkg.in/cheggaaa/pb.v1"
 )
 
 var (
@@ -75,13 +77,14 @@ func (o BenchmarkOptions) enabled() bool {
 	return o.MaxDuration != 0 || o.MaxRequests != 0
 }
 
-func runWorker(t transport.Transport, m benchmarkMethod, s *benchmarkState, run *limiter.Run) {
+func runWorker(t transport.Transport, m benchmarkMethod, s *benchmarkState, run *limiter.Run, pb *pb.ProgressBar) {
 	for cur := run; cur.More(); {
 		latency, err := m.call(t)
 		if err != nil {
 			s.recordError(err)
 			continue
 		}
+		pb.Increment()
 
 		s.recordLatency(latency)
 	}
@@ -132,6 +135,8 @@ func runBenchmark(out output, allOpts Options, m benchmarkMethod) {
 		states[i] = newBenchmarkState(statter)
 	}
 
+	progressBar := pb.StartNew(opts.MaxRequests)
+
 	run := limiter.New(opts.MaxRequests, opts.RPS, opts.MaxDuration)
 	stopOnInterrupt(out, run)
 
@@ -143,7 +148,7 @@ func runBenchmark(out output, allOpts Options, m benchmarkMethod) {
 			wg.Add(1)
 			go func(c transport.Transport) {
 				defer wg.Done()
-				runWorker(c, m, state, run)
+				runWorker(c, m, state, run, progressBar)
 			}(c)
 		}
 	}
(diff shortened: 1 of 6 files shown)


benchmark.go
```diff
@@ -100,6 +104,7 @@ func runBenchmark(out output, allOpts Options, m benchmarkMethod) {
 	if opts.RPS > 0 && opts.MaxDuration > 0 {
 		// The RPS * duration in seconds may cap opts.MaxRequests.
 		rpsMax := int(float64(opts.RPS) * opts.MaxDuration.Seconds())
+		// FIXME: if we are using small time like 500ms the calculation is still 0
```
prashantv: hmm why is that? `Seconds` is a float, so it should be a positive integer

benchmark.go
```diff
@@ -132,6 +137,9 @@ func runBenchmark(out output, allOpts Options, m benchmarkMethod) {
 		states[i] = newBenchmarkState(statter)
 	}
 
+	progressBar := pb.StartNew(opts.MaxRequests)
```
prashantv: There's 2 main types of benchmarks:
 - by total number of requests
 - by total duration

we're only taking total number of requests into account but we should also think about duration.

we could progress bars for both if both are specified?

glide.yaml
```diff
@@ -16,6 +16,7 @@ import:
   version: ^1
 - package: github.com/uber/jaeger-client-go
   version: ^1
+- package: gopkg.in/cheggaaa/pb.v1
```
billf: the github repo supports semver, maybe better to go directly with that and `^1`

benchmark.go
billf: if RPS and MaxDuration are both low enough, the product of multiplying them together will be a fraction less than `1`. e.g. `10 * time.Millisecond.Seconds() = 0.01` and the cast will truncate `int(0.01) = 0`

https://play.golang.org/p/VrpN9zrZW4

benchmark.go
jeffbean: Thanks for this. Found a race testing this with Duration. After playing around with the library I forked it and might try to upstream more fixes. Seems small enough and active that this can wait for the upstream to take my fix or we can use my fork and improve the library for our needs as well. 

https://github.com/cheggaaa/pb/pull/103

peter-edge:
@jeffbean what is the status on this?

jeffbean:
This lib keeps introducing data races. i will come back to this when I fix more upstream issues.

PR #197, opened 2017-04-30
Change so Encoding without procedure shows explorable methods if possible

### LINKED ISSUE #192: When procedure is missing, list all available procedures
Currently we just return an error:
```
$ yab -p 127.0.0.1:5437 moe -t /usr/share/uber-idl/code.uber.internal/rt/moe/moe.thrift
Failed while parsing input: no procedure specified, specify --procedure [procedure]
```

Instead we should list all available procedures

jeffbean:
Fixes #192 

Prints out the respective Methods or Services available based on `Encoding` and `ThirftFile`

Its quite generous in that if we specify JSON but also specify the Thrift file it will still explore the Services and Methods in the file. Let me know if we should limit to thrift encoding only.

### DIFF
--- a/encoding/not_found.go
+++ b/encoding/not_found.go
@@ -25,11 +25,12 @@ import (
 	"strings"
 )
 
-type notFoundError struct {
-	msg       string
-	available []string
+// NotFoundError is a helpful responce error give better thrift encoding feeback
+type NotFoundError struct {
+	Msg       string
+	Available []string
 }
 
-func (e notFoundError) Error() string {
-	return fmt.Sprintf("%v\n\t%v", e.msg, strings.Join(e.available, "\n\t"))
+func (e NotFoundError) Error() string {
+	return fmt.Sprintf("%v\n\t%v", e.Msg, strings.Join(e.Available, "\n\t"))
 }
--- a/request_test.go
+++ b/request_test.go
@@ -178,6 +178,7 @@ func TestGetHeaders(t *testing.T) {
 
 func TestNewSerializer(t *testing.T) {
 	tests := []struct {
+		desc     string
 		encoding encoding.Encoding
 		opts     RequestOptions
 		want     encoding.Encoding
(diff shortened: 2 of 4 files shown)


encoding/not_found.go
```diff
@@ -25,11 +25,12 @@ import (
 	"strings"
 )
 
-type notFoundError struct {
-	msg       string
-	available []string
+// NotFoundError is a helpful responce error give better thrift encoding feeback
```
prashantv: nit: typo, responce -> response

request_test.go
```diff
@@ -233,39 +234,94 @@ func TestNewSerializer(t *testing.T) {
...
-			opts:     RequestOptions{Procedure: "procedure"},
-			want:     encoding.JSON,
+			opts:     RequestOptions{ThriftFile: validThrift},
+			wantErr: encoding.NotFoundError{
+				Msg:       "no Thrift service specified, specify --method Service::Method, available services:",
+				Available: []string{"Simple"},
+			},
 		},
 		{
+			desc:     "Raw encoder with a thrift file and no service",
```
prashantv: hmm what does it mean to use raw encoding with a thrift file? thrift files can only be used with a thrift encoding

request_test.go
```diff
@@ -233,39 +234,94 @@ func TestNewSerializer(t *testing.T) {
...
+			desc:     "Thrift encoder with a thrift file and no method",
+			encoding: encoding.Thrift,
+			opts:     RequestOptions{ThriftFile: validThrift, Procedure: "Simple"},
+			wantErr: encoding.NotFoundError{
+				Msg:       "no Thrift method specified, specify --method Service::Method, available methods:",
+				Available: []string{"bar", "foo", "thriftEx", "withDefault"},
+			},
+		},
+		{
+			desc:     "JSON encoder with a thrift file and no method",
```
prashantv: same as earlier, not sure if we need this.

request_test.go
```diff
@@ -233,39 +234,94 @@ func TestNewSerializer(t *testing.T) {
...
+			encoding: encoding.JSON,
+			opts:     RequestOptions{ThriftFile: validThrift, Procedure: "Simple"},
+			wantErr: encoding.NotFoundError{
+				Msg:       "no Thrift method specified, specify --method Service::Method, available methods:",
+				Available: []string{"bar", "foo", "thriftEx", "withDefault"},
+			},
+		},
+		{
+			desc:     "Raw encoder with a thrift file and no method",
```
prashantv: raw + thrift isn't required

jeffbean:
Close in preference to #204

PR #204, opened 2017-06-01
List Thrift services when procedure is empty

### LINKED ISSUE #192: When procedure is missing, list all available procedures

prashantv:
Fixes #192.

### DIFF
--- a/encoding/thrift_request.go
+++ b/encoding/thrift_request.go
@@ -35,7 +35,13 @@ import (
 
 const _multiplexedSeparator = ":"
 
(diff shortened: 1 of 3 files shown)


jeffbean:
I will abandon my PR in preference to this one.

PR #205, opened 2017-06-08
Set RPC-Thrift-Envelope HTTP header for Thrift requests

prashantv:
See yarpc/yarpc-go/issues/741 for more details. This is the first step in making
yab work with YARPC for HTTP+Thrift transparently regardless of whether the request is enveloped or not.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -34,6 +34,7 @@ import (
 	"github.com/yarpc/yab/testdata/gen-go/integration"
 	yintegration "github.com/yarpc/yab/testdata/yarpc/integration"
 	"github.com/yarpc/yab/testdata/yarpc/integration/fooserver"
+	"github.com/yarpc/yab/transport"
 
 	athrift "github.com/apache/thrift/lib/go/thrift"
 	"github.com/opentracing/opentracing-go"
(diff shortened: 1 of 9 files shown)


integration_test.go
```diff
@@ -263,6 +264,44 @@ func TestIntegrationProtocols(t *testing.T) {
...
+
+	for _, tt := range tests {
+
+		var called bool
+		server := httptest.NewServer(http.HandlerFunc(func(_ http.ResponseWriter, r *http.Request) {
+			called = true
+			envelope := r.Header.Get(transport.HTTPThriftEnvelopeheader)
+			assert.Equal(t, tt.wantEnvelopeHeader, envelope)
+		}))
+		defer server.Close()
```
abhinav: nit: Because this isn't a function, all the servers will keep running until the test exits. I would push this into an anonymous function.

main_test.go
```diff
@@ -786,11 +787,17 @@ func TestWithTransportSerializer(t *testing.T) {
 				Type:  wire.Call,
 				Value: wire.NewValueStruct(wire.Struct{}),
 			}),
+			wantTransportHeaders: map[string]string{
+				"RPC-Thrift-Envelope": "true",
```
abhinav: nit: pretty sure that the canonical form of the header key (as defined by net/http) is `Rpc-Thrift-Envelope`.

You could also just use https://golang.org/pkg/net/http/#CanonicalHeaderKey

integration_test.go
kriskowal: May as well add a t.Run wrapper in the loop.

peter-edge:
@prashantv what is the status on this?

prashantv:
I didn't want to merge this without making sure we were doing the same thing in YARPC. I had a WIP PR in YARPC:
https://github.com/yarpc/yarpc-go/pull/1097

The YARPC request object does not have the flexibility to implement this, from the above PR:
```
We need to come up with a good approach to pass some information from the encoding to the transport on the request. I added an arbitrary TransportHeaders map[string]string, but I'm not sure if that's the right approach.
```

I'm not planning to work on this in YARPC till the above issue is resolved.

PR #210, opened 2017-08-28
Fix timeouts in templates being ignored

### LINKED ISSUE #208: yab: Timeout specified in template is not used
Even when a timeout is specified in the YAML file, it is ignored.

prashantv:
We parse YAML templates before flags so that users can override template
args using flags. This didn't work for timeout since we used the
`default` flag, which will always set the option in the struct if the
flag was not specified. This caused the timeout value read from YAML to
be overridden.

We can avoid go-flags from setting the value by setting the default
value on the struct. go-flags displayed this default as an integer, so
use default-mask to display it as a human readable string.

Fixes #208

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 Changelog
 =========
 
+# 0.10.2 (Unreleased)
+* Fix timeouts specified in tempaltes being ignored.
+
 # 0.10.1 (2017-04-24)
 * Fix for HTTP not respecting context timeout.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 Changelog
 =========
 
+# 0.10.2 (Unreleased)
+* Fix timeouts specified in tempaltes being ignored.
```
lrewega: s/tempaltes/templates/

PR #215, opened 2017-09-06
Support yab as a shebang without -y

### LINKED ISSUE #198: Yab templates with `yab -y` shebang does not work on Linux
We suggest yab templates use the following shebang `#!/usr/bin/env yab -y`, which works on Darwin, but not Linux.

prashantv:
Shebangs cannot contain flags on Linux, so our recommendation to use
```
#!/usr/bin/env yab -y
```
doesn't work on Linux.

Instead, have yab look at the first argument, and if it's a valid file
ending in .yab, then don't require -y. This is a little bit of "magic"
but will make the user experience better (so users don't have to
remember to use a separate script called yaby or something).

Fixes #198.


main.go
```diff
@@ -123,6 +123,13 @@ yab includes a full man page (man yab), which is also available online: http://y
 		return nil, fmt.Errorf("error reading defaults: %v", err)
 	}
 
+	// Check if the first argument is a yab template. This is to support using
+	// yab as a shebang, since flags aren't supported in shebangs.
+	if len(args) > 0 && isYabTemplate(args[0]) {
+		fmt.Println("found yab template at arg", args)
```
abhinav: Did you mean to leave the Println in?

prashantv: Nope, thanks for catching!

PR #220, opened 2017-09-22
Add transport middleware plugin

APTy:
This adds the `TransportMiddleware` interface, which may be implemented to hook into and modify outgoing `yab` requests.


APTy:
@prashantv @abhinav any thoughts on this or is anyone else available for review?

main.go
```diff
@@ -326,18 +325,14 @@ func runWithOptions(opts Options, out output, logger *zap.Logger) {
 	serializer = withTransportSerializer(transport.Protocol(), serializer, opts.ROpts)
 
 	// req is the transport.Request that will be used to make a call.
-	req, err := serializer.Request(reqInput)
+	rawReq, err := serializer.Request(reqInput)
```
abhinav: nit: any reason to call this `rawReq` instead of just `req`. They're both the same type.

request.go
```diff
@@ -124,3 +128,49 @@ func detectEncoding(opts RequestOptions) encoding.Encoding {
 
 	return encoding.JSON
 }
+
+// prepares the request by injecting metadata, applying plugin-based transport middleware,
+// before finally adding any user-provided override values
+func prepareRequest(req *transport.Request, headers map[string]string, opts Options) (*transport.Request, error) {
+	reqWithMeta := mutateRequestWithMetadata(req, opts.TOpts)
```
abhinav: nit: We don't need to be this verbose with the operations. The operations
being executed are pretty trivial so it's preferable not to jump around, which
is especially difficult with the long similar-looking names.

Can you perform the operations inline? You can group blocks of code with
comments to serve the same purpose as these names.

```go
// Add request metadata
req.TargetService = ...

// Apply middleware
ctx, cancel := ...
defer cancel()
req, err := transportmiddleware.Apply(ctx, req)
if err != nil {
  return nil, err
}

// Apply command line arguments
req.Headers = ...
req.TransportHeaders = ...
```

I understand you have tests for the three different functions. I would
recommend collapsing those tests so that you're testing prepareRequest
directly. The reason for this being that we do want to test whether CLI
overrides actually override the default metadata and the middleware.

---

If you strongly prefer to keep the functions, please rename them to something
like, `addRequestMetadata`, `applyMiddleware`, and `applyCLIOverrides`
respectively.

request.go
```diff
@@ -124,3 +128,49 @@ func detectEncoding(opts RequestOptions) encoding.Encoding {
...
+	req.Headers = appendMap(req.Headers, headers)
+	req.TransportHeaders = appendMap(req.TransportHeaders, opts.TOpts.TransportHeaders)
+	req.Baggage = appendMap(req.Baggage, opts.ROpts.Baggage)
+	req.Timeout = timeout
+	return req
+}
+
+// copy key-value pairs from src to dest, overwriting existing values
+func appendMap(dest, src map[string]string) map[string]string {
+	res := dest
```
abhinav: nit: why is `res` declared at all. You can just use `dest` in place of `res`.

request.go
```diff
@@ -124,3 +128,49 @@ func detectEncoding(opts RequestOptions) encoding.Encoding {
...
+	}
+	req.Headers = appendMap(req.Headers, headers)
+	req.TransportHeaders = appendMap(req.TransportHeaders, opts.TOpts.TransportHeaders)
+	req.Baggage = appendMap(req.Baggage, opts.ROpts.Baggage)
+	req.Timeout = timeout
+	return req
+}
+
+// copy key-value pairs from src to dest, overwriting existing values
+func appendMap(dest, src map[string]string) map[string]string {
```
abhinav: nit: It's less of an append and more of an `(update/merge)Map`. append implies ordering.

transportmiddleware/transportmiddleware.go
```diff
@@ -0,0 +1,33 @@
...
+
+// Register sets the provided transport middleware to be used on future
+// calls to Apply(). Calls to Register() will overwrite previously registered
+// middlewares; that is, only one middleware is allowed at a time.
+func Register(tm TransportMiddleware) {
+	registeredMiddleware = tm
+}
+
+// TransportMiddleware
+type TransportMiddleware interface {
```
abhinav: incomplete comment

transportmiddleware/transportmiddleware.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/yarpc/yab/transport"
+)
+
+var registeredMiddleware TransportMiddleware
+
+// Register sets the provided transport middleware to be used on future
+// calls to Apply(). Calls to Register() will overwrite previously registered
+// middlewares; that is, only one middleware is allowed at a time.
+func Register(tm TransportMiddleware) {
+	registeredMiddleware = tm
```
abhinav: We should consider thread-safety here. Let's acquire a write lock before changing the value.

transportmiddleware/transportmiddleware.go
```diff
@@ -0,0 +1,33 @@
...
+	//
+	// Implementations are prevented from modifying the Request object
+	// in the case that an error is encountered.
+	Apply(ctx context.Context, req *transport.Request) (*transport.Request, error)
+}
+
+// Apply mutates a Request using the previously registered TransportMiddleware.
+func Apply(ctx context.Context, req *transport.Request) (*transport.Request, error) {
+	if registeredMiddleware == nil {
+		return req, nil
```
abhinav: Acquire a read lock, get the middleware, release the lock, and then apply middleware.

transportmiddleware/transportmiddleware.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/yarpc/yab/transport"
+)
+
+var registeredMiddleware TransportMiddleware
+
+// Register sets the provided transport middleware to be used on future
+// calls to Apply(). Calls to Register() will overwrite previously registered
+// middlewares; that is, only one middleware is allowed at a time.
+func Register(tm TransportMiddleware) {
```
abhinav: A little thing worth having on things that change globals like this is a way to restore the old value.

```
func Register(tm TransportMiddleware) (restore func()) {
  oldMW := registeredMiddleware
  // ..
  return func() {
    registeredMiddleware = oldMW
  }
}
```

This makes your tests cleaner because you don't have to worry about polluting global state. Tests can just do,

```
restore := Register(..)
defer restore()
```

main.go
APTy: i'm not a big fan of reassigning/mutating data in general, which is why i'm generally pretty explicitly about it throughout this PR

abhinav: Yeah, except that's unavoidable in this situation. Mutation is part of the
lifecycle of the object here. The two references are the same. For all cases,
`req == rawReq` so there's no reason for them to be two separate variables.

APTy:
fixed the comments minus possible `RequestInterceptor` name change

and i can squash all these commits at the end

request_test.go
```diff
@@ -306,3 +310,78 @@ func TestDetectEncoding(t *testing.T) {
...
+	assert.Equal(t, "foo", req.Method)
+	assert.Equal(t, "bar", req.TargetService)
+}
+
+func TestNewRequestWithTransportMiddleware(t *testing.T) {
+	req := &transport.Request{Method: "foo"}
+	topts := TransportOptions{ServiceName: "bar"}
+	restore := transportmiddleware.Register(mockTransportMiddleware{})
+	defer restore()
+	req, err := prepareRequest(req, nil, Options{TOpts: topts})
```
abhinav: nit: For readability, when passing a naked nil as a position argument,
add a comment indicating the parameter name.

    prepareRequest(req, nil /* headers */, Options{...})

transportmiddleware/interface.go
```diff
@@ -0,0 +1,50 @@
...
+	"sync"
+
+	"github.com/yarpc/yab/transport"
+)
+
+// stores the currently registered middleware
+var registeredMiddleware Interface
+
+// serializes access to the currently registered middleware
+var registerLock sync.RWMutex
```
abhinav: nit: Use a var group here.

```
var (
  // ..
  registeredMiddleware Interface

  // ..
  registerLock sync.RWMutex
)
```

transportmiddleware/interface.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Register sets the provided transport middleware to be used on future
+// calls to Apply(). Calls to Register() will overwrite previously registered
+// middlewares; that is, only one middleware is allowed at a time.
+func Register(newMW Interface) (restore func()) {
```
abhinav: Add,

```
// Returns a function to undo the change made by this call.
```

transportmiddleware/interface_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/yarpc/yab/transport"
```
abhinav: nit: Import group ordering. It should be,

- standard library
- local imports (packages of the yab repo)
- external imports

transportmiddleware/interface_test.go
```diff
@@ -0,0 +1,85 @@
...
+		wantErr      bool
+		testRace     bool
+	}{
+		{ /* run without options */ },
+		{wantErr: true},
+		{dontRegister: true},
+		{testRace: true},
+	}
+	for idx, tt := range tests {
+		var restore func() = func() {}
```
abhinav: nit:

    restore := func() {}

transportmiddleware/interface_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+		// create the test middleware
+		tm := &headerTransportMiddleware{
+			wantErr: tt.wantErr,
+		}
+		if tt.testRace {
+			go Register(tm)
+		}
```
abhinav: This seems like it's testing a bunch of unrelated things together.

For the data race test at least, I would create a separate test.

You'll want multiple goroutines trying to Register at the same time. You
can use a `chan` to synchronize the Register call to increase likelihood
of conflict. You can do the same for the `restore` call as well.

```go
func TestRegisterRace(t *testing.T) {
  registerCh := make(chan struct{})
  restoreCh := make(chan struct{})
  var wg sync.WaitGroup

  for i := 0; i < 100; i++ {
    wg.Add(1)
    go func() {
      <-registerCh  // wait until register signal is received
      restore := Register(..)
      
      <- restoreCh
      restore()
      wg.Done()
    }()
  }

  close(registerCh)
  close(restoreCh)
  wg.Wait()
  
  // check that middleware is nil now
}
```

APTy: This is great, haven't seen `close()` used like that before.. I like it 👍 

Also I don't think we can guarantee that the `registeredMiddleware` will be `nil` at the end, since it sort of implies we are doing many `register(a) -> restore(a) -> register(b) -> restore(b)` calls in a linear fashion. If we were to ever do `register(a) -> register(b) -> restore(a) -> restore(b)`, we would be left with `a` at the end, since that was the prior value when `b` was registered.

abhinav: Oh, true. Okay, yeah we can skip that test, I suppose.

request.go
```diff
@@ -124,3 +128,40 @@ func detectEncoding(opts RequestOptions) encoding.Encoding {
...
+	}
+	req.Headers = updateMap(req.Headers, headers)
+	req.TransportHeaders = updateMap(req.TransportHeaders, opts.TOpts.TransportHeaders)
+	req.Baggage = updateMap(req.Baggage, opts.ROpts.Baggage)
+	req.Timeout = timeout
+	return req, nil
+}
+
+// copy key-value pairs from src to dest, overwriting existing values
+func updateMap(dest, src map[string]string) map[string]string {
```
prashantv: maybe `mergeMap`?

request.go
```diff
@@ -124,3 +128,40 @@ func detectEncoding(opts RequestOptions) encoding.Encoding {
...
 	return encoding.JSON
 }
+
+// prepares the request by injecting metadata, applying plugin-based transport middleware,
+// before finally adding any user-provided override values
+func prepareRequest(req *transport.Request, headers map[string]string, opts Options) (*transport.Request, error) {
+	// Add request metadata
+	req.TargetService = opts.TOpts.ServiceName
+
+	// Apply middleware
```
prashantv: why do we apply middleware first? does it make more sense to apply middleware at the very end?

e.g., when you use middleware in other frameworks, you have a request with a body, baggage, headers, and the middleware can then modify it

request_test.go
```diff
@@ -306,3 +310,78 @@ func TestDetectEncoding(t *testing.T) {
...
+	defer restore()
+	req, err := prepareRequest(req, nil /* headers */, Options{TOpts: topts})
+	assert.NoError(t, err)
+	assert.Equal(t, "bar", req.Method)
+	assert.Equal(t, "bar", req.TargetService)
+}
+
+type mockTransportMiddleware struct{}
+
+func (tm mockTransportMiddleware) Apply(_ context.Context, req *transport.Request) (*transport.Request, error) {
```
prashantv: once we apply middleware after the request is created, verify the request properties here

request.go
```diff
@@ -124,3 +128,40 @@ func detectEncoding(opts RequestOptions) encoding.Encoding {
 
 	return encoding.JSON
 }
+
+// prepares the request by injecting metadata, applying plugin-based transport middleware,
+// before finally adding any user-provided override values
+func prepareRequest(req *transport.Request, headers map[string]string, opts Options) (*transport.Request, error) {
+	// Add request metadata
+	req.TargetService = opts.TOpts.ServiceName
```
prashantv: the fact that service is set, but headers are not seems like a strange design. we should consistently set everything in the same place

transportmiddleware/interface.go
```diff
@@ -0,0 +1,53 @@
+// Package transportmiddleware provides an interface for hooking into
+// outgoing yab requests.
+package transportmiddleware
```
prashantv: is it worth having a special interface for this? does this make more sense as `transport.Middleware`

transportmiddleware/interface.go
```diff
@@ -0,0 +1,53 @@
...
+	// serializes access to the currently registered middleware
+	registerLock sync.RWMutex
+)
+
+// Register sets the provided transport middleware to be used on future
+// calls to Apply(). Calls to Register() will overwrite previously registered
+// middlewares; that is, only one middleware is allowed at a time.
+// Returns a function to undo the change made by this call.
+func Register(newMW Interface) (restore func()) {
+	registerLock.Lock()
```
prashantv: (sorry for the back and forth about this topic)
i don't think we want a lock here. if we have multiple callers racing to call `Register`, the behaviour is completely undefined. that's bad behaviour, we'd prefer if that triggered a race.

request.go
APTy: The idea was to apply the middleware first, such that it could be overridden by the CLI options afterwards.

Another option would be to create the fully-formed request first, but ask that `RequestInterceptor` implementations do not override pre-existing values, if they wish to take advantage of them.

APTy:
Thanks! I'm liking where this is at. A couple thoughts:

- I agree that middlewares should be applied as a final pre-flight step.. but do we need to be more clear to implementors that they have the option to _not_ override pre-existing values? Or is this already implied by the nature of the middleware?

- I am concerned that middleware is applied in an all-or-nothing fashion. That is, if the middleware fails to apply, the request fails. We definitely want to surface these errors to the user, but if the middleware can't be applied, and the user wishes to make their request without it, they don't have much recourse to do so.

abhinav:
Summarizing offline discussion with @prashantv here:

This PR should be fine as-is. For middleware where we're concerned about
breaking the request path, we should expose an environment variable and a flag
to opt-out of that middleware.

Let's implement flag injection from plugins in a separate PR. You'll want to
export some variant of [`Group.AddGroup`][1] method from go-flags to plugins
without exposing the go-flags types.

For your specific middleware, we recommend that the failure path explodes and
explicitly requires users to opt out using either the environment variable or
the flag.

[1]: https://godoc.org/github.com/jessevdk/go-flags#Group.AddGroup

APTy:
Sounds great! So we'll fix this on the implementation side for now. I'll add flag injection in a new PR.

request.go
```diff
@@ -124,3 +127,41 @@ func detectEncoding(opts RequestOptions) encoding.Encoding {
...
+	req.Baggage = mergeMap(req.Baggage, opts.ROpts.Baggage)
+	req.Timeout = timeout
+
+	// Add request metadata
+	req.TargetService = opts.TOpts.ServiceName
+
+	// Apply middleware
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	req, err := transport.ApplyInterceptor(ctx, req)
```
prashantv: feel free to `return transport.ApplyIntercetpr(...)` rather than the `if` below

request.go
```diff
@@ -124,3 +127,41 @@ func detectEncoding(opts RequestOptions) encoding.Encoding {
...
+
+// prepares the request by injecting metadata, applying plugin-based transport middleware,
+// before finally adding any user-provided override values
+func prepareRequest(req *transport.Request, headers map[string]string, opts Options) (*transport.Request, error) {
+	// Apply command line arguments
+	timeout := opts.ROpts.Timeout.Duration()
+	if timeout == 0 {
+		timeout = time.Second
+	}
+	req.Headers = mergeMap(req.Headers, headers)
```
prashantv: do we still need `mergeMap` at all?

request_test.go
```diff
@@ -306,3 +309,87 @@ func TestDetectEncoding(t *testing.T) {
...
+	assert.Equal(t, "baz", req.Method)
+	assert.Equal(t, "bar", req.TargetService)
+}
+
+type mockRequestInterceptor struct {
+	method  string
+	baggage map[string]string
+}
+
+func (ri mockRequestInterceptor) Apply(_ context.Context, req *transport.Request) (*transport.Request, error) {
```
prashantv: can we add a test which returns an error

transport/interface.go
```diff
@@ -36,6 +36,7 @@ type Request struct {
 	TransportHeaders map[string]string
 	ShardKey         string
 	Body             []byte
+	TargetService    string
 }
```
prashantv: super nit: can you make this the first field in the struct

transport/request_interceptor_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+		// create test request
+		headers := map[string]string{"zim": "zam"}
+		rawReq := &Request{
+			Headers: headers,
+			Method:  "get",
+		}
+
+		// modify the test request
+		req, err := ApplyInterceptor(context.TODO(), rawReq)
```
prashantv: just use `context.Background()`

PR #222, opened 2017-10-06
Add flag parser plugin for injecting custom CLI flags

APTy:
Following #220, we should create a way for plugins to specify custom flags to `yab` invocations.

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,6 +32,7 @@ import (
 
 	"github.com/yarpc/yab/encoding"
 	"github.com/yarpc/yab/peerprovider"
+	"github.com/yarpc/yab/plugin"
 	"github.com/yarpc/yab/transport"
 
 	"github.com/casimir/xdg-go"
@@ -135,6 +136,12 @@ yab includes a full man page (man yab), which is also available online: http://y
 	setGroupDescs(parser, "request", "Request Options", toGroff(_reqOptsDesc))
 	setGroupDescs(parser, "transport", "Transport Options", toGroff(_transportOptsDesc))
 	setGroupDescs(parser, "benchmark", "Benchmark Options", toGroff(_benchmarkOptsDesc))
+
+	if errs := plugin.AddToParser(pluginParserAdapter{parser}); errs != nil {
+		for _, err := range errs {
+			out.Warnf("WARNING: Error adding plugin-based custom flags: %v. Continuing.", err)
+		}
+	}
 
 	remaining, err := parser.ParseArgs(args)
 	// If there are no arguments specified, write the help.
--- a/main.go
+++ b/main.go
@@ -41,7 +41,6 @@ import (
 	opentracing_ext "github.com/opentracing/opentracing-go/ext"
 	"github.com/uber/jaeger-client-go"
 	"github.com/uber/tchannel-go"
-	"go.uber.org/multierr"
 	"go.uber.org/zap"
 )
 
@@ -139,9 +138,7 @@ yab includes a full man page (man yab), which is also available online: http://y
 	setGroupDescs(parser, "benchmark", "Benchmark Options", toGroff(_benchmarkOptsDesc))
 
 	if err := plugin.AddToParser(pluginParserAdapter{parser}); err != nil {
-		for _, err := range multierr.Errors(err) {
-			out.Warnf("WARNING: Error adding plugin-based custom flags: %v. Continuing.", err)
-		}
+		out.Warnf("WARNING: Error adding plugin-based custom flags: %+v.", err)
 	}
 
 	remaining, err := parser.ParseArgs(args)
--- a/main.go
+++ b/main.go
@@ -41,6 +41,7 @@ import (
 	opentracing_ext "github.com/opentracing/opentracing-go/ext"
 	"github.com/uber/jaeger-client-go"
 	"github.com/uber/tchannel-go"
+	"go.uber.org/multierr"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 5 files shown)


plugin/plugin_test.go
```diff
@@ -0,0 +1,130 @@
...
+	return nil
+}
+
+type fooFlags struct {
+	Bar string `long:"bar" description:"Sets the 'bar' value."`
+}
+
+// defer-able test cleanup for global state
+func resetFlags(_flags []*flag) func() {
+	flags = _flags
```
abhinav: Isn't this setFlags with a `reset` return?

also, nit: local variables shouldn't be `_` prefixed

plugin/plugin_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+func TestAddToParser(t *testing.T) {
+	tests := []struct {
+		shouldErr                 bool
+		errOnCallCount            int
+		numErrs                   int
+		numAddFlagGroupCallCounts int
+		numParserFlags            int
+	}{
+		{numErrs: 0, numAddFlagGroupCallCounts: 3, numParserFlags: 3},
```
abhinav: can skip `numErrs: 0` since it's zero-value

plugin/plugin_test.go
```diff
@@ -0,0 +1,130 @@
...
+			errOnCallCount: tt.errOnCallCount,
+		}
+
+		errs := AddToParser(p)
+		if tt.shouldErr {
+			assert.NotNil(t, errs)
+		} else {
+			assert.Nil(t, errs)
+		}
+		assert.Equal(t, tt.numErrs, len(errs))
```
abhinav: this will probably go away once this is a single collated error

main.go
```diff
@@ -136,6 +138,12 @@ yab includes a full man page (man yab), which is also available online: http://y
 	setGroupDescs(parser, "transport", "Transport Options", toGroff(_transportOptsDesc))
 	setGroupDescs(parser, "benchmark", "Benchmark Options", toGroff(_benchmarkOptsDesc))
 
+	if err := plugin.AddToParser(pluginParserAdapter{parser}); err != nil {
+		for _, err := range multierr.Errors(err) {
+			out.Warnf("WARNING: Error adding plugin-based custom flags: %v. Continuing.", err)
+		}
```
abhinav: optional: Can do,

```
if err := plugin.AddToParser(..); err != nil {
  out.Warnf("WARNING: Error adding plugin-based custom flags: %+v", err)
}
```
multierr errors use a nice multi-line formatted message if `%+v` was used.

See the example in https://godoc.org/go.uber.org/multierr#Combine

plugin/plugin.go
```diff
@@ -0,0 +1,65 @@
...
+//	AddFlags("Foo Options", "", &foo{})
+//
+// This would inject a custom flag group that looks like:
+//
+//	Foo Options
+// 		--bar	Sets the 'bar' value.
+//
+// In order to retrieve the mutated results of the set flag, users of AddFlags() should
+// retain a reference to the `data` object and check its values after parsing is complete.
+func AddFlags(shortDescription string, longDescription string, data interface{}) {
```
abhinav: minor nit: s/shortDescription/groupName

plugin/plugin_test.go
```diff
@@ -0,0 +1,134 @@
...
+		numErrs                   int
+		numAddFlagGroupCallCounts int
+		numParserFlags            int
+	}{
+		{numAddFlagGroupCallCounts: 3, numParserFlags: 3},
+		{shouldErr: true, numErrs: 3},
+		{shouldErr: true, errOnCallCount: 2, numErrs: 1, numAddFlagGroupCallCounts: 2, numParserFlags: 2},
+	}
+
+	threeFlags := []*flag{{}, {}, {}}
```
abhinav: This is still a very confusing test case. It may be easier to make it not-a-table-test.

plugin/plugin_test.go
```diff
@@ -0,0 +1,134 @@
...
+	mockFlags := make([]*flag, numFlags)
+	for i := range mockFlags {
+		mockFlags[i] = &flag{shortDescription: fmt.Sprintf("foo-%d", i)}
+	}
+	defer setFlags(mockFlags)()
+
+	p := &mockParser{}
+
+	errs := AddToParser(p)
+	assert.Nil(t, errs)
```
abhinav: NoError

plugin/plugin.go
```diff
@@ -0,0 +1,65 @@
...
+func AddFlags(groupName string, longDescription string, data interface{}) {
+	_flags = append(_flags, &flag{
+		groupName:       groupName,
+		longDescription: longDescription,
+		data:            data,
+	})
+}
+
+// Parser is any object that can add flag groups to itself before performing its parse.
+type Parser interface {
```
prashantv: Any advantage in having this interface not match the `go-flags` interface exactly, and just calling it `FlagParser`? I'm not sure why we have `pluginParserAdapter`, doesn't seem to add much value.

plugin/plugin.go
```diff
@@ -0,0 +1,65 @@
...
+	for _, f := range _flags {
+		flagErr := p.AddFlagGroup(f.groupName, f.longDescription, f.data)
+		if flagErr != nil {
+			err = multierr.Append(err, fmt.Errorf("adding %v to parser: %v", f.groupName, flagErr))
+		}
+	}
+	return err
+}
+
+// stores the list of currently registered flags
```
prashantv: move the type and the variable to the top of the file. we should probably have the type first, then `_flags`

plugin/plugin.go
APTy: It leaks the internals a bit, by exposing go-flags as part of the public API. Figured we'd want to avoid that

I realize the go-flags tags are exposed anyway when users add their options struct, but it's more subtle

Mostly preference, I'll leave the final call to you all

prashantv: Fair enough, I don't have super strong opinions here, so don't mind either way. This seems fine too.

APTy: Okay I may just leave it as-is in that case, thanks!

PR #225, opened 2017-10-27
Add gRPC transport.Transport

peter-edge:
This adds a grpc `transport.Transport` based on yarpc-go's `*grpc.Outbound`, as well as integration testing using JSON. The work to actually add gRPC to yab will come afterwards as this is more complicated - yab assumes there will only be tchannel and http, so there's a lot of this:

```go
if something {
  return "tchannel"
}
return "http"
```

That I'll need to refactor to assume >2 transports.

### DIFF
--- a/transport/grpc.go
+++ b/transport/grpc.go
@@ -132,11 +132,7 @@ func (t *grpcTransport) Call(ctx context.Context, request *Request) (*Response, 
 	if err != nil {
 		return nil, err
 	}
-	response, err := transportResponseToResponse(transportResponse)
-	if err != nil {
-		return nil, err
-	}
-	return response, nil
+	return transportResponseToResponse(transportResponse)
 }
 
 func (t *grpcTransport) Close() error {
--- a/transport/grpc.go
+++ b/transport/grpc.go
@@ -124,18 +124,18 @@ func (t *grpcTransport) Call(ctx context.Context, request *Request) (*Response, 
 
 	ctx, cancel := requestContextWithTimeout(ctx, request)
 	defer cancel()
-	transportResponse, err := t.Outbound.Call(ctx, t.requestToTransportRequest(request))
+	transportResponse, err := t.Outbound.Call(ctx, t.requestToYARPCRequest(request))
 	if err != nil {
 		return nil, err
 	}
-	return transportResponseToResponse(transportResponse)
+	return yarpcResponseToResponse(transportResponse)
 }
 
 func (t *grpcTransport) Close() error {
 	return multierr.Combine(t.Transport.Stop(), t.Outbound.Stop())
 }
 
-func (t *grpcTransport) requestToTransportRequest(request *Request) *transport.Request {
+func (t *grpcTransport) requestToYARPCRequest(request *Request) *transport.Request {
 	return &transport.Request{
 		Caller:          t.Caller,
 		Service:         request.TargetService,
@@ -160,7 +160,7 @@ func requestContextWithTimeout(ctx context.Context, request *Request) (context.C
 	return context.WithTimeout(ctx, timeout)
 }
 
-func transportResponseToResponse(transportResponse *transport.Response) (*Response, error) {
+func yarpcResponseToResponse(transportResponse *transport.Response) (*Response, error) {
 	response := &Response{
 		Headers: transportResponse.Headers.Items(),
 	}
--- a/transport/grpc.go
+++ b/transport/grpc.go
@@ -85,12 +85,8 @@ func newGRPC(options GRPCOptions) (*grpcTransport, error) {
 		return nil, errGRPCNoEncoding
 	}
 
-	identifiers := make([]apipeer.Identifier, len(options.Addresses))
-	for i, address := range options.Addresses {
-		identifiers[i] = hostport.Identify(address)
-	}
 	transport := grpc.NewTransport(grpc.Tracer(options.Tracer))
-	outbound := transport.NewOutbound(peer.Bind(roundrobin.New(transport), peer.BindPeers(identifiers)))
+	outbound := transport.NewOutbound(peer.Bind(roundrobin.New(transport), peer.BindPeers(peersToIdentifiers(options.Addresses))))
 	if err := transport.Start(); err != nil {
 		return nil, err
 	}
@@ -155,7 +151,7 @@ func (t *grpcTransport) requestToTransportRequest(request *Request) *transport.R
 
 func requestContextWithTimeout(ctx context.Context, request *Request) (context.Context, context.CancelFunc) {
 	if _, ok := ctx.Deadline(); ok {
-		return ctx, noop
+		return ctx, func() {}
 	}
 	timeout := time.Second
 	if request.Timeout > 0 {
@@ -181,4 +177,10 @@ func transportResponseToResponse(transportResponse *transport.Response) (*Respon
 	return response, nil
 }
 
-func noop() {}
+func peersToIdentifiers(peers []string) []apipeer.Identifier {
+	identifiers := make([]apipeer.Identifier, len(peers))
+	for i, peer := range peers {
+		identifiers[i] = hostport.Identify(peer)
+	}
+	return identifiers
+}
--- a/transport/grpc.go
+++ b/transport/grpc.go
@@ -1,0 +1,188 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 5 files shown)


peter-edge:
May be affected by https://github.com/yarpc/yarpc-go/pull/1346 but shouldn't be if backwards compatible.

transport/grpc.go
```diff
@@ -0,0 +1,188 @@
...
+
+	"github.com/opentracing/opentracing-go"
+	"go.uber.org/multierr"
+	apipeer "go.uber.org/yarpc/api/peer"
+	"go.uber.org/yarpc/api/transport"
+	"go.uber.org/yarpc/peer"
+	"go.uber.org/yarpc/peer/hostport"
+	"go.uber.org/yarpc/peer/roundrobin"
+	"go.uber.org/yarpc/transport/grpc"
+	"golang.org/x/net/context"
```
willhug: should we use the core "context" package?

transport/grpc.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+var (
+	errGRPCNoAddresses = errors.New("must specify at least one grpc address")
+	errGRPCNoTracer    = errors.New("must specify grpc tracer")
```
willhug: should we just fill this with a noop?  Or does this already happen automatically for all transports?

transport/grpc.go
```diff
@@ -0,0 +1,188 @@
...
+		identifiers[i] = hostport.Identify(address)
+	}
+	transport := grpc.NewTransport(grpc.Tracer(options.Tracer))
+	outbound := transport.NewOutbound(peer.Bind(roundrobin.New(transport), peer.BindPeers(identifiers)))
+	if err := transport.Start(); err != nil {
+		return nil, err
+	}
+	if err := outbound.Start(); err != nil {
+		_ = transport.Stop()
+		return nil, err
```
willhug: multierr.Combine?

transport/grpc.go
```diff
@@ -0,0 +1,188 @@
...
+		return nil, errGRPCNoProcedure
+	}
+
+	ctx, cancel := requestContextWithTimeout(ctx, request)
+	defer cancel()
+	transportResponse, err := t.Outbound.Call(ctx, t.requestToTransportRequest(request))
+	if err != nil {
+		return nil, err
+	}
+	response, err := transportResponseToResponse(transportResponse)
```
willhug: `return transportResponseToResponse(...)`?

transport/grpc.go
```diff
@@ -0,0 +1,188 @@
...
+		ShardKey:        request.ShardKey,
+		RoutingKey:      t.RoutingKey,
+		RoutingDelegate: t.RoutingDelegate,
+		Body:            bytes.NewReader(request.Body),
+	}
+}
+
+func requestContextWithTimeout(ctx context.Context, request *Request) (context.Context, context.CancelFunc) {
+	if _, ok := ctx.Deadline(); ok {
+		return ctx, noop
```
willhug: should we just inline the `func(){}`?  The `noop` var is not used anywhere else.

transport/grpc_test.go
```diff
@@ -0,0 +1,273 @@
...
+	numInbounds int,
+	procedures []transport.Procedure,
+) (_ *testEnv, err error) {
+	yarpcTransport := grpc.NewTransport()
+	if err := yarpcTransport.Start(); err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err != nil {
+			err = multierr.Append(err, yarpcTransport.Stop())
```
willhug: doesn't multierr.Append already do a nil check?

transport/grpc_test.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+type testEnv struct {
+	Caller         string
+	Transport      TransportCloser
+	YARPCTransport *grpc.Transport
+	YARPCInbounds  []*grpc.Inbound
+}
+
+func newTestEnv(
```
willhug: `newGRPCTestEnv`? This scope is shared across all `yab/transport`

transport/grpc_test.go
```diff
@@ -0,0 +1,273 @@
...
+		}
+		addresses[i] = listener.Addr().String()
+		yarpcInbound := yarpcTransport.NewInbound(listener)
+		yarpcInbound.SetRouter(newTestRouter(procedures))
+		if err := yarpcInbound.Start(); err != nil {
+			return nil, err
+		}
+		defer func() {
+			if err != nil {
+				err = multierr.Append(err, yarpcInbound.Stop())
```
willhug: same, Append should already nil check.

transport/grpc_test.go
```diff
@@ -0,0 +1,273 @@
...
+	addresses := make([]string, numInbounds)
+	yarpcInbounds := make([]*grpc.Inbound, numInbounds)
+	for i := 0; i < numInbounds; i++ {
+		listener, err := net.Listen("tcp", "127.0.0.1:0")
+		if err != nil {
+			return nil, err
+		}
+		addresses[i] = listener.Addr().String()
+		yarpcInbound := yarpcTransport.NewInbound(listener)
+		yarpcInbound.SetRouter(newTestRouter(procedures))
```
willhug: what's the argument about just putting all these inbounds into a dispatcher and starting/stopping that once?  Is there sufficient custom functionality here to merit it's own testRouter?

transport/grpc.go
peter-edge: This is required due to dependencies that expect this, I did every permutation, see https://github.com/yarpc/yab/pull/227

transport/grpc.go
peter-edge: This matches what other transports do.

transport/grpc.go
peter-edge: I thought this was fine as adding the error from `Stop` when the `Start` error is what was the real problem could be confusing.

transport/grpc.go
peter-edge: Done

transport/grpc.go
peter-edge: I figured this might save an alloc? Actually not sure how inline functions are allocated.

transport/grpc_test.go
peter-edge: The only exposed test router in yarpc-go is one that uses mocks, which I don't want.

peter-edge: Oh, you mean a dispatcher as total - I like this current setup without relying on config, ie explicitly setting it up, I don't see too much benefit going to a dispatcher, this gives me more theoretical control.

transport/grpc.go
prashantv: +1, although we tend to just do `transport.Stop()`, don't think we use `errcheck` here.

transport/grpc.go
```diff
@@ -0,0 +1,184 @@
...
+		return nil, errGRPCNoTracer
+	}
+	if options.Caller == "" {
+		return nil, errGRPCNoCaller
+	}
+	if options.Encoding == "" {
+		return nil, errGRPCNoEncoding
+	}
+
+	identifiers := make([]apipeer.Identifier, len(options.Addresses))
```
prashantv: can you make a private function, it's a pure function that does `[]string -> []apipeer.Identifier`

transport/grpc.go
prashantv: feel free to just do `func() {}`, since the timeout path allocates anyway

transport/grpc.go
```diff
@@ -0,0 +1,184 @@
...
+
+func requestContextWithTimeout(ctx context.Context, request *Request) (context.Context, context.CancelFunc) {
+	if _, ok := ctx.Deadline(); ok {
+		return ctx, noop
+	}
+	timeout := time.Second
+	if request.Timeout > 0 {
+		timeout = request.Timeout
+	}
+	return context.WithTimeout(ctx, timeout)
```
prashantv: do we need this at all? i think that the caller should always pass a context with the timeout, they just include the value in the `Request` so that you can pass it through headers for propagation if needed.

transport/grpc_test.go
prashantv: it does

transport/grpc_test.go
prashantv: +1, maybe even `runGRPCTest`

transport/grpc.go
peter-edge: Done

transport/grpc.go
peter-edge: Done

transport/grpc.go
peter-edge: I don't think the caller always does this, and this is for safety/to take into account the impls in the other transports.

transport/grpc_test.go
peter-edge: Done

transport/grpc_test.go
peter-edge: Done

transport/grpc_test.go
peter-edge: Done

transport/grpc_test.go
peter-edge: Whoops, this was on purpose actually - I only want to call `yarpcTransport.Stop()` if `err != nil`, `err` is the error being returned - if there was a problem creating and starting everything, I want to tear this down.

peter-edge:
All comments should be addressed @willhug @prashantv

transport/grpc.go
```diff
@@ -0,0 +1,186 @@
...
+		return nil, err
+	}
+	return transportResponseToResponse(transportResponse)
+}
+
+func (t *grpcTransport) Close() error {
+	return multierr.Combine(t.Transport.Stop(), t.Outbound.Stop())
+}
+
+func (t *grpcTransport) requestToTransportRequest(request *Request) *transport.Request {
```
prashantv: nit: maybe `requestToYARPCRequest`, since we're already in the transport package so saying `ToTransportRequest` is a little unclear

transport/grpc.go
```diff
@@ -0,0 +1,186 @@
...
+		return ctx, func() {}
+	}
+	timeout := time.Second
+	if request.Timeout > 0 {
+		timeout = request.Timeout
+	}
+	return context.WithTimeout(ctx, timeout)
+}
+
+func transportResponseToResponse(transportResponse *transport.Response) (*Response, error) {
```
prashantv: same as earlier, maybe even just `convertYARPCResponse`

PR #226, opened 2017-10-27
Cleanup Travis and Makefiles

peter-edge:
- Move from golang 1.7/1.8 to 1.8/1.9 in Travis CI
- Add `/cover` and `/cover.out` to `.gitignore`
- Add `grep -v \/testdata\/` to setting the `PACKAGES` variable so this isn't printed:

```
go build -i ./testdata/... ./peerprovider/... ./transport/... ./limiter/... ./templateargs/... ./plugin/... ./internal/... ./encoding/... ./statsd/... ./thrift/... ./ratelimit/... ./unmarshal/... ./scripts/... ./sorted/... .
warning: "./testdata/..." matched no packages
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PACKAGES := $(shell glide novendor)
+PACKAGES := $(shell glide novendor | grep -v \/testdata\/)
 
 export GO15VENDOREXPERIMENT=1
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,4 +1,4 @@
-PACKAGES := $(shell glide novendor)
+PACKAGES := $(shell glide novendor | grep -v \/testdata\/)
```
abhinav: nit: `grep -v '/testdata/'` is more readable.

prashantv: +1, will push a commit

PR #227, opened 2017-10-27
Update golang.org/x/net/context references to context

peter-edge:


peter-edge:
I guess this isn't fully possible because of tchannel-go, maybe partially?

```
# github.com/yarpc/yab/transport
transport/tchannel_test.go:131: cannot use func literal (type func("context".Context, error)) as type func("github.com/yarpc/yab/vendor/golang.org/x/net/context".Context, error) in argument to "github.com/yarpc/yab/vendor/github.com/uber/tchannel-go/json".Register
transport/tchannel_test.go:180: cannot use func literal (type func("context".Context, *raw.Args) (*raw.Res, error)) as type func("github.com/yarpc/yab/vendor/golang.org/x/net/context".Context, *raw.Args) (*raw.Res, error) in argument to testutils.RegisterFunc
transport/tchannel_test.go:310: cannot use func literal (type func("context".Context, *raw.Args) (*raw.Res, error)) as type func("github.com/yarpc/yab/vendor/golang.org/x/net/context".Context, *raw.Args) (*raw.Res, error) in argument to testutils.RegisterFunc
FAIL	github.com/yarpc/yab/transport [build failed]
```

PR #228, opened 2017-10-30
WIP: Add gRPC to main

peter-edge:
This builds on the previous PR to actually add grpc into the main. This adds a `--protocol` flag, which upon further review we probably do not want.

I want this as a discussion point to figure out how we want to do this, either to require explicitly setting `grpc://` as a prefix, or to add the `--protocol` flag before further testing. Both work as of now, ie you can either do `--peer 0.0.0.0:8088 --protocol grpc` or `--peer grpc://0.0.0.0:8088`

Tested working locally (some internal naming changed):

```
$ yab --thrift=idl/uber/infra/lottery/lottery.thrift --service lottery --procedure Lottery::play --peer 0.0.0.0:8088 --request '{"request":{"number":10}}' --protocol grpc -d 10s
{
  "body": {
    "result": {
      "won": true
    }
  }
}

Benchmark parameters:
  CPUs:            8
  Connections:     16
  Concurrency:     1
  Max requests:    0
  Max duration:    10s
  Max RPS:         0
Latencies:
  0.5000: 1.481367ms
  0.9000: 4.120448ms
  0.9500: 4.978417ms
  0.9900: 7.441082ms
  0.9990: 27.914933ms
  0.9995: 36.157199ms
  1.0000: 91.973059ms
Elapsed time:      10.001s
Total requests:    77744
RPS:               7773.27
$ yab --thrift=idl/uber/infra/lottery/lottery.thrift --service lottery --procedure Lottery::play --peer 0.0.0.0:8087 --request '{"request":{"number":10}}' --protocol tchannel -d 10s
{
  "body": {
    "result": {
      "won": true
    }
  },
  "ok": true,
  "trace": "0"
}

Benchmark parameters:
  CPUs:            8
  Connections:     16
  Concurrency:     1
  Max requests:    0
  Max duration:    10s
  Max RPS:         0
Latencies:
  0.5000: 1.200049ms
  0.9000: 3.444818ms
  0.9500: 4.577865ms
  0.9900: 8.301695ms
  0.9990: 25.494045ms
  0.9995: 42.75586ms
  1.0000: 50.849356ms
Elapsed time:      10.001s
Total requests:    96247
RPS:               9623.46
```

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -45,7 +45,9 @@ var (
 	errPeerOptions     = errors.New("do not specify peers using --peer and --peer-list")
 )
 
+// works for either tchannel or grpc
 func remapLocalHost(hostPorts []string) {
+	// this works for either tchannel or grpc
 	ip, err := tchannel.ListenIP()
 	if err != nil {
 		panic(err)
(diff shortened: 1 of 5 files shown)


transport.go
```diff
@@ -45,7 +45,9 @@ var (
 	errPeerOptions     = errors.New("do not specify peers using --peer and --peer-list")
 )
 
+// works for either tchannel or grpc
```
prashantv: do we need this to work for grpc? we should listen on all interfaces for grpc, don't think this is needed.

transport.go
```diff
@@ -58,37 +60,69 @@ func remapLocalHost(hostPorts []string) {
...
+func parsePeer(explicitProtocol string, peer string) (protocol, host string, valid bool) {
+	// If we get a pure host:port, then we assume tchannel or grpc.
 	if _, _, err := net.SplitHostPort(peer); err == nil && !strings.Contains(peer, "://") {
-		return "tchannel", peer
+		if explicitProtocol == "" {
+			// If no explicit protocol specified, assume tchannel.
+			return "tchannel", peer, true
+		}
+		switch explicitProtocol {
+		case "http", "https", "ftp", "tchannel", "grpc":
```
prashantv: why do we have `ftp` here?

transport.go
```diff
@@ -58,37 +60,69 @@ func remapLocalHost(hostPorts []string) {
...
 
-func parsePeer(peer string) (protocol, host string) {
-	// If we get a pure host:port, then we assume tchannel.
+// host and port must be specified in peer
+// explicitProtocol is from --protocol flag
+// protocol will be tchannel, grpc, http, https, ftp in valid case
+// boolean will be false in invalid case
+// TODO(pedge): ftp actually valid? there is a test for it
+func parsePeer(explicitProtocol string, peer string) (protocol, host string, valid bool) {
+	// If we get a pure host:port, then we assume tchannel or grpc.
```
prashantv: this whole method is very confusing after adding the `explicitProtocol` support, if there's an explicit protocol, then there's no point guessing, that should be what we always return.

basically, if `explicitProtocol` is specified, then use it. if it doesn't match with the URL protocol (e.g., `--protocol grpc --peer tchannel://foo:1234`), then return an error. otherwise, guess the protocol as we currently do.

transport.go
```diff
@@ -143,16 +177,32 @@ func getTransport(opts TransportOptions, encoding encoding.Encoding, tracer open
...
+			Tracer:          tracer,
+		})
+	}
 
-		topts := transport.TChannelOptions{
+	hostPorts, valid := getHosts(opts.Protocol, opts.Peers)
+	if !valid {
+		return nil, fmt.Errorf("invalid peers '%v'", opts.Peers)
+	}
+	remapLocalHost(hostPorts)
```
prashantv: don't think we should do this for grpc

transport.go
peter-edge: This was part of a test that seemed to test that ftp was valid, I wasn't sure on this one.

peter-edge:
I think I'm just going to redo this to use `grpc://` only instead of being able to specify with a flag.

PR #231, opened 2017-11-16
Prepare for v0.12.0 release

APTy:


CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+# 0.12.0 (2017-11-16)
+* Add gRPC support
+* Cleanup Travis and Makefiles
+* Add plugin for injecting custom CLI flags
+* Add plugin for pre-flight request modification
+
```
prashantv: Can you update this to:
```
# 0.12.0 (2017-11-16)
* Add gRPC+Thrift support. Protobuf is not yet supported.
* Allow build-time plugins to register middleware for request modification.
* Allow build-time plugins to register custom CLI flags.
```

We tend to focus on user-facing changes (so ignore Travis/Makefiles), and I explicitly want to call out the gRPC is only with Thrift, since gRPC is normally with protobuf, and people might expect something different.

PR #233, opened 2018-05-23
Add Jaeger throttling config

isaachier:


isaachier:
Ya I agree. @yurishkuro @vprithvi any thoughts about how to test debug trace throttling in this code without mocking Jaeger components here?

integration_test.go
```diff
@@ -89,13 +89,10 @@ func verifyBaggage(ctx context.Context) error {
 		return errors.New("missing span")
 	}
 
-	spanCtx, ok := span.Context().(jaeger.SpanContext)
+	_, ok := span.Context().(jaeger.SpanContext)
 	if !ok {
```
prashantv: nit: can you inline the `ok := ` into the if,
```go
if _, ok := span.Context().(jaeger.SpanContext); !ok {
  [...]
}
```

main.go
```diff
@@ -360,7 +361,23 @@ func getTracer(opts Options, out output) (opentracing.Tracer, io.Closer) {
 		closer io.Closer
 	)
 	if opts.TOpts.Jaeger && !opts.TOpts.NoJaeger {
-		tracer, closer = jaeger.NewTracer(opts.TOpts.CallerName, jaeger.NewConstSampler(true), jaeger.NewNullReporter())
+		var err error
+		tracer, closer, err = jaeger_config.Configuration{
+			ServiceName: opts.TOpts.CallerName,
+			Throttler: &jaeger_config.ThrottlerConfig{
+				SynchronousInitialization: true,
```
prashantv: can you explain this a little -- i'm assuming it means we try to get the throttling information synchronously? if that fails (e.g., on machines without jaeger), what is the impact?

utils_for_test.go
```diff
@@ -84,6 +85,71 @@ func writeFile(t *testing.T, prefix, contents string) string {
...
+
 func getTestTracer(serviceName string) (opentracing.Tracer, io.Closer) {
-	return jaeger.NewTracer(serviceName, jaeger.NewConstSampler(true), jaeger.NewNullReporter())
+	const (
+		credits = 10
+	)
+	return getTestTracerWithCredits(serviceName, credits, nil)
+}
+
+func getTestTracerWithCredits(serviceName string, credits float64, t *testing.T) (opentracing.Tracer, io.Closer) {
```
prashantv: why is credits a float?

prashantv: make `t *testing.T` the first argument

utils_for_test.go
```diff
@@ -84,6 +85,71 @@ func writeFile(t *testing.T, prefix, contents string) string {
...
+	defer d.Unlock()
+	if d.t == nil {
+		return
+	}
+	require.True(d.t, d.numDebugSpans <= d.max, "Incorrect number of debug traces")
+}
+
 func getTestTracer(serviceName string) (opentracing.Tracer, io.Closer) {
-	return jaeger.NewTracer(serviceName, jaeger.NewConstSampler(true), jaeger.NewNullReporter())
+	const (
```
prashantv: nit: no need for var group, just `const credits = 10` is fine

utils_for_test.go
```diff
@@ -84,6 +85,71 @@ func writeFile(t *testing.T, prefix, contents string) string {
...
+}
+
+type debugTraceCounter struct {
+	sync.Mutex
+	numDebugSpans int
+	max           int
+	t             *testing.T
+}
+
+func newDebugTraceCounter(numDebugTraces int, t *testing.T) jaeger.Reporter {
```
prashantv: `t *testing.T` should be the first argument

utils_for_test.go
```diff
@@ -84,6 +85,71 @@ func writeFile(t *testing.T, prefix, contents string) string {
...
+	return &debugTraceCounter{numDebugSpans: 0, max: numDebugTraces, t: t}
+}
+
+func (d *debugTraceCounter) Report(span *jaeger.Span) {
+	if span == nil {
```
prashantv: is this actually possible?

utils_for_test.go
```diff
@@ -84,6 +85,71 @@ func writeFile(t *testing.T, prefix, contents string) string {
...
+	defer d.Unlock()
+	if span.Context().(jaeger.SpanContext).IsDebug() {
+		d.numDebugSpans += 1
+	}
+}
+
+func (d *debugTraceCounter) Close() {
+	d.Lock()
+	defer d.Unlock()
+	if d.t == nil {
```
prashantv: will this ever happen?

utils_for_test.go
```diff
@@ -84,6 +85,71 @@ func writeFile(t *testing.T, prefix, contents string) string {
...
+	sync.Mutex
+	credits float64
+}
+
+func (d *debugThrottler) IsAllowed(operation string) bool {
+	const (
+		creditsPerDebugTrace = 1
+	)
+	d.Lock()
+	defer d.Unlock()
```
prashantv: nit: can you add a blank line after the `defer`

main.go
isaachier: Synchronous initialization is required for scripting tools such as yab that will likely exist in a short-lived process. For a service, we can periodically fetch the credit information (i.e. every 10 seconds). That is not feasible in a short script, so we set this flag to indicate we need to request credits from the Jaeger agent synchronously if there are no credits available for a given operation (credits represent number of debug spans we can emit without violating the throttling API). On machines without Jaeger, this would fail the request and refuse to emit debug spans, which is fine for those machines without agents. I would have thought the initial query could take a while to determine that the agent is not running and return an error, but local tests indicate no significant lag.

integration_test.go
isaachier: Ya sure.

utils_for_test.go
isaachier: Just because the rate limiting API uses a token bucket to increment the total... floats are easier to use when you have a constant rate over time. The spending amount is always 1.

isaachier: Makes sense. Good idea.

utils_for_test.go
isaachier: Good to know. Didn't think about that.

utils_for_test.go
isaachier: 👍

main.go
yurishkuro: @isaachier add this justification as a code comment.

utils_for_test.go
isaachier: Doesn't seem to be. Will remove. I miss `assert(span != nil)` :(.

utils_for_test.go
isaachier: Sure

isaachier:
Added some comments.

utils_for_test.go
isaachier: Changing this to `int`s for internal mocks.

main.go
```diff
@@ -360,7 +361,27 @@ func getTracer(opts Options, out output) (opentracing.Tracer, io.Closer) {
 		closer io.Closer
 	)
 	if opts.TOpts.Jaeger && !opts.TOpts.NoJaeger {
-		tracer, closer = jaeger.NewTracer(opts.TOpts.CallerName, jaeger.NewConstSampler(true), jaeger.NewNullReporter())
+		// yab must set the `SynchronousInitialization` flag to indicate that
```
isaachier: Comment explaining `SynchronousInitialization`.

isaachier:
@prashantv anything else?

.travis.yml
```diff
@@ -2,8 +2,8 @@ sudo: false
 language: go
 
 go:
-    - 1.8
-    - 1.9
+    - "1.9"
```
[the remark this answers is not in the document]
isaachier: The 1.10 addition is needed for the new Zap that we use. Something about type aliasing there. If you are referring to the quotes, Travis explains that without quotes, `1.10` is a float, so it becomes `1.1`.

main.go
```diff
@@ -360,7 +361,27 @@ func getTracer(opts Options, out output) (opentracing.Tracer, io.Closer) {
...
+		// the Jaeger client must fetch debug credits synchronously. In a
+		// short-lived process like yab, the Jaeger client cannot afford to
+		// postpone the credit request for later in time.
```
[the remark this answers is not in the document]
isaachier: Sure

main.go
```diff
@@ -360,7 +361,27 @@ func getTracer(opts Options, out output) (opentracing.Tracer, io.Closer) {
...
+		var err error
+		tracer, closer, err = jaeger_config.Configuration{
```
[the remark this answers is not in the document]
isaachier: definitely

utils_for_test.go
```diff
@@ -84,6 +85,67 @@ func writeFile(t *testing.T, prefix, contents string) string {
...
+
+	if span.Context().(jaeger.SpanContext).IsDebug() {
+		d.numDebugSpans += 1
+	}
+}
+
+func (d *debugTraceCounter) Close() {
+	d.Lock()
+	defer d.Unlock()
```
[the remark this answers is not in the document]
isaachier: Will do

utils_for_test.go
```diff
@@ -84,6 +85,67 @@ func writeFile(t *testing.T, prefix, contents string) string {
...
+	if d.t == nil {
+		return
+	}
+	require.True(d.t, d.numDebugSpans <= d.max, "Incorrect number of debug traces")
+}
+
 func getTestTracer(serviceName string) (opentracing.Tracer, io.Closer) {
-	return jaeger.NewTracer(serviceName, jaeger.NewConstSampler(true), jaeger.NewNullReporter())
+	const credits = 10
+	return getTestTracerWithCredits(nil, serviceName, credits)
```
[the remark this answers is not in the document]
isaachier: OK if that is the entire usage, I am totally fine. I was under the impression there were more callers.

main.go
```diff
@@ -354,14 +355,46 @@ type noEnveloper interface {
...
 	)
 	if opts.TOpts.Jaeger && !opts.TOpts.NoJaeger {
-		tracer, closer = jaeger.NewTracer(opts.TOpts.CallerName, jaeger.NewConstSampler(true), jaeger.NewNullReporter())
-	} else if len(opts.ROpts.Baggage) > 0 {
+		return createJaegerTracer(opts, out)
+	}
+	if len(opts.ROpts.Baggage) > 0 {
 		out.Fatalf("To propagate baggage, you must opt-into a tracing client, i.e., --jaeger")
 	}
 	return tracer, closer
```
prashantv: no need for `vars` at the beginning, you can just do:
```
return opentracing.NoopTracer{}, closer
```

main.go
```diff
@@ -354,17 +354,38 @@ type noEnveloper interface {
...
+	return tracer, closer
+}
+
 func getTracer(opts Options, out output) (opentracing.Tracer, io.Closer) {
-	var (
-		tracer opentracing.Tracer = opentracing.NoopTracer{}
-		closer io.Closer
-	)
 	if opts.TOpts.Jaeger && !opts.TOpts.NoJaeger {
-		tracer, closer = jaeger.NewTracer(opts.TOpts.CallerName, jaeger.NewConstSampler(true), jaeger.NewNullReporter())
```
isaachier: @prashantv why was yab using a null reporter? Do you have a specific need to omit the root span?

isaachier: I have just changed this based on @yurishkuro's advice. Unless you mind, we might as well do this all at once.

yurishkuro: I was going to suggest not making this discussion a part of this PR, but move to #234.

isaachier: I'm going to undo this and potentially open a new PR about this.

main.go
```diff
@@ -354,17 +355,45 @@ type noEnveloper interface {
...
+		Throttler: &jaeger_config.ThrottlerConfig{
+			SynchronousInitialization: true,
+		},
+	}.NewTracer(
+		// SamplingPriority overrides sampler decision when below
+		// throttling threshold. Better to use "always false" sampling and
+		// only enable the span when we have not hit the throttling
+		// threshold.
+		jaeger_config.Sampler(jaeger.NewConstSampler(false)),
+		jaeger_config.Reporter(jaeger.NewNullReporter()),
```
yurishkuro: I think these two customizations are necessary, but can be discussed separately - #234.

isaachier:
@prashantv can you please merge this PR?

PR #238, opened 2018-08-28
Add protobuf as supported encoding

### LINKED ISSUE #237: Add protobuf support
Would love to see proto support. having one benchmarking tool for all encodings and transports would be nice.

robbertvanginkel:
Adds `proto` as supported encoding for creating requests.

Fixes #237


transport.go
```diff
@@ -148,6 +148,10 @@ func getTransport(opts TransportOptions, encoding encoding.Encoding, tracer open
 		return nil, err
 	}
 
+	if enc == encoding.Protobuf {
```
robbertvanginkel: This currently ties the protobuf encoding to grpc. Theres some that as a result of evaluating the peer list infers that host:port combinations should default to tchannel. Unfortunately, at point in code the information if the protocol was inferred or specified is lost. If this is a problem I can attempt to refactor the code to infer protocol based on encoding, but allow overrides.

encoding/protobuf.go
```diff
@@ -0,0 +1,98 @@
...
+	}, nil
+}
+
+func (p protoSerializer) Encoding() Encoding {
+	return Protobuf
+}
+
+func (p protoSerializer) Request(body []byte) (*transport.Request, error) {
+	req := dynamic.NewMessage(p.method.GetInputType())
+	err := req.UnmarshalJSON(body)
```
peter-edge: do in single line with `if` statement below

encoding/protobuf.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	return &transport.Request{
+		Method: p.fullMethodName,
+		Body:   bytes,
+	}, nil
+}
+
+func (p protoSerializer) Response(body *transport.Response) (interface{}, error) {
+	resp := dynamic.NewMessage(p.method.GetOutputType())
+	err := resp.Unmarshal(body.Body)
```
peter-edge: do in single line with `if` statement below

encoding/protobuf.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	sd, ok := dsc.(*desc.ServiceDescriptor)
+	if !ok {
+		return nil, fmt.Errorf("target server does not expose service %q", svc)
+	}
+	mtd := sd.FindMethodByName(mth)
+	if mtd == nil {
+		return nil, fmt.Errorf("service %q does not include a method named %q", svc, mth)
+	}
+	return protoSerializer{
```
peter-edge: I'd make this a pointer (and attach all methods below to `*protoSerializer`, I'm betting @akshayjshah will have input re: the memory argument however

encoding/protobuf.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+type protoSerializer struct {
+	fullMethodName string
+	method         *desc.MethodDescriptor
+}
+
+// NewProtobuf returns a protobuf serializer.
+func NewProtobuf(fullMethodName string, source ProtoDescriptorSource) (Serializer, error) {
```
peter-edge: This might be standard in this package, but if not, I'd prefer `NewProtobufSerializer`

encoding/protobuf.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/golang/protobuf/proto"
+	"github.com/jhump/protoreflect/desc"
+	"github.com/jhump/protoreflect/dynamic"
+	"github.com/yarpc/yab/transport"
+)
+
+// SplitMethod takes a method name like Service::Method and splits it
+// into Service and Method.
+func SplitMethod(fullMethod string) (svc, method string, err error) {
```
peter-edge: Is there a common file for these things? Does this need to be public?

encoding/protobuf_source.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+
+	"github.com/jhump/protoreflect/desc"
+	"github.com/jhump/protoreflect/desc/protoparse"
+)
+
+// ProtoDescriptorSource is a source of protobuf descriptor information. It can be backed by a FileDescriptorSet
+// proto (like a file generated by protoc).
+// Adapted from https://github.com/fullstorydev/grpcurl
+type ProtoDescriptorSource interface {
```
peter-edge: `Provider`?

encoding/protobuf_source.go
```diff
@@ -0,0 +1,79 @@
...
+
+// ProtoDescriptorSourceFromProtoFiles creates a DescriptorSource that is backed by the named files,
+// whose contents are Protocol Buffer source files. The given importPaths are used to locate
+// any imported files.
+func ProtoDescriptorSourceFromProtoFiles(importPaths []string, fileNames ...string) (ProtoDescriptorSource, error) {
+	p := protoparse.Parser{
+		ImportPaths:      importPaths,
+		InferImportPaths: len(importPaths) == 0,
+	}
+	fds, err := p.ParseFiles(fileNames...)
```
peter-edge: General note: I haven't gone through the `protoreflect` package carefully, but as this is going from file to `FileDescriptor` without protoc, is this not dangerous? If `protoc` is not used to create the `FileDescriptors`, a lot of testing will be needed for edge cases.

robbertvanginkel:
Discussed this internally, and we'd prefer to not tie yab to `protoc` (and all woes that come with it). Instead, yab will support two ways of getting the descriptor to transform a blob into the correct binary format: server reflection and precompiled proto descriptors.

I'll update the PR to accept precompiled proto descriptors and will follow up with one for reflection.

encoding/protobuf.go
robbertvanginkel: Isn't used outside of this package so doesn't need to be public. From what I understand the `::` separator is a yarpc concept for splitting, not a thrift thing (although the thrift encoding package has a similar method). There is a general file, so if I'm correct this is a yarpc thing and not thrift I'll move this to the encoding file and change the proto and thrift implementations to point at that.

prashantv: Is `::` the right separator for protobuf methods? It's the scheme used by Thrift to join the Thrift service name with the method name. However, I think protobuf uses `/`.

I don't mind supporting `::` as well, but we should support the grpc native naming scheme too

glide.yaml
```diff
@@ -20,6 +20,8 @@ import:
   version: ^2.14
 - package: go.uber.org/yarpc
   version: ^1.21
+- package: github.com/jhump/protoreflect
```
peter-edge: You might want to pin this to a commit for now since there are no releases.

options.go
```diff
@@ -40,8 +40,9 @@ type Options struct {
 
 // RequestOptions are request related options
 type RequestOptions struct {
-	Encoding     encoding.Encoding `short:"e" long:"encoding" description:"The encoding of the data, options are: Thrift, JSON, raw. Defaults to Thrift if the method contains '::' or a Thrift file is specified"`
+	Encoding     encoding.Encoding `short:"e" long:"encoding" description:"The encoding of the data, options are: Thrift, proto, JSON, raw. Defaults to Thrift if the method contains '::' or a Thrift file is specified"`
 	ThriftFile   string            `short:"t" long:"thrift" description:"Path of the .thrift file"`
+	ProtoSet     []string          `long:"protoset" description:"A binary file containing a compiled protobuf FileDescriptorSet."`
```
peter-edge: I know grpcurl uses `--protoset` but I'm not crazy about it, "Proto Set" isn't really a Protobuf concept. What about `--file-descriptor-set-file` or `--file-descriptor-set-bin`? And add a shorthand as well.

test.proto
```diff
@@ -0,0 +1 @@
+syntax = "proto3"; message Foo{ int32 test = 1; }; service Bar { rpc Baz(Foo) returns (Foo); }
```
peter-edge: At least format this haha

testdata/protobuf/generate.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+#
+# generate compiled binary protodescriptors, protoc required then commit files for test usage
+# 
+set -e
```
peter-edge: `set -euo pipefail`

testdata/protobuf/generate.sh
```diff
@@ -0,0 +1,23 @@
...
+
+THIS_DIR=$(cd -P -- "$(dirname -- "$0")" && pwd -P)
```
peter-edge: Quotes around `"$(...)"`

Also when referencing a variable, I always recommend i.e. `${THIS_DIR}` instead of `$THIS_DIR` but it shouldn't matter below.

testdata/protobuf/dependencies/main.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+
+import "dep.proto";
```
peter-edge: I know this seems like it might not matter, but can you do a case where the imports are in another directory?

protobuf/protobuf_source.go
```diff
@@ -0,0 +1,90 @@
...
+// proto (like a file generated by protoc).
+// Adapted from https://github.com/fullstorydev/grpcurl
+type ProtoDescriptorSource interface {
+	// FindSymbol returns a descriptor for the given fully-qualified symbol name.
+	FindSymbol(fullyQualifiedName string) (desc.Descriptor, error)
+}
+
+// ProtoDescriptorSourceFromProtoSets creates a DescriptorSource that is backed by the named files, whose contents
+// are encoded FileDescriptorSet protos.
+func ProtoDescriptorSourceFromProtoSets(fileNames ...string) (ProtoDescriptorSource, error) {
```
peter-edge: `FromFileDescriptorSetBins`?

encoding/protobuf.go
akshayjshah: This isn't that performance-sensitive, so it doesn't matter here. Unless you're being *very* picky about performance or specifically want pass-by-value semantics, pointers are more common in Go.

testdata/protobuf/dependencies/main.proto
robbertvanginkel: Would a subdirectory suffice for the case you're trying to catch or do you want to test something with multiple import paths? Eg `main.proto`, `dependencies/dep.proto` with
-  `import "dependencies/dep.proto"` in main or
- `import "dep.proto"` and protoc `-I dependencies/`

test.proto
robbertvanginkel: This wasn't supposed to be included.

options.go
robbertvanginkel: Went with `file-descriptor-set-bin` added `-F` as short. There is already a lowercase `f`, so I'm not a big fan but it already seems like there are more cases where upper and lowercase short are for completely different arguments.

glide.yaml
robbertvanginkel: Pinned. Man looking into this sent me into a spiral of reading about the upcoming v2 api for golang protobuf.

robbertvanginkel:
Pending clarification on what test case the import should cover this should be good to go.

testdata/protobuf/dependencies/main.proto
peter-edge: Both would be great, but the latter is my primary concern I think.

robbertvanginkel: Added test cases for both.

encoding/protobuf.go
```diff
@@ -0,0 +1,96 @@
...
+	resp := dynamic.NewMessage(p.method.GetOutputType())
+	if err := resp.Unmarshal(body.Body); err != nil {
+		return nil, fmt.Errorf("could not parse given response body as message of type %q: %v", p.method.GetInputType().GetFullyQualifiedName(), err)
+	}
+	str, err := resp.MarshalJSON()
+	if err != nil {
+		return nil, err
+	}
+	var objmap map[string]*json.RawMessage
+	if err = json.Unmarshal(str, &objmap); err != nil {
```
peter-edge: Missed this, shouldn’t this be jsonpb? Encoding/protobuf and the protobuf json format are not compatible

peter-edge: *encoding/json

robbertvanginkel: This is to transform a json string into a `map[string]interface{}`, so that yab can display the response on the cli.  The json string is marshalled from binary to the string by https://godoc.org/github.com/jhump/protoreflect/dynamic#Message.MarshalJSON), so I think in this case this is valid.

Could add a comment for clarity.

peter-edge: Why just just print the string from MarshalIndent? The jsonpb Marshaler has indent options too. Protobuf encoded as json cant be validly marshalled or unmarshalled using encoding/json, there are edge cases. See the protobuf json spec for all the details.

robbertvanginkel: Because I don't need to return a string but an `interface{}` type that yab understands for printing to the user (a native golang map or slice).

I need `proto bytes` -> `go map`, which I'm currently doing `bytes` -> `string` with `jsonpb` and then `string` to `map[string]interface{}` with `encoding/json`. I'm not trying to unmarshal the bytes directly with `encoding/json`. Is there a way to unmarshal bytes into a go type without going through json (for which I think I need both jsonpb and encoding/json)?

peter-edge: Unmarshalling jsonpb-marshalled strings using encoding/json has the potential to fail, it’ll work until it potentially doesn’t.

Is it difficult to just make yab understand strings to print instead of map or slice? That’s the best bet.

encoding/protobuf.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	dsc, err := source.FindSymbol(svc)
+	if err != nil {
+		return nil, fmt.Errorf("failed to query for service for symbol %q: %v", svc, err)
+	}
+	sd, ok := dsc.(*desc.ServiceDescriptor)
+	if !ok {
+		return nil, fmt.Errorf("target server does not expose service %q", svc)
+	}
+	mtd := sd.FindMethodByName(mth)
```
kriskowal: Distinguishing `mtd` / `mth` and `dsc` / `desc` don’t seem to follow any particular rule. Perhaps we should abandon brevity. If I were to guess:

- `svc`: `serviceName`
- `mth`: `methodName`
- `dsc`: `serviceSymbol`
- `sd`: `serviceDescriptor`
- `mtd`: `methodDescriptor`

encoding/protobuf.go
```diff
@@ -0,0 +1,96 @@
...
+	req := dynamic.NewMessage(p.method.GetInputType())
+	if err := req.UnmarshalJSON(body); err != nil {
+		return nil, fmt.Errorf("could not parse given request body as message of type %q: %v", p.method.GetInputType().GetFullyQualifiedName(), err)
+	}
+	bytes, err := proto.Marshal(req)
+	if err != nil {
+		return nil, fmt.Errorf("could marshal message of type %q: %v", p.method.GetInputType().GetFullyQualifiedName(), err)
+	}
+	return &transport.Request{
+		Method: fmt.Sprintf("%s::%s", p.serviceName, p.methodName),
```
kriskowal: This is fine, but string concatenation would work, `p.serviceName + "::" + p.methodName`, and avoids the surprising cost of `...interface {}`.

encoding/protobuf.go
kriskowal: I believe this solves the problem of incorporating the JSON interpretation of the Protobuf response body in the JSON envelope `{"response": …}`, which carries the headers and status code depending on the transport protocol. This is important to ensure that we can unpack yab output with `jq` or similar. I believe any JSON string generated by Protobuf should be parseable by `encoding/json`.

encoding/protobuf.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return objmap, nil
+}
+
+func (p protoSerializer) CheckSuccess(body *transport.Response) error {
+	_, err := p.Response(body)
+	return err
+}
+
+func splitMethod(fullMethod string) (svc, method string, err error) {
```
kriskowal: It will be a fine day when yab shares code like this with yarpc.

protobuf/filedescriptorsets.go
```diff
@@ -0,0 +1,82 @@
...
+	files map[string]*desc.FileDescriptor
+}
+
+func (fs *fileSource) FindSymbol(fullyQualifiedName string) (desc.Descriptor, error) {
+	for _, fd := range fs.files {
+		if dsc := fd.FindSymbol(fullyQualifiedName); dsc != nil {
+			return dsc, nil
+		}
+	}
+	return nil, fmt.Errorf("Symbol not found: %q", fullyQualifiedName)
```
abhinav: nit: error message should probably start with lower case

protobuf/filedescriptorsets_test.go
```diff
@@ -0,0 +1,99 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := NewDescriptorProviderFileDescriptorSetBins(tt.fileNames...)
+			if tt.errMsg == "" {
+				assert.NotNil(t, got)
+				assert.NoError(t, err)
+			} else {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), tt.errMsg, "%v: invalid error", tt.name)
+			}
+			if tt.lookupSymbol != "" {
```
abhinav: nit: can invert the if condition and return early to reduce indentation

transport.go
```diff
@@ -59,9 +59,9 @@ func remapLocalHost(hostPorts []string) {
 }
 
 func parsePeer(peer string) (protocol, host string) {
-	// If we get a pure host:port, then we assume tchannel.
+	// If we get a pure host:port, we return unspecified to determine based on encoding
 	if _, _, err := net.SplitHostPort(peer); err == nil && !strings.Contains(peer, "://") {
-		return "tchannel", peer
+		return "unspecified", peer
```
abhinav: nit: "unspecified" should probably be a `const`, or maybe we should just use an empty string for protocol when it's unspecified. magic special values are a bit of a smell.

encoding/protobuf_test.go
```diff
@@ -0,0 +1,167 @@
...
+				assert.Contains(t, err.Error(), tt.errMsg, "%v: invalid error", tt.desc)
+			}
+		})
+	}
+}
+
+func TestProtobufRequest(t *testing.T) {
+	tests := []struct {
+		desc   string
+		bsIn   []byte
```
prashantv: the test cases are using strings for input, so make this `input string` and then convert to `[]byte` as part of the test execution logic

encoding/protobuf_test.go
```diff
@@ -0,0 +1,167 @@
...
+		bsOut  []byte
```
prashantv: this is usually called `want`

encoding/protobuf_test.go
```diff
@@ -0,0 +1,167 @@
...
+			require.NoError(t, err, "Failed to create serializer")
+
+			got, err := serializer.Request(tt.bsIn)
+			if tt.errMsg == "" {
+				assert.NoError(t, err, "%v", tt.desc)
+				require.NotNil(t, got, "%v: Invalid request")
+				assert.Equal(t, tt.bsOut, got.Body)
+			} else {
+				assert.Nil(t, got, "%v: Error cases should not return any bytes", tt.desc)
+				require.Error(t, err, "%v", tt.desc)
```
prashantv: same as earlier, no need for `tt.desc` here (or below)

encoding/protobuf_test.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+		})
+	}
+}
+
+func TestProtobufResponse(t *testing.T) {
+	tests := []struct {
+		desc      string
+		bsIn      []byte
+		outAsJSON string
```
prashantv: consider `input` and `want`

encoding/protobuf_test.go
```diff
@@ -0,0 +1,167 @@
...
+			require.NoError(t, err, "Failed to create serializer")
+
+			response := &transport.Response{
+				Body: tt.bsIn,
+			}
+			got, err := serializer.Response(response)
+			if tt.errMsg == "" {
+				assert.NoError(t, err, "%v", tt.desc)
+				assert.NotNil(t, got, "%v: Invalid request")
+				r, _ := json.Marshal(got)
```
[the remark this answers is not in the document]
prashantv: +1

encoding/protobuf_test.go
```diff
@@ -0,0 +1,167 @@
...
+				assert.JSONEq(t, tt.outAsJSON, string(r))
+
+				err = serializer.CheckSuccess(response)
+				assert.NoError(t, err)
+			} else {
```
prashantv: drop `else` and return early instead

options.go
```diff
@@ -40,20 +40,21 @@ type Options struct {
...
-	HeadersJSON  string            `long:"headers" unquote:"false" description:"The headers in JSON or YAML format"`
-	HeadersFile  string            `long:"headers-file" description:"Path of a file containing the headers in JSON or YAML"`
-	Baggage      map[string]string `short:"B" long:"baggage" description:"Individual context baggage header as a key:value pair per flag"`
-	Health       bool              `long:"health" description:"Hit the health endpoint, Meta::health"`
-	Timeout      timeMillisFlag    `long:"timeout" default-mask:"1s" description:"The timeout for each request. E.g., 100ms, 0.5s, 1s. If no unit is specified, milliseconds are assumed."`
-	YamlTemplate string            `short:"y" long:"yaml-template" description:"Send a tchannel request specified by a YAML template"`
-	TemplateArgs map[string]string `short:"A" long:"arg" description:"A list of key-value template arguments, specified as -A foo:bar -A user:me"`
+	Encoding          encoding.Encoding `short:"e" long:"encoding" description:"The encoding of the data, options are: Thrift, proto, JSON, raw. Defaults to Thrift if the method contains '::' or a Thrift file is specified"`
```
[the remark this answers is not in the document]
prashantv: +1

options.go
```diff
@@ -40,20 +40,21 @@ type Options struct {
...
+	ThriftFile        string            `short:"t" long:"thrift" description:"Path of the .thrift file"`
+	FileDescriptorSet []string          `short:"F" long:"file-descriptor-set-bin" description:"A binary file containing a compiled protobuf FileDescriptorSet."`
+	Procedure         string            `long:"procedure" description:"The full Thrift method name (Svc::Method) to invoke"`
```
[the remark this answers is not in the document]
prashantv: if proto package names are part of this field, call that out too.

options.go
```diff
@@ -40,20 +40,21 @@ type Options struct {
...
-	Baggage      map[string]string `short:"B" long:"baggage" description:"Individual context baggage header as a key:value pair per flag"`
-	Health       bool              `long:"health" description:"Hit the health endpoint, Meta::health"`
-	Timeout      timeMillisFlag    `long:"timeout" default-mask:"1s" description:"The timeout for each request. E.g., 100ms, 0.5s, 1s. If no unit is specified, milliseconds are assumed."`
-	YamlTemplate string            `short:"y" long:"yaml-template" description:"Send a tchannel request specified by a YAML template"`
-	TemplateArgs map[string]string `short:"A" long:"arg" description:"A list of key-value template arguments, specified as -A foo:bar -A user:me"`
+	Encoding          encoding.Encoding `short:"e" long:"encoding" description:"The encoding of the data, options are: Thrift, proto, JSON, raw. Defaults to Thrift if the method contains '::' or a Thrift file is specified"`
+	ThriftFile        string            `short:"t" long:"thrift" description:"Path of the .thrift file"`
+	FileDescriptorSet []string          `short:"F" long:"file-descriptor-set-bin" description:"A binary file containing a compiled protobuf FileDescriptorSet."`
```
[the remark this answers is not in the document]
prashantv: Should this field be called `ProtoDescriptor` or something? `F` seems overly generic as an argument too, maybe `-D` for descriptor?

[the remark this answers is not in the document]
prashantv: Is it a single file or multiple? it takes a slice, but the name and description make it seem like a single value.

protobuf/filedescriptorsets.go
```diff
@@ -0,0 +1,82 @@
...
+
+// NewDescriptorProviderFileDescriptorSetBins creates a DescriptorSource that is backed by the named files, whose contents
+// are encoded FileDescriptorSet protos.
+func NewDescriptorProviderFileDescriptorSetBins(fileNames ...string) (DescriptorProvider, error) {
+	files := &descriptor.FileDescriptorSet{}
+	for _, fileName := range fileNames {
+		b, err := ioutil.ReadFile(fileName)
+		if err != nil {
+			return nil, fmt.Errorf("could not load protoset file %q: %v", fileName, err)
+		}
```
prashantv: optional nit: new line after `}` (same throughout this change)

protobuf/filedescriptorsets.go
```diff
@@ -0,0 +1,82 @@
...
+		var fs descriptor.FileDescriptorSet
+		err = proto.Unmarshal(b, &fs)
+		if err != nil {
```
prashantv: inline err, and use `:=`
```
if err := proto.Unmarshal(b, &fs); err != nil {
```

protobuf/filedescriptorsets.go
```diff
@@ -0,0 +1,82 @@
...
+			return nil, err
+		}
+	}
+	return &fileSource{files: resolved}, nil
+}
+
+func resolveFileDescriptor(unresolved map[string]*descriptor.FileDescriptorProto, resolved map[string]*desc.FileDescriptor, filename string) (*desc.FileDescriptor, error) {
+	if r, ok := resolved[filename]; ok {
+		return r, nil
+	}
```
prashantv: more new lines!

protobuf/filedescriptorsets_test.go
prashantv: +1, please do this for all tests

request.go
```diff
@@ -100,6 +101,12 @@ func NewSerializer(opts RequestOptions) (encoding.Serializer, error) {
 	e := detectEncoding(opts)
 	if e == encoding.Thrift {
 		return encoding.NewThrift(opts.ThriftFile, opts.Procedure, opts.ThriftMultiplexed)
+	} else if e == encoding.Protobuf {
```
prashantv: can you change this code to use a `switch` instead since there's multiple cases now,
```
switch e := detectEncoding(opts); e {
case encoding.Thrift: 
  [...]
case encoding.Protobuf:
  [...]
}
```

protobuf/filedescriptorsets_test.go
```diff
@@ -0,0 +1,99 @@
...
+	tests := []struct {
+		name         string
+		fileNames    []string
+		errMsg       string
+		lookupSymbol string
+		symbolErrMsg string
+	}{
+		{
+			name:         "pass",
+			fileNames:    []string{"../testdata/protobuf/simple/simple.proto.bin"},
```
prashantv: every path here is prefixed with `../testdata/protobuf/` -- maybe the test execution logic should add that prefix and keep the test-table cleaner?

transport.go
prashantv: +1

encoding/protobuf.go
prashantv: Just for my curiousity -- is this actually used by protobuf or only YARPC?

.gitattributes
```diff
@@ -0,0 +1 @@
+*.bin binary
```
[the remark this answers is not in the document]
robbertvanginkel: Don't need necessarily, but git diff was showing me compiled protodescriptors as pain text. Note that this is `.gitattributes`, not `.gitignore`: it doesn't ignore anything, it just says "binary files differ" on files that end in .bin instead of attempting to show you text diff on binary files, which I think is a nice to have here.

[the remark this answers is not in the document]
prashantv: Ahh makes sense -- feel free to add that if it helps then.

request.go
robbertvanginkel: Moved this down a few lines now. Protobuf currently doesn't support the "if no method is specified return the list of available methods" behaviour. Leaving thrift special cased here and moving the proto encoding to the already existing switch statement for json and raw encodings.

prashantv: Does protobuf not support in our current implementation, or does the underlying protoreflect package not support it? If it's just our implementation, then we should implement it as a follow-up, and structure the code assuming we will and leave a TODO. I found this method which makes it seems like it should be possible using protoreflect:
https://godoc.org/github.com/jhump/protoreflect/desc#ServiceDescriptor.GetMethods

encoding/protobuf.go
robbertvanginkel: Currently, this is a yarpc thing. The outbound request is made by the yarpc library and splits the procedure on `::` and to convert it to the grpc wire format.

request.go
robbertvanginkel: It should support it. Let me add a todo for it.

encoding/protobuf.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (p protoSerializer) Encoding() Encoding {
+	return Protobuf
+}
+
+func (p protoSerializer) Request(body []byte) (*transport.Request, error) {
+	req := dynamic.NewMessage(p.method.GetInputType())
+	if err := req.UnmarshalJSON(body); err != nil {
+		return nil, fmt.Errorf("could not parse given request body as message of type %q: %v", p.method.GetInputType().GetFullyQualifiedName(), err)
```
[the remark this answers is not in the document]
robbertvanginkel: Protoreflect doesn't is actually not to terrible. It specifies the types on which field names are invalid, tells you about what type you passed in and what it expects.

Some examples:
```
$ ./yab --file-descriptor-set-bin=/Users/robbert/health.bin raffle grpc.health.v1.Health/Check -r '{"srvice": "raffle"}' -p localhost:8000
Failed while parsing request input: could not parse given request body as message of type "grpc.health.v1.HealthCheckRequest": Message type grpc.health.v1.HealthCheckRequest has no known field named srvice
$ ./yab --file-descriptor-set-bin=/Users/robbert/health.bin raffle grpc.health.v1.Health/Check -r '{"service": 1}' -p localhost:8000
Failed while parsing request input: could not parse given request body as message of type "grpc.health.v1.HealthCheckRequest": Bad input. Expecting string. Instead got: 1.
```

I don't think a todo in yab would help much, the context for better errors is all in the protoreflect package.

PR #239, opened 2018-09-25
Support grpc reflection when using proto encoding

robbertvanginkel:
When using the proto encoding, instead of requiring the user to supply a precompiled filedescriptorset with `--file-descriptor-set-bin` support using the [grpc server reflection](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md) api to retrieve the required filedescriptors from the server.

### DIFF
--- a/main.go
+++ b/main.go
@@ -297,7 +297,7 @@ func runWithOptions(opts Options, out output, logger *zap.Logger) {
 		out.Fatalf("Failed while loading headers input: %v\n", err)
 	}
 
-	serializer, err := NewSerializer(opts.ROpts)
+	serializer, err := NewSerializer(opts)
 	if err != nil {
 		out.Fatalf("Failed while parsing input: %v\n", err)
 	}
--- a/main.go
+++ b/main.go
@@ -297,11 +297,6 @@ func runWithOptions(opts Options, out output, logger *zap.Logger) {
 		out.Fatalf("Failed while loading headers input: %v\n", err)
 	}
 
-	serializer, err := NewSerializer(opts)
-	if err != nil {
-		out.Fatalf("Failed while parsing input: %v\n", err)
-	}
-
 	if opts.TOpts.CallerName != "" {
 		if _, ok := warningCallerNames[opts.TOpts.CallerName]; ok {
 			// TODO: when logger is hooked up this should use the WARN level message
@@ -315,6 +310,11 @@ func runWithOptions(opts Options, out output, logger *zap.Logger) {
 		}
 	} else {
 		opts.TOpts.CallerName = "yab-" + os.Getenv("USER")
+	}
+
+	serializer, err := NewSerializer(opts)
+	if err != nil {
+		out.Fatalf("Failed while parsing input: %v\n", err)
 	}
 
 	tracer, closer := getTracer(opts, out)
--- a/protobuf/source_reflection.go
+++ b/protobuf/source_reflection.go
@@ -1,0 +1,71 @@
+package protobuf
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	"github.com/jhump/protoreflect/desc"
+	"github.com/jhump/protoreflect/grpcreflect"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+	rpb "google.golang.org/grpc/reflection/grpc_reflection_v1alpha"
+	"google.golang.org/grpc/resolver"
+	"google.golang.org/grpc/resolver/manual"
+)
+
+var (
+	// ErrorCouldNotDialReflectionServer is an error for when the reflection server could not be reached.
+	ErrorCouldNotDialReflectionServer = errors.New("could not reach reflection server")
+)
+
+// ReflectionArgs are args for constructing a DescriptorProvider that reaches out to a reflection server.
+type ReflectionArgs struct {
+	Caller  string
+	Service string
+	Peers   []string
+	Timeout time.Duration
+}
+
+// NewDescriptorProviderReflection returns a DescriptorProvider DescriptorProvider that reaches out to
+// a reflection server to access filedescriptors.
+func NewDescriptorProviderReflection(args ReflectionArgs) (DescriptorProvider, error) {
+	metadataContext := metadata.NewOutgoingContext(context.Background(),
+		map[string][]string{
+			"rpc-caller":   []string{args.Caller},
+			"rpc-service":  []string{args.Service},
+			"rpc-encoding": []string{"proto"},
+		})
+	r, deregisterScheme := manual.GenerateAndRegisterManualResolver()
+	defer deregisterScheme()
+	peers := make([]resolver.Address, len(args.Peers))
+	for i, p := range args.Peers {
+		peers[i] = resolver.Address{Addr: p, Type: resolver.Backend}
+	}
+	r.InitialAddrs(peers)
+
+	conn, err := grpc.DialContext(context.Background(),
+		r.Scheme()+":///any.peers.registered.for.this.scheme",
+		grpc.WithTimeout(args.Timeout),
+		grpc.WithBlock(),
+		grpc.WithInsecure())
+	if err != nil {
+		return nil, ErrorCouldNotDialReflectionServer
+	}
+	pbClient := rpb.NewServerReflectionClient(conn)
+	return &grpcreflectSource{
+		client: grpcreflect.NewClient(metadataContext, pbClient),
+	}, nil
+}
+
+type grpcreflectSource struct {
+	client *grpcreflect.Client
+}
+
+func (s *grpcreflectSource) FindSymbol(fullyQualifiedName string) (desc.Descriptor, error) {
+	file, err := s.client.FileContainingSymbol(fullyQualifiedName)
+	if err != nil {
+		return nil, err
+	}
+	return file.FindSymbol(fullyQualifiedName), nil
+}
--- a/protobuf/source_reflection.go
+++ b/protobuf/source_reflection.go
@@ -2,7 +2,7 @@ package protobuf
 
 import (
 	"context"
-	"errors"
+	"fmt"
 	"time"
 
 	"github.com/jhump/protoreflect/desc"
@@ -14,11 +14,6 @@ import (
 	"google.golang.org/grpc/resolver/manual"
 )
 
-var (
-	// ErrorCouldNotDialReflectionServer is an error for when the reflection server could not be reached.
-	ErrorCouldNotDialReflectionServer = errors.New("could not reach reflection server")
-)
-
 // ReflectionArgs are args for constructing a DescriptorProvider that reaches out to a reflection server.
 type ReflectionArgs struct {
 	Caller  string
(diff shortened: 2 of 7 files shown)


protobuf/source_reflection.go
```diff
@@ -0,0 +1,71 @@
...
+
+// ReflectionArgs are args for constructing a DescriptorProvider that reaches out to a reflection server.
+type ReflectionArgs struct {
+	Caller  string
+	Service string
+	Peers   []string
+	Timeout time.Duration
+}
+
+// NewDescriptorProviderReflection returns a DescriptorProvider DescriptorProvider that reaches out to
```
amckinney: `DescriptorProvider` stutter.

protobuf/source_reflection.go
```diff
@@ -0,0 +1,71 @@
...
+// a reflection server to access filedescriptors.
```
amckinney: nit: `filedescriptors` -> `file descriptors`.

protobuf/source_reflection.go
```diff
@@ -0,0 +1,71 @@
...
+func NewDescriptorProviderReflection(args ReflectionArgs) (DescriptorProvider, error) {
+	metadataContext := metadata.NewOutgoingContext(context.Background(),
```
amckinney: Should we apply the configured timeout to our `context.Context` here as well?

protobuf/source_reflection.go
```diff
@@ -0,0 +1,71 @@
...
+		map[string][]string{
+			"rpc-caller":   []string{args.Caller},
+			"rpc-service":  []string{args.Service},
+			"rpc-encoding": []string{"proto"},
+		})
+	r, deregisterScheme := manual.GenerateAndRegisterManualResolver()
+	defer deregisterScheme()
```
amckinney: nit: should we rename `deregisterScheme` to `unregister` or `cleanup`?

protobuf/source_reflection_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+)
+
+func TestReflection(t *testing.T) {
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	defer ln.Close()
```
amckinney: nit: I'm not sure that we should assume the result is non-nil here. Should we instead re-order this to:
```
require.NoError(t, err)
defer ln.Close()
```

protobuf/source_reflection_test.go
```diff
@@ -0,0 +1,74 @@
...
+	go s.Serve(ln)
+
+	source, err := NewDescriptorProviderReflection(ReflectionArgs{
+		Timeout: time.Second,
+		Peers:   []string{ln.Addr().String()},
+	})
+	assert.NoError(t, err)
+	assert.NotNil(t, source)
+
+	result, err := source.FindSymbol("grpc.reflection.v1alpha.ServerReflectionRequest")
```
amckinney: I recall discussing offline something related to the fully-qualified name. Did we confirm that its impossible for their to be multiple types with the same fully qualified name, such as:

```
// foo/foo.proto

package foo;

message foo {
  bar name = 1;
}

message bar {}
```

```
// foo/foo2.proto

package foo.foo;

import "foo/foo.proto"

message bar {}
```

Where `foo.foo.bar` represents both the nested-message in `foo/foo.proto` and the top-level message in `foo/foo2.proto`.

protobuf/source_reflection_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	result, err = source.FindSymbol("wat")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "Symbol not found: wat")
+	assert.Nil(t, result)
+}
+
+func TestReflectionMultiplePeers(t *testing.T) {
+	listenRefuser, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err, "failed to listen on a port")
```
amckinney: Should we `defer listenRefuser.Close()` here?

protobuf/source_reflection_test.go
```diff
@@ -0,0 +1,74 @@
...
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err, "failed to listen on a port")
+	ln.Close()
+
+	got, err := NewDescriptorProviderReflection(ReflectionArgs{
+		Timeout: time.Second,
+		Peers:   []string{ln.Addr().String()},
+	})
+
+	assert.Error(t, err)
```
amckinney: Is it possible to assert on a specific error message we expect? If it contains variable information, we could use `assert.Contains`.

protobuf/source_reflection_test.go
```diff
@@ -0,0 +1,74 @@
...
+	assert.Nil(t, got)
```
amckinney: A variety of these tests assert that the response is `nil` when the error is non-nil. I think we can drop these cases since we only really care whether or not if there was an error. If the error is `non-nil`, the result of the response should be treated as irrelevant and can be discarded.

request_test.go
```diff
@@ -297,6 +318,25 @@ func TestNewSerializer(t *testing.T) {
 		})
 	}
 }
+func TestNewSerializerProtobufReflection(t *testing.T) {
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	defer ln.Close()
```
amckinney: Similar to note above about ordering here.

request.go
```diff
@@ -86,44 +86,60 @@ func getHeaders(inline, file string, override map[string]string) (map[string]str
...
+		return encoding.NewJSON(opts.ROpts.Procedure), nil
 	case encoding.Raw:
-		return encoding.NewRaw(opts.Procedure), nil
+		return encoding.NewRaw(opts.ROpts.Procedure), nil
 	}
 
 	return nil, errUnrecognizedEncoding
 }
 
+func newProtoDescriptorProvider(ROpts RequestOptions, TOpts TransportOptions) (protobuf.DescriptorProvider, error) {
```
amckinney: nit: I noticed other places in the code refer to these variables as `ropts` and `topts`. Could we update them here as well? Capitalized parameters looks a little strange.

protobuf/source_reflection.go
AlexAPB: nit @amckinney `FileDescriptorSet`?

protobuf/source_reflection.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	r.InitialAddrs(peers)
+
+	conn, err := grpc.DialContext(context.Background(),
+		r.Scheme()+":///any.peers.registered.for.this.scheme",
+		grpc.WithTimeout(args.Timeout),
+		grpc.WithBlock(),
+		grpc.WithInsecure())
+	if err != nil {
+		return nil, ErrorCouldNotDialReflectionServer
```
AlexAPB: can we concatenate/extend the error so we can give the user more details on failed connections?

protobuf/source_reflection_test.go
```diff
@@ -0,0 +1,74 @@
...
+	require.NoError(t, err)
+	s := grpc.NewServer()
+	reflection.Register(s)
+	go s.Serve(ln)
+
+	source, err := NewDescriptorProviderReflection(ReflectionArgs{
+		Timeout: time.Second,
+		Peers:   []string{ln.Addr().String()},
+	})
+	assert.NoError(t, err)
```
AlexAPB: I think we want this to be `require`, ie it doesn't make much sense continuing the test if we have a failure here

protobuf/source_reflection.go
```diff
@@ -0,0 +1,71 @@
...
+	r, deregisterScheme := manual.GenerateAndRegisterManualResolver()
+	defer deregisterScheme()
+	peers := make([]resolver.Address, len(args.Peers))
+	for i, p := range args.Peers {
+		peers[i] = resolver.Address{Addr: p, Type: resolver.Backend}
+	}
+	r.InitialAddrs(peers)
+
+	conn, err := grpc.DialContext(context.Background(),
+		r.Scheme()+":///any.peers.registered.for.this.scheme",
```
AlexAPB: Is this a leftover TODO? This doesn't look right.
 ```go
+":///any.peers.registered.for.this.scheme"
```

File: protobuf/source_reflection_test.go (hunk 9 above)
Comment (AlexAPB) [reply protobuf/source_reflection_test.go]: I'm fairly indifferent, but I lean in favour of keeping them because it ensures that we're being idomatic and that no-one will rely on undefined behaviour (of eg a semi initialized struct)

File: main.go [hunk 15]
```diff
@@ -317,6 +312,11 @@ func runWithOptions(opts Options, out output, logger *zap.Logger) {
 		opts.TOpts.CallerName = "yab-" + os.Getenv("USER")
 	}
 
+	serializer, err := NewSerializer(opts)
```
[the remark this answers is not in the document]
robbertvanginkel: It is necessary, as when using the reflection serializer you need `opts.TOpts.CallerName`.

It doesn't change any behaviour in yab as only `opts.TOpts` is modified and before this change only `opts.ROpts` was passed to `NewSerializer`.

protobuf/source_reflection.go
robbertvanginkel: I'll make it file descriptors, as technically according to the api you get multiple of those rather than one set.

protobuf/source_reflection.go
robbertvanginkel: Maybe, I'm still a bit fuzzy on stream and timeout behaviour. Any timeout that would be set would apply to opening and closing the whole stream for all the reflection related operations. Additionally if we'd use a `context.WithTimeout` here that would set a timeout from the construction of this object rather then from when we actually start making requests.

I'd say that having a timeout on opening a connection to the reflection server is enough, if that passes then the retrieval of the required file descriptors currently doesn't have a timeout. I think that behaviour is similar to not having a timeout on thrift compilation.

protobuf/source_reflection.go
robbertvanginkel: I actually remember reading through https://english.stackexchange.com/questions/25931/unregister-vs-deregister when I was deciding on `de-` vs `un-` register. Thought this fit best.

protobuf/source_reflection.go
robbertvanginkel: Yeah its a bit funky. The only way to get the dialer to attempt to round robin dial multiple host:port pairs is to register them with a specific scheme, then instruct the dialer to resolve some random url on that scheme.

I could remove the `any.peers.registered.for.this.scheme` part, but the `+":///"` has to stay.

protobuf/source_reflection.go
robbertvanginkel: yas

protobuf/source_reflection_test.go
robbertvanginkel: Our implementation of the reflection server won't start if there's ambiguity in the fully qualified names that are registered with the server.

protobuf/source_reflection_test.go
robbertvanginkel: I prefer to check if you error that you return nil, just so there's no pointers to half initialised objects lying around.

amckinney: Yea, works for me!

protobuf/source_reflection.go
amckinney: Nice, thanks for the reference! Seems reasonable to me.

protobuf/source_reflection.go
amckinney: Yea, that seems reasonable to me. We can always update this later if it ever becomes an issue.

PR #240, opened 2018-10-05
Adress some late codereview feedback on #238

robbertvanginkel:
There were some late comments on #238, addressing them here.

### DIFF
--- a/encoding/protobuf.go
+++ b/encoding/protobuf.go
@@ -20,26 +20,26 @@ type protoSerializer struct {
 
 // NewProtobuf returns a protobuf serializer.
 func NewProtobuf(fullMethodName string, source protobuf.DescriptorProvider) (Serializer, error) {
-	svc, mth, err := splitMethod(fullMethodName)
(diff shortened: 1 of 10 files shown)


encoding/protobuf.go
```diff
@@ -71,11 +71,11 @@ func (p protoSerializer) Response(body *transport.Response) (interface{}, error)
 	if err != nil {
 		return nil, err
 	}
-	var objmap map[string]*json.RawMessage
-	if err = json.Unmarshal(str, &objmap); err != nil {
+	var unmarshaledJson json.RawMessage
```
abhinav: nittiest of nits: per Go naming conventions, this should be unmarshaledJSON :)

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

transport.go
```diff
@@ -37,6 +37,8 @@ import (
 	"github.com/uber/tchannel-go"
 )
 
+const transportUnspecified = "unspecified"
```
abhinav: unused/can delete?

PR #241, opened 2018-10-09
Print available proto methods for unrecognized arg

robbertvanginkel:
Thrift currently has a feature that if the method specified in the
procedure is not present (or not available on the server), it prints the
available methods as a hint to the user. This brings the proto
implementation to feature parity.

Example:
```
$ ./yab --file-descriptor-set-bin=/Users/robbert/health.bin raffle grpc.health.v1.Health -r '{"service": "raffle"}' -p localhost:8000
Failed while parsing input: service "grpc.health.v1.Health" does not include a method named "", available methods:
	grpc.health.v1.Health/Check
```

### DIFF
--- a/encoding/protobuf.go
+++ b/encoding/protobuf.go
@@ -108,7 +108,10 @@ func findProtoMethodDescriptor(s *desc.ServiceDescriptor, m string) (*desc.Metho
 		for i, method := range s.GetMethods() {
 			available[i] = s.GetFullyQualifiedName() + "/" + method.GetName()
 		}
-		errMsg := fmt.Sprintf("service %q does not include a method named %q", s.GetFullyQualifiedName(), m)
+		errMsg := "no proto method specified, specify --method package.Service/Method"
+		if m != "" {
+			errMsg = fmt.Sprintf("service %q does not include a method named %q", s.GetFullyQualifiedName(), m)
+		}
 		return nil, notFoundError{errMsg + ", available methods:", available}
 	}
 	return methodDescriptor, nil
--- a/encoding/protobuf_test.go
+++ b/encoding/protobuf_test.go
@@ -2,12 +2,14 @@ package encoding
 
 import (
 	"encoding/json"
+	"fmt"
 	"testing"
+
+	"github.com/yarpc/yab/protobuf"
+	"github.com/yarpc/yab/transport"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"github.com/yarpc/yab/protobuf"
-	"github.com/yarpc/yab/transport"
 )
 
 func TestNewProtobuf(t *testing.T) {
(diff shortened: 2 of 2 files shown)


encoding/protobuf_test.go
```diff
@@ -22,11 +22,16 @@ func TestNewProtobuf(t *testing.T) {
...
 		{
 			desc:   "no method",
+			method: "Bar",
+			errMsg: "service \"Bar\" does not include a method named \"\", available methods:\n\tBar/Baz",
+		},
+		{
+			desc:   "missing method for service",
 			method: "Bar/baq",
-			errMsg: `service "Bar" does not include a method named "baq"`,
+			errMsg: "service \"Bar\" does not include a method named \"baq\", available methods:\n\tBar/Baz",
```
prashantv: nit: use backtick strings `service "Bar" does not include [...]`

same above

encoding/protobuf_test.go
```diff
@@ -22,11 +22,16 @@ func TestNewProtobuf(t *testing.T) {
 		},
 		{
 			desc:   "no method",
+			method: "Bar",
+			errMsg: "service \"Bar\" does not include a method named \"\", available methods:\n\tBar/Baz",
```
prashantv: can we make this error nicer, specifically out that no method was spcified

encoding/protobuf.go
```diff
@@ -39,8 +39,12 @@ func NewProtobuf(fullMethodName string, source protobuf.DescriptorProvider) (Ser
 
 	methodDescriptor := serviceDescriptor.FindMethodByName(methodName)
 	if methodDescriptor == nil {
-		//TODO: return a list of available methods
-		return nil, fmt.Errorf("service %q does not include a method named %q", serviceName, methodName)
+		available := make([]string, len(serviceDescriptor.GetMethods()))
```
prashantv: since the logic to build the message is adding noise to this function, i'd suggest adding a `findMethod` private method that will do the `serviceDescriptor.FindMethodByName` and build the error message as you do now

encoding/protobuf_test.go
robbertvanginkel: I originally used the backticks but raw formatting with the tab and newline isn't much of an improvement:

```
                        method: "Bar",
-                       errMsg: "service \"Bar\" does not include a method named \"\", available methods:\n\tBar/Baz",
+                       errMsg: `service "Bar" does not include a method named "", available methods:
+       Bar/Baz`,
                },
```

encoding/protobuf.go
robbertvanginkel: Will change.

encoding/protobuf_test.go
prashantv: Makes sense, another option,
`fmt.Sprintf("service %q does not include a method named %q [...]", "Bar", [...])`

What you have now is fine too

PR #248, opened 2019-01-31
Prepare for 0.14.2 release

prashantv:


robbertvanginkel:
Does this also need the change in `version.go`?

prashantv:
Yep, updated

PR #255, opened 2019-04-16
Bump Go versions to 1.11.x and 1.12.x for Travis

prashantv:


abhinav:
Build appears broken because of the same issue as https://github.com/yarpc/yarpc-go/pull/1708/commits/9e89ac51d8617feca51c799481ccf1451587b2e3.

The workaround was to filter out `//line` directives from the code generated by Ragel.

https://github.com/yarpc/yarpc-go/blob/778ba3cea5f10b767aa2cd90694b6bcd97e2ae0b/etc/bin/generate.sh#L60-L65

prashantv:
> The workaround was to filter out //line directives from the code generated by Ragel.

Thanks @abhinav, pulled in the same fix here.

PR #258, opened 2019-04-17
Bash autocomplete script

alshopov:
It relies on:
  * autocomplete functionality from go-flags package
    https://godoc.org/github.com/jessevdk/go-flags#hdr-Completion
  * functions ('have') from bash-completion in order to allow
    user to explicitly opt in
    https://salsa.debian.org/debian/bash-completion
  * script being placed in the etc/bash_completion.d directory belonging
    to bash-completion - this is similar to the man page in share

Packaging will need to properly place files

### DIFF
--- a/etc/bash_completion.d/yab
+++ b/etc/bash_completion.d/yab
@@ -1,4 +1,8 @@
 # bash completion for yab executable
+# based on https://godoc.org/github.com/jessevdk/go-flags#hdr-Completion
+# depends on https://salsa.debian.org/debian/bash-completion
+
+
 have yab &&
 _yab()
 {
--- a/etc/bash_completion.d/yab
+++ b/etc/bash_completion.d/yab
@@ -1,0 +1,16 @@
+# bash completion for yab executable
+have yab &&
+_yab()
+{
+    # All arguments except the first one
+    args=("${COMP_WORDS[@]:1:$COMP_CWORD}")
+
+    # Only split on newlines
+    local IFS=$'\n'
+
+    # Call completion (note that the first element of COMP_WORDS is
+    # the executable itself)
+    COMPREPLY=($(GO_FLAGS_COMPLETION=1 ${COMP_WORDS[0]} "${args[@]}"))
+    return 0
+} &&
+complete -F _yab yab


etc/bash_completion.d/yab
```diff
@@ -0,0 +1,16 @@
+# bash completion for yab executable
```
prashantv: please mention that this is based on the example from
https://godoc.org/github.com/jessevdk/go-flags#hdr-Completion

etc/bash_completion.d/yab
```diff
@@ -0,0 +1,16 @@
...
+have yab &&
```
prashantv: nit: newline (probably a couple) separating the header from the body

etc/bash_completion.d/yab
```diff
@@ -0,0 +1,16 @@
...
+    args=("${COMP_WORDS[@]:1:$COMP_CWORD}")
+
+    # Only split on newlines
+    local IFS=$'\n'
+
+    # Call completion (note that the first element of COMP_WORDS is
+    # the executable itself)
+    COMPREPLY=($(GO_FLAGS_COMPLETION=1 ${COMP_WORDS[0]} "${args[@]}"))
+    return 0
+} &&
```
prashantv: assume `&&` here is to make sure we only register the function if `yab` is found?

can we instead exit early if it's not found and avoid the `&&`?

etc/bash_completion.d/yab
prashantv: what is `have yab`?

alshopov: The meaning is: is there a `yab` executable somewhere in `$PATH`. This prevents the autocomplete functionality from loading if there is no yab executable available. Too many autocomplete functions make the shell run slower. This also allows for splitting the shipping of the yab binary and its autocomplete functionality.
`have` is documented [here](https://salsa.debian.org/debian/bash-completion/blob/master/bash_completion#L113). While it is deprecated in newer versions, brew [uses older versions](https://github.com/Homebrew/homebrew-core/blob/master/Formula/bash-completion.rb#L6) so for now I am stuck with it.

alshopov: Done

etc/bash_completion.d/yab
alshopov: Done

etc/bash_completion.d/yab
alshopov: Are you really sure you want this?

1. This is the style of bash-completion. Should we break from that?
2. This is shell we are talking about. `&&` uses the status code not what the function prints (which is what it actually returns). The check will be longer and not buy anything:
```
if [ $? -eq 0 ]
then
    # register autocomplete
else
    # do nothing
fi
```
3. We cannot `exit`. This whole script is executing by sourcing - within the invoking shell rather than in a subshell. An `exit` will close the shell and prevent you from running. If you mean _`return`_ - again, it will be uglier

prashantv: If this is the style that bash completions use, then it seems fine.

etc/bash_completion.d/yab
prashantv: Useful stack question about this
https://stackoverflow.com/questions/12874920/have-keyword-for-bash-completion/12874971#12874971

Since this is a standard check for bash, seems fine.

PR #261, opened 2019-05-09
thrift: Treat null map values as ignore field/key

prashantv:
Currently, if a value is specified as null, it causes an error
parsing (e.g., cannot parse null as i32). This limits how useful
yab templates can be, as any specified field must always be set to
a value.

However, what we'd like is a YAML template with something like:

```
request:
  minVersion: ${minVersion:null}
  maxVersion: ${maxVersion:null}
```

We may only want to specify one or the other condition, but today this
template would cause errors and it would try to parse "null". Instead we
should ignore null, as null is not a value we can propagate on the wire.

### DIFF
--- a/thrift/to_wire.go
+++ b/thrift/to_wire.go
@@ -137,6 +137,11 @@ func mapToValue(keySpec, valueSpec compile.TypeSpec, value interface{}) (wire.Ma
 
 	items := make([]wire.MapItem, 0, len(valueMap))
 	for k, v := range valueMap {
+		if v == nil {
+			// If no value is provided, then ignore this key.
+			continue
+		}
+
 		keyValue := convertMapKey(keySpec, k)
 		kw, err := toWireValue(keySpec, keyValue)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


thrift/to_wire.go
```diff
@@ -137,6 +137,11 @@ func mapToValue(keySpec, valueSpec compile.TypeSpec, value interface{}) (wire.Ma
 
 	items := make([]wire.MapItem, 0, len(valueMap))
 	for k, v := range valueMap {
+		if v == nil {
+			// If no value is provided, then ignore this key.
```
abhinav: Do you think it's worth adding similar logic to `listToValue` as well?

prashantv: Hmm good question, it is nice to be consistent, but ignoring null list values has an impact on other elements (since they're ordered, it affects indexes). With maps/structs, ignoring a single field has no impact on other fields so it seems safer.

I could be convinced that the index of an item in a list isn't important though -- it's just the order that matters, and that is preserved.

abhinav: I don't feel strongly about supporting null-skipping for lists but
yeah, only the ordering matters. Indexes aren't sent over the wire.
And similar to the rest of this change: `null`/`nil` isn't a valid
value for list items either.

We can defer that to a future change if you're unsure about it.

PR #262, opened 2019-05-14
Add option to specify HTTP method

### LINKED ISSUE #252: Add support for other HTTP methods
Yab defaults to `POST`, used by YARPC, but we should support alternative methods similar to how we allow modifying URLs and headers.

prashantv:
Yab supports "raw" requests to HTTP for REST servers, but it only uses
POST requests. Add a flag to control the HTTP method used to make the
request.

Fixes #252

### DIFF
--- a/options.go
+++ b/options.go
@@ -83,6 +83,7 @@ type TransportOptions struct {
 	ShardKey         string            `long:"sk" description:"The shard key is a transport header that clues where to send a request within a clustered traffic group."`
 	Jaeger           bool              `long:"jaeger" description:"Use the Jaeger tracing client to send Uber style traces and baggage headers"`
 	TransportHeaders map[string]string `short:"T" long:"topt" description:"Transport options for TChannel, protocol headers for HTTP"`
+	HTTPMethod       string            `long:"http-method" description:"The HTTP method to use"`
 
 	// This is a hack to work around go-flags not allowing disabling flags:
 	// https://github.com/jessevdk/go-flags/issues/191
@@ -114,6 +115,7 @@ func newOptions() *Options {
 
 	// Defaults
 	opts.ROpts.Timeout = timeMillisFlag(time.Second)
+	opts.TOpts.HTTPMethod = "POST"
 
 	// Set flag aliases
 	opts.ROpts.MethodName.dest = &opts.ROpts.Procedure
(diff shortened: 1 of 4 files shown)


options.go
```diff
@@ -83,6 +83,7 @@ type TransportOptions struct {
 	ShardKey         string            `long:"sk" description:"The shard key is a transport header that clues where to send a request within a clustered traffic group."`
 	Jaeger           bool              `long:"jaeger" description:"Use the Jaeger tracing client to send Uber style traces and baggage headers"`
 	TransportHeaders map[string]string `short:"T" long:"topt" description:"Transport options for TChannel, protocol headers for HTTP"`
+	HTTPMethod       string            `long:"http-method" description:"The HTTP method to use"`
```
abhinav: Should we mention that it defaults to POST here? Also, that this only applies if the transport is HTTP?

transport/http.go
```diff
@@ -65,6 +66,9 @@ func NewHTTP(opts HTTPOptions) (Transport, error) {
 	if opts.TargetService == "" {
 		return nil, errMissingTarget
 	}
+	if opts.Method == "" {
+		opts.Method = "POST"
+	}
```
abhinav: Default being applied in two places is a smell. Ideally, either just TransportOptions applies the default or just HTTPOptions. I'm leaning towards the latter.

options.go
prashantv: go-flags has this useful feature where if you set the defaults before you parse the struct, it shows them automatically, e.g., from the help output:
```
      --http-method=             The HTTP method to use (default: POST)
```

It's a little nicer to use that rather than `default:"POST"` since the default won't be applied if you create the command without parsing using go-flags. With the `newCommand` approach, the main package tests can just set Go fields and avoid parsing, which simplifies tests and provides compiler verification that the field name is correct.

transport/http.go
prashantv: Yeah this definitely isn't ideal, but the goal is:
 * Default to POST in the flags (and have it displayed in the yab help)
 * Match the default if not specified here.

The options are:
 * Only setting it in `NewHTTP`, in which case the CLI help page won't know about it
 * Only set it in `newCommand` or the flag struct, in which case code that does `NewHTTP` without setting a method will use `""`, which defaults to `GET`.

Neither seems great, having 2 different default values is confusing, while not having the default shown in the help page isn't ideal either.

options.go
abhinav: Ah, I didn't realize go-flags will show the defaults you've set below.

transport/http.go
abhinav: Cool, that sounds reasonable.

PR #263, opened 2019-05-15
Introduce yamlalias to simplify aliases in templates

prashantv:
For yab templates, we support a few aliases to simplify
usage for users. E.g., we support shardKey, shardkey, shard-key or sk.

This requires a lot of manual wiring up to work. Instead, introduce
a yamlalias package which will use reflection to generate a new struct
that contains additional fields for each alias.

### DIFF
--- a/internal/yamlalias/wrap.go
+++ b/internal/yamlalias/wrap.go
@@ -37,8 +37,8 @@ func UnmarshalStrict(in []byte, out interface{}) error {
 	return yaml.UnmarshalStrict(in, addAliases(out))
 }
 
-// addAliases generates a new struct that contains all exported fields
-// from the passed in struct, but adds additional fields for any
+// addAliases generates a new struct for unmarshalling that contains all exported
+// fields from the passed in struct, but also adds additional fields for any
 // aliases.
 //
 // Given a struct such as:
@@ -65,12 +65,13 @@ func UnmarshalStrict(in []byte, out interface{}) error {
 // out.UserNameYamlAlias1 = &dest.UserName
 // out.UserNameYamlAlias2 = &dest.UserName
 // out.TTL = &dest.TTL
+// return out
 //
 // The YAML library respects existing pointers when unmarshalling, and
 // does not replace them:
 // https://gist.github.com/prashantv/fa4f92b4b95f936d68495be250ed3506
-func addAliases(v interface{}) interface{} {
-	rv := reflect.ValueOf(v).Elem()
+func addAliases(dest interface{}) interface{} {
+	rv := reflect.ValueOf(dest).Elem()
 	rt := rv.Type()
 
 	fields := make([]reflect.StructField, 0, rt.NumField())
@@ -113,10 +114,10 @@ func addAliases(v interface{}) interface{} {
 	// The struct we generated has pointers to the original fields.
 	// Set all the pointers to the fields in the original struct.
 	generated := reflect.StructOf(fields)
-	generatedV := reflect.New(generated)
+	out := reflect.New(generated)
 	for i, dest := range dests {
-		generatedV.Elem().Field(i).Set(dest)
+		out.Elem().Field(i).Set(dest)
 	}
 
-	return generatedV.Interface()
+	return out.Interface()
 }
--- a/internal/yamlalias/wrap.go
+++ b/internal/yamlalias/wrap.go
@@ -52,7 +52,7 @@ func UnmarshalStrict(in []byte, out interface{}) error {
 // original struct and a new field for each alias.
 //
 //   struct {
-//     UserName string `yaml:"username" yaml-aliases:"userName,user-name"`
+//     UserName *string `yaml:"username" yaml-aliases:"userName,user-name"`
 //     UserNameYamlAlias1 *string `yaml:"userName"`
 //     UserNameYamlAlias2 *string `yaml:"user-name"`
 //     TTL *time.Duration `yaml:"ttl"`
--- a/internal/yamlalias/wrap.go
+++ b/internal/yamlalias/wrap.go
@@ -1,0 +1,93 @@
+// Package yamlalias adds support for aliases when parsing YAML.
+//
+// Aliases are specified as a comma-separated list specified via
+// "yaml-aliases" struct tag.
+//
+// Example:
+//
+//   type Config struct {
+//     UserName string `yaml:"username" yaml-aliases:"userName,user-name"`
+//   }
+//
+// This will parse the following YAML examples in the same way:
+//   {"username": "John Doe"}
+//   {"userName": "John Doe"}
+//   {"user-name": "John Doe"}
+//
+// Note: It only adds aliases for top-level fields in the struct.
+package yamlalias
+
+import (
+	"fmt"
+	"reflect"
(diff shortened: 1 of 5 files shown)


internal/yamlalias/wrap.go
```diff
@@ -0,0 +1,93 @@
...
+// UnmarshalStrict is a wrapper for yaml.UnmarshalStrict but adds
+// support for any yaml-aliases specified in out.
+func UnmarshalStrict(in []byte, out interface{}) error {
+	return yaml.UnmarshalStrict(in, addAliases(out))
+}
+
+// addAliases generates a new struct that contains all exported fields
+// from the passed in struct, but adds additional fields for any
+// aliases.
+func addAliases(v interface{}) interface{} {
```
abhinav: I think it's a bit difficult to follow what this is doing in Go terms. Some comments referencing examples would help.

I *think* what this does is, given:

```
var cfg struct {
  UserName string `yaml:"username" yaml-aliases:"userName,user-name"`
}
```

It generates:

```
var dest struct {
  UserName *string `yaml:"username"`
  UserNameYamlAlias1 *string `yaml:"userName"`
  UserNameYamlAlias2 *string `yaml:"user-name"`
}
dest.UserName = &cfg.UserName
dest.UserNameYamlAlias1 = &cfg.UserName
dest.UserNameYamlAlias2 = &cfg.UserName
```

If that's correct, it's surprising to me that the YAML library respects the
pointer and fills the right destination rather than replacing the pointer.
Worth calling out?

prashantv: That's exactly right, added documentation as suggested. Here's the repro for yaml respecting existing pointers:
https://play.golang.org/p/F1AoTeUVkAW

I've added a link to a gist in the docs, but open to alternative suggestions.

The test validate all the behaviour in an end-to-end fashion, so not sure it's worth adding a test specific just for the YAML part.

abhinav: A test for just the YAML part isn't necessary. I just wasn't expecting this to work.
Looking at the yaml.Unmarshal docs, it appears deliberate so we can rely on it.

> If an internal pointer within a struct is not initialized, the yaml package will initialize it [..]

https://godoc.org/gopkg.in/yaml.v2#Unmarshal

internal/yamlalias/wrap.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// addAliases generates a new struct that contains all exported fields
+// from the passed in struct, but adds additional fields for any
+// aliases.
+//
+// Given a struct such as:
+//
+//   type Config struct {
+//     UserName string `yaml:"username" yaml-aliases:"userName,user-name"`
```
abhinav: Worth adding what happens for non-alias cases.

```suggestion
//     UserName string `yaml:"username" yaml-aliases:"userName,user-name"`
//     TTL time.Duration `yaml:"ttl"`
```

internal/yamlalias/wrap.go
```diff
@@ -0,0 +1,117 @@
...
+//   }
+//
+// It will create a new struct with fields for each alias:
+//
+//   struct {
+//     UserName string `yaml:"username" yaml-aliases:"userName,user-name"`
+//     UserNameYamlAlias1 *string `yaml:"userName"`
+//     UserNameYamlAlias2 *string `yaml:"user-name"`
```
abhinav:
```suggestion
//     UserNameYamlAlias2 *string `yaml:"user-name"`
//     TTL *time.Duration `yaml:"ttl"`
```

internal/yamlalias/wrap.go
```diff
@@ -0,0 +1,117 @@
...
+// from the passed in struct, but adds additional fields for any
+// aliases.
+//
+// Given a struct such as:
+//
+//   type Config struct {
+//     UserName string `yaml:"username" yaml-aliases:"userName,user-name"`
+//   }
+//
+// It will create a new struct with fields for each alias:
```
abhinav:
```suggestion
// It will create a new struct with a pointer field for each field in the
// original struct and a new field for each alias.
```

internal/yamlalias/wrap.go
```diff
@@ -0,0 +1,117 @@
...
+//     UserName string `yaml:"username" yaml-aliases:"userName,user-name"`
+//     UserNameYamlAlias1 *string `yaml:"userName"`
+//     UserNameYamlAlias2 *string `yaml:"user-name"`
+//   }
+//
+// A value of that struct is returned, with the new pointer fields pointing
+// to the original field that defined the aliases:
+//
+// s.UserNameYamlAlias1 = &UserName
+// s.UserNameYamlAlias2 = &UserName
```
abhinav:
```suggestion
// A value of that struct is returned with the pointer fields pointing to
// fields of the original struct.
//
// out.UserName = &dest.UserName
// out.UserNameYamlAlias1 = &dest.UserName
// out.UserNameYamlAlias2 = &dest.UserName
// out.TTL = &dest.TTL
```

internal/yamlalias/wrap.go
```diff
@@ -0,0 +1,117 @@
...
+			dests = append(dests, rv.Field(i).Addr())
+		}
+	}
+
+	// The struct we generated has pointers to the original fields.
+	// Set all the pointers to the fields in the original struct.
+	generated := reflect.StructOf(fields)
+	generatedV := reflect.New(generated)
+	for i := range fields {
+		generatedV.Elem().Field(i).Set(dests[i])
```
abhinav: nit: `fields` value isn't used and `dests` has the same length so:

```suggestion
	for i, dest := range dests {
		generatedV.Elem().Field(i).Set(dest)
```

internal/yamlalias/wrap_test.go
```diff
@@ -0,0 +1,149 @@
...
+		{
+			msg:  "pointer field set directly",
+			data: `{"foo": "foo"}`,
+			want: &AliasToPtr{stringPtr("foo")},
+		},
+		{
+			msg:  "pointer field that is set by alias",
+			data: `{"bar": "bar"}`,
+			want: &AliasToPtr{stringPtr("bar")},
+		},
```
abhinav: What happens if the pointer field is already pointing to another value?
(This is probably its own test outside this table.)

```
var dest string
out := AliasToPtr{Foo: &dest}
yamlalias.Unmarshal([]byte(`{"bar": "bar"}`), &out)
```

prashantv: Added a new test `TestWrapPointerField`, `dest` is updated through the alias as well.

internal/yamlalias/wrap.go
```diff
@@ -0,0 +1,123 @@
...
+//   type Config struct {
+//     UserName string `yaml:"username" yaml-aliases:"userName,user-name"`
+//     TTL time.Duration `yaml:"ttl"`
+//   }
+//
+// It will create a new struct with a pointer field for each field in the
+// original struct and a new field for each alias.
+//
+//   struct {
+//     UserName string `yaml:"username" yaml-aliases:"userName,user-name"`
```
abhinav:
```suggestion
//     UserName *string `yaml:"username" yaml-aliases:"userName,user-name"`
```

PR #269, opened 2019-07-09
Add error tracing to release script

prashantv:
The release script failed again:
https://travis-ci.com/yarpc/yab/jobs/214229866

It's not clear why it failed, since the error doesn't have much
information:
"Script failed with status 1"

I tried to reproduce locally but the release succeeded. Separate the
commands and use error tracing to better understand the error. Error
tracing from:
https://unix.stackexchange.com/questions/462156/how-do-i-find-the-line-number-in-bash-when-an-error-occured

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -euo pipefail
+set -euo pipefail -o errtrace -o functrace
 IFS=$'\n\t'
 
 die() {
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -34,18 +43,22 @@ echo "CHANGELOG:"
...
         body: $changelog,
-      }' | \
-    curl --user "$GITHUB_USER:$GITHUB_TOKEN" -X POST --data @- \
-      "https://api.github.com/repos/$GITHUB_REPO/releases" | \
+      }')
+
+RELEASE_URL="https://api.github.com/repos/$GITHUB_REPO/releases"
+RELEASE_OUT=$(echo "$RELEASE_ARGS" | curl --user "$GITHUB_USER:$GITHUB_TOKEN" -X POST --data @- "$RELEASE_URL")
+
+echo "Release to $RELEASE_URL got:\n$RELEASE_OUT"
```
abhinav: For multi-line output, use separate echoes if possible.

```suggestion
echo "Release to $RELEASE_URL got:"
echo "  $RELEASE_OUT"
```

It's slightly more readable and you don't have to deal with issues like,

```
$ bash
$ echo "foo\nbar"
foo\nbar
$ echo -e "foo\nbar"
foo
bar
```

PR #271, opened 2019-09-05
Fix proto serialization from yaml templates

robbertvanginkel:
Previously we were interpreting the input to `Request(body []byte)` as
json, whereas it should be interpreted as Yaml. As proto  reflection only works on json, convert the yaml to json first and attempt to construct the proto payload from that.

### DIFF
--- a/encoding/protobuf.go
+++ b/encoding/protobuf.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/yarpc/yab/protobuf"
 	"github.com/yarpc/yab/transport"
+	"github.com/yarpc/yab/unmarshal"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/jhump/protoreflect/desc"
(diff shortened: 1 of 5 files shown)


encoding/protobuf.go
```diff
@@ -54,8 +55,18 @@ func (p protoSerializer) Encoding() Encoding {
 }
 
 func (p protoSerializer) Request(body []byte) (*transport.Request, error) {
+	m, err := unmarshal.YAML(body)
+	if err != nil {
+		return nil, err
```
prashantv: let's add test coverage

encoding/protobuf.go
```diff
@@ -54,8 +55,18 @@ func (p protoSerializer) Encoding() Encoding {
...
+	}
+
+	json, err := json.Marshal(m)
+	if err != nil {
+		return nil, err
```
prashantv: should be able to pass in yaml data that cannot be converted to JSON (e.g., map with integers).

robbertvanginkel:
Agree that it isn't great, but better than what we have now. Let's look into supporting a custom serializer and the yaml template functions for proto in the future.

PR #274, opened 2019-09-06
Support --health flag for grpc/proto

### LINKED ISSUE #250: gRPC: simplify service health check with `--health`
Currently for gRPC health checking, we have to specify the health check endpoint package and method, based on the corresponding [protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md), that's not ideal because now every user needs to be aware of the protocol. 

```
yab <service> grpc.health.v1.Health/Check -r '{"service": <service>}' -p grpc://<host:port>
```
It would be great to simply that. For example, we already use `--health` to health check tchannel Meta:health endpoint, it would be nice if we can do the same for gRPC services:
```
yab <service> --health -p grpc://<host:port>
```

robbertvanginkel:
Fixes #250.

### DIFF
--- a/encoding/encoding.go
+++ b/encoding/encoding.go
@@ -60,9 +60,8 @@ const (
 )
 
 var (
-	errNilEncoding = errors.New("cannot Unmarshal into nil Encoding")
-	// ErrHealthThriftOnly is returned if the user specifies an unsupported encoding with --health.
-	ErrHealthThriftOnly = errors.New("--health can only be used with Thrift")
+	errNilEncoding                 = errors.New("cannot Unmarshal into nil Encoding")
+	errUnspecifiedHealthSerializer = errors.New("unspecifiedHealthSerializer should not be called directly")
 )
 
 func (e Encoding) String() string {
@@ -90,13 +89,17 @@ func (e *Encoding) UnmarshalFlag(s string) error {
 }
 
 // GetHealth returns a serializer for the Health endpoint.
-func (e Encoding) GetHealth() (Serializer, error) {
+func (e Encoding) GetHealth(serviceName string) (Serializer, error) {
 	switch e {
-	case UnspecifiedEncoding, Thrift:
+	case UnspecifiedEncoding:
+		return unspecifiedHealthSerializer{serviceName: serviceName}, nil
+	case Thrift:
 		method, spec := getHealthSpec()
 		return thriftSerializer{method, spec, defaultOpts}, nil
+	case Protobuf:
+		return protoHealthSerializer{serviceName: serviceName}, nil
 	default:
-		return nil, ErrHealthThriftOnly
+		return nil, fmt.Errorf("--health not supported with encoding %q", e.String())
 	}
 }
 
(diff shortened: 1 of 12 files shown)


encoding/encoding.go
```diff
@@ -90,13 +88,15 @@ func (e *Encoding) UnmarshalFlag(s string) error {
 }
 
 // GetHealth returns a serializer for the Health endpoint.
-func (e Encoding) GetHealth() (Serializer, error) {
+func (e Encoding) GetHealth(servicename string) (Serializer, error) {
```
prashantv: nit: camelCase, use `serviceName`

encoding/protobuf_health.go
```diff
@@ -0,0 +1,58 @@
+package encoding
+
+import (
+	"encoding/json"
+	"errors"
+
+	"github.com/golang/protobuf/jsonpb"
```
prashantv: nit: import groups need some cleanup here

encoding/protobuf_health.go
```diff
@@ -0,0 +1,58 @@
...
+	str, err := m.MarshalToString(&msg)
+	if err != nil {
+		return nil, err
+	}
+	var unmarshaledJSON json.RawMessage
+	if err = json.Unmarshal([]byte(str), &unmarshaledJSON); err != nil {
+		return nil, err
+	}
+	return unmarshaledJSON, nil
+}
```
prashantv: i thought proto types had `MarshalJSON` natively, should we just return `msg`?

encoding/protobuf_health_test.go
```diff
@@ -0,0 +1,101 @@
...
+func TestProtobufHealth(t *testing.T) {
+	s := &protoHealthSerializer{}
+	assert.Equal(t, Protobuf, s.Encoding())
+}
+
+func TestProtobufHealthRequest(t *testing.T) {
+	tests := []struct {
+		desc        string
+		serviceName string
+		bsIn        []byte
```
prashantv: call this `reqBody`

request.go
```diff
@@ -92,7 +92,21 @@ func NewSerializer(opts Options) (encoding.Serializer, error) {
 			return nil, errHealthAndProcedure
 		}
 
-		return opts.ROpts.Encoding.GetHealth()
+		e := opts.ROpts.Encoding
+		if e == encoding.UnspecifiedEncoding {
+			topts, err := loadTransportPeers(opts.TOpts)
```
prashantv: can we try to avoid duplicating this logic. instead, we might want to handle the health request in `withTransportSerializer` or something, and we can just pass in the transport to `GetHealth`?

encoding/protobuf_health.go
robbertvanginkel: If we were using gogo generated types it would be, but golang generated types do not and need to use the `jsonpb.Marshaller`. Not ideal, but I'd prefer this to taking on a gogo dependency.

request.go
robbertvanginkel: Currently `GetHealth` returns a Serializer, and a serializer is required to create a transport, so that'd be circular. I'll try to find a refactor to avoid having to potentially call `loadTransportPeers` twice and dedupe this.

encoding/encoding.go
```diff
@@ -169,3 +172,31 @@ func (e rawSerializer) Response(res *transport.Response) (interface{}, error) {
...
+func (unspecifiedHealthSerializer) CheckSuccess(res *transport.Response) error {
+	return errUnspecifiedHealthSerializer
+}
+
+func (unspecifiedHealthSerializer) ThriftHealthSerializer() Serializer {
+	method, spec := getHealthSpec()
+	return &thriftSerializer{method, spec, defaultOpts}
+}
+
+func (unspecifiedHealthSerializer) ProtoHealthSerializer(serviceName string) Serializer {
```
prashantv: rather than pass the `serviceName` again here, can we store it in a field for `unspecifiedHealthSerializer` and use that instead

encoding/protobuf_health.go
prashantv: ahh, makes sense, thanks.

main.go
```diff
@@ -396,12 +396,26 @@ func getTracer(opts Options, out output) (opentracing.Tracer, io.Closer) {
...
+type healthSerializer interface {
+	ThriftHealthSerializer() encoding.Serializer
+	ProtoHealthSerializer(serviceName string) encoding.Serializer
+}
+
 // withTransportSerializer may modify the serializer for the transport used.
 // E.g. Thrift payloads are not enveloped when used with TChannel or gRPC.
-func withTransportSerializer(p transport.Protocol, s encoding.Serializer, rOpts RequestOptions) encoding.Serializer {
+func withTransportSerializer(p transport.Protocol, s encoding.Serializer, opts Options) encoding.Serializer {
+	if opts.ROpts.Health && s.Encoding() == encoding.UnspecifiedEncoding {
```
prashantv: rather than checking for health / encoding, can we just check for the interface?

PR #279, opened 2019-10-01
main_test prefactor: Use t.Run and add msg

prashantv:
Some cleanup of main_test to make it easier to debug tests, and make
the tests more correct by only missing the one issue.

This avoids the tests failing when the load order is reordered.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -77,13 +77,22 @@ func TestRunWithOptions(t *testing.T) {
 		wants   []string
 	}{
 		{
-			desc:   "No thrift file, fail to get method spec",
+			desc: "No thrift file, fail to get method spec",
+			opts: Options{
+				TOpts: TransportOptions{
+					ServiceName: "foo",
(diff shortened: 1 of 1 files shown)


main_test.go
```diff
@@ -77,23 +77,33 @@ func TestRunWithOptions(t *testing.T) {
 		wants   []string
 	}{
 		{
-			desc:   "No thrift file, fail to get method spec",
+			desc: "No thrift file, fail to get method spec",
+			opts: Options{
+				TOpts: TransportOptions{
+					ServiceName: "foo",
+					Peers:       []string{"1.1.1.1"},
```
robbertvanginkel: Should we make these `127.0.0.1`, so if for some reason we end up in a codepath that resolves this we don't block on trying to access the internet?

prashantv: :+1: 

updated this test. there are a lot of other references, might do a follow up for those.

PR #282, opened 2019-10-01
Refactor transport & encoding detection

### LINKED ISSUE #281: RPC-Encoding is not set for `--health`
When hitting YARPC services, `--health` fails as it requires `RPC-Encoding` which is not set.

```
Failed while making call: tchannel error ErrCodeBadRequest: failed to decode "thrift" request body for procedure "Meta::health" of service "auto-tconfig" from caller "yab-prashant": expected encoding "thrift" but got "raw"
```

prashantv:
The code to detect encoding/protocol is split up into phases, since we
have to create/detect one before the other.

This has caused issues (e.g., unspecifiedHealthSeriailizer) which then
led to issues like https://github.com/yarpc/yab/issues/281

Rather than fix these as one-offs, refactor the code so we try to
determine the transport protocol & encoding together as one step before
we create the transport or the serializer.

This should make it easier to reason about which encoding/transport
we'll use when the user doesn't specify one.

Fixes #281.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -438,6 +438,7 @@ func TestGRPCReflectionSource(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.desc, func(t *testing.T) {
 			gotOut, gotErr := runTestWithOpts(tt.opts)
+			fmt.Println("got", gotErr)
 			assert.Contains(t, gotErr, tt.wantErr)
 			assert.Contains(t, gotOut, tt.wantRes)
 		})
--- a/integration_test.go
+++ b/integration_test.go
@@ -438,7 +438,6 @@ func TestGRPCReflectionSource(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.desc, func(t *testing.T) {
 			gotOut, gotErr := runTestWithOpts(tt.opts)
-			fmt.Println("got", gotErr)
 			assert.Contains(t, gotErr, tt.wantErr)
 			assert.Contains(t, gotOut, tt.wantRes)
 		})
--- a/request.go
+++ b/request.go
@@ -86,24 +86,23 @@ func getHeaders(inline, file string, override map[string]string) (map[string]str
 }
 
 // NewSerializer creates a Serializer for the specific encoding.
-func NewSerializer(opts Options) (encoding.Serializer, error) {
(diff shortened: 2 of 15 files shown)


integration_test.go
```diff
@@ -438,6 +438,7 @@ func TestGRPCReflectionSource(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.desc, func(t *testing.T) {
 			gotOut, gotErr := runTestWithOpts(tt.opts)
+			fmt.Println("got", gotErr)
```
robbertvanginkel: Clean up debug code.

request.go
```diff
@@ -128,14 +127,11 @@ func NewSerializer(opts Options) (encoding.Serializer, error) {
...
 
-func newProtoDescriptorProvider(ropts RequestOptions, topts TransportOptions) (protobuf.DescriptorProvider, error) {
+func newProtoDescriptorProvider(ropts RequestOptions, topts TransportOptions, resolved resolvedProtocolEncoding) (protobuf.DescriptorProvider, error) {
 	if len(ropts.FileDescriptorSet) > 0 {
 		return protobuf.NewDescriptorProviderFileDescriptorSetBins(ropts.FileDescriptorSet...)
 	}
-	topts, err := loadTransportPeers(topts)
-	if err != nil {
-		return nil, err
-	}
```
robbertvanginkel: Nice :)

transport.go
```diff
@@ -171,19 +156,19 @@ func getTransport(opts TransportOptions, enc encoding.Encoding, tracer opentraci
...
 	}
 
-	if protocol == "grpc" {
+	if resolved.protocol == transport.GRPC {
 		return transport.NewGRPC(transport.GRPCOptions{
 			Addresses:       getHosts(opts.Peers),
 			Tracer:          tracer,
 			Caller:          opts.CallerName,
-			Encoding:        enc.String(),
+			Encoding:        resolved.enc.String(),
```
robbertvanginkel: Note to self we can bring back test+error that were deleted in https://github.com/yarpc/yab/pull/274/files#diff-7f53f14597270e73b9fb5e7892df8ce3.

integration_test.go
prashantv: :+1:

transport.go
prashantv: sweet!

PR #285, opened 2019-10-02
Prepare for 0.16.1 release

robbertvanginkel:


CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+# 0.16.1 (2019-10-02)
+* Fix bug that prevented nested json to be marshalled into proto encoded
+  requests.
+* Fix bug that required users to explictly specify encoding using `-e` when
+  making health requests. Now health requests default to proto when using grpc
+  and thrift when using tchannel.
```
prashantv: more capitalization nits, `gRPC`, `TChannel` . optionally `Thrift` and `Proto` (we always capitalized thrift, but not proto in the release below).

CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
...
+* Updated dependencies.
```
prashantv: since this changelog is for users, i don't think this should impact them, so i'd skip it

PR #286, opened 2019-10-02
encoding: Move out not-found error

prashantv:
This is a prefactoring to improving the error messages for gRPC
when the service cannot be found.

The NotFoundError will allow for more consistency in error messages across packages.


encoding/encodingerror/not_found.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+	"sort"
+	"strings"
+)
+
+// NotFound error is used when a symbol isn't found (or isn't specified)
+// and there is a list of available symbols.
+type NotFound struct {
+	Encoding  string
+	Typ       string
```
robbertvanginkel: Nit: `Typ`'s meaning wasn't very clear till I read the examples.

prashantv: Yeah I didn't like that name, renamed to `SearchType`.

PR #288, opened 2019-10-08
encoding: Use fully qualified method names for Thrift

prashantv:
When a symbol is not found, list the full `Service::Method` in the list so it matches gRPC, and simplifies copy+paste into the procedure field.

### DIFF
--- a/encoding/thrift_request.go
+++ b/encoding/thrift_request.go
@@ -137,20 +137,20 @@ func (e thriftSerializer) WithoutEnvelopes() Serializer {
 }
 
 func findMethod(service *compile.ServiceSpec, methodName string) (*compile.FunctionSpec, error) {
-	functions := service.Functions
-
-	if service.Parent != nil {
-		// Generate a list of functions that includes the inherited functions.
-		functions = make(map[string]*compile.FunctionSpec)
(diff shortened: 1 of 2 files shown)


encoding/thrift_request.go
```diff
@@ -137,20 +137,20 @@ func (e thriftSerializer) WithoutEnvelopes() Serializer {
...
-	if method, found := functions[methodName]; found {
-		return method, nil
+	// If we can't find the service, let's build a list of fully qualified methods.
+	var available []string
+	for cur := service; cur != nil; cur = cur.Parent {
+		for fName := range cur.Functions {
+			fullyQualified := fmt.Sprintf("%v::%v", service.Name, fName)
+			available = append(available, fullyQualified)
+		}
 	}
```
abhinav: Should we sort `available` since `cur.Functions` is a map?

```suggestion
	}
	sort.Strings(available)
```

prashantv: The sorting is handled in the `NotFound` error now,
https://github.com/yarpc/yab/pull/286/files#diff-dd699dedc37aa8fa93f3c7436c9cf9f6R75

PR #289, opened 2019-10-08
protobuf: Better errors when service is not specified/found

prashantv:
Use encodingerror.NotFound to match the Thrift experience when a service
is not found.

Instead of "Symbol not found: Bar", the error will now look like:

```
could not find gRPC service "Bar". Available gRPC service:
    grpc.reflection.v1alpha.ServerReflection
```
Since this will list the available options, it should improve the
user experience significantly, especially since the proto package
names add another attribute to remember for gRPC.


protobuf/filedescriptorsets.go
```diff
@@ -71,13 +72,26 @@ type fileSource struct {
 	files map[string]*desc.FileDescriptor
 }
 
-func (fs *fileSource) FindSymbol(fullyQualifiedName string) (desc.Descriptor, error) {
+func (fs *fileSource) FindService(fullyQualifiedName string) (*desc.ServiceDescriptor, error) {
+	// available := map[string][]string
```
robbertvanginkel: Trailing comment?

abhinav:
😄 

image

prashantv:
Glad you found it before release, will remove this closed commit from `dev` and open a new PR.

PR #290, opened 2019-10-14
protobuf: Better errors when service is not specified/found

prashantv:
Use encodingerror.NotFound to match the Thrift experience when a service
is not found.

Instead of "Symbol not found: Bar", the error will now look like:

```
could not find gRPC service "Bar". Available gRPC service:
    grpc.reflection.v1alpha.ServerReflection
```
Since this will list the available options, it should improve the
user experience significantly, especially since the proto package
names add another attribute to remember for gRPC.

This is the same as #289 without the binary file.


protobuf/source.go
```diff
@@ -4,10 +4,9 @@ import "github.com/jhump/protoreflect/desc"
 
 // DescriptorProvider is a source of protobuf descriptor information. It can be backed by a FileDescriptorSet
 // proto (like a file generated by protoc).
-// Adapted from https://github.com/fullstorydev/grpcurl
 type DescriptorProvider interface {
-	// FindSymbol returns a descriptor for the given fully-qualified symbol name.
-	FindSymbol(fullyQualifiedName string) (desc.Descriptor, error)
+	// FindService returns a servie descriptor for the given fully-qualified symbol name.
```
abhinav:
```suggestion
	// FindService returns a service descriptor for the given fully-qualified symbol name.
```

PR #293, opened 2020-03-19
gRPC: Propogate routing delegate/key to reflection calls

### LINKED ISSUE #292: gRPC reflection calls ignore routing key/routing delegates
Call here is only setting `rpc-service`:
https://github.com/yarpc/yab/blob/dev/protobuf/source_reflection.go#L53

prashantv:
The gRPC reflection logic creates a custom gRPC connection (using native
gRPC) compared to the call (which uses YARPC), with inconsistent
handling of routing headers.

The reflection code was not using the routing delegate/key previously,
so fix the logic to propogate these headers.

Fixes #292

### DIFF
--- a/protobuf/source_reflection.go
+++ b/protobuf/source_reflection.go
@@ -9,6 +9,7 @@ import (
 	"github.com/jhump/protoreflect/desc"
 	"github.com/jhump/protoreflect/grpcreflect"
 	"github.com/yarpc/yab/encoding/encodingerror"
+	yproto "go.uber.org/yarpc/encoding/protobuf"
 	ygrpc "go.uber.org/yarpc/transport/grpc"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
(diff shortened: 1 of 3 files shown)


protobuf/source_reflection_test.go
```diff
@@ -99,3 +102,97 @@ func TestReflectionNotRegistered(t *testing.T) {
...
+	defer ln.Close()
+
+	s := grpc.NewServer()
+	go s.Serve(ln)
+
+	// Ensure that all streams are closed by the end of the test.
+	defer s.GracefulStop()
+
+	drs := &dummyReflectionServer{}
+	rpb.RegisterServerReflectionServer(s, drs)
```
AlexAPB: We should register the server before calling. `Serve` :
- https://godoc.org/google.golang.org/grpc#Server.RegisterService

This register call just calls `s.RegisterService`, ie
```
func RegisterServerReflectionServer(s *grpc.Server, srv ...) {
	s.RegisterService(&_ServerReflection_serviceDesc, srv)
}
```

>  This must be called before invoking Serve.

protobuf/source_reflection.go
```diff
@@ -47,12 +50,20 @@ func NewDescriptorProviderReflection(args ReflectionArgs) (DescriptorProvider, e
...
-		map[string][]string{
-			"rpc-caller":   []string{args.Caller},
-			"rpc-service":  []string{args.Service},
-			"rpc-encoding": []string{"proto"},
-		})
+
+	routingHeaders := metadata.Pairs(
+		ygrpc.CallerHeader, args.Caller,
+		ygrpc.ServiceHeader, args.Service,
+		ygrpc.EncodingHeader, "proto",
```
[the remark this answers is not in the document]
prashantv: Good idea, done.

protobuf/source_reflection_test.go
prashantv: Meant to register this before `Serve`, thanks. Fixed.

PR #297, opened 2020-03-24
Use Go modules and GitHub Workflows

abhinav:
Since our Travis-based release process is broken, this switches yab to
using GitHub workflows for CI and release.

In the process, this switches to Go modules because GH workflows
doesn't work nicely with Glide.


.github/workflows/go.yml
```diff
@@ -0,0 +1,75 @@
...
+        go-version: 1.14.x
+
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Load cached dependencies
+      uses: actions/cache@v1
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
```
prashantv: no action needed, just a fyi for future:
this is from:
https://github.com/actions/cache/blob/master/examples.md#go---modules

.github/workflows/go.yml
```diff
@@ -0,0 +1,75 @@
...
+        restore-keys: |
+          ${{ runner.os }}-go-
```
prashantv: is there a way to reference the same steps + params as build rather than copy+paste?

.github/workflows/go.yml
```diff
@@ -0,0 +1,75 @@
...
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Setup Go
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.14.x
```
prashantv: (no action for this change) if we wanted to test multiple versions (not important for yab, since it's not a library), can we easily do that?

.github/workflows/go.yml
abhinav: Not that i know of. I mean, if we wanted to use YAML pointers, maybe? I wondered if it was possible to share the setup between the two but there was no obvious way.

There's some discussion in https://github.community/t5/GitHub-Actions/reusing-sharing-inheriting-steps-between-jobs-declarations/td-p/37849 but I don't see resolutions incoming.

.github/workflows/go.yml
abhinav: Yeah, there's support for multiple versions. i have that in one of my repos.

PR #300, opened 2020-03-27
encoding: Determine whether to envelope before creating serializer

### LINKED ISSUE #214: yab panic when using raw encoding and `--disable-thrift-envelope`
```
yab -e raw -p http://127.0.0.1:31337/ anything --procedure here --disable-thrift-envelope
```
```
panic: interface conversion: encoding.rawSerializer is not main.noEnveloper: missing method WithoutEnvelopes

goroutine 1 [running]:
main.withTransportSerializer(0x2, 0x1130740, 0xc42057b9b0, 0xc42057b8ac, 0x3, 0x0, 0x0, 0x7ffcb32a48fb, 0x4, 0xc4200f8020, ...)
	src/github.com/yarpc/yab/main.go:370 +0x5c
main.runWithOptions(0xc42057b8ac, 0x3, 0x0, 0x0, 0x7ffcb32a48fb, 0x4, 0xc4200f8020, 0x0, 0x0, 0x0, ...)
	src/github.com/yarpc/yab/main.go:320 +0x808
main.parseAndRun(0x1130a40, 0xc42000e018)
	src/github.com/yarpc/yab/main.go:206 +0x4ba
main.main()
	src/github.com/yarpc/yab/main.go:83 +0x46
```

prashantv:
After refactoring transport/encoding detection (#282), we can now
determine whether to envelope before creating the serializer, rather
than creating the serializer, then upcasting to disable envelopes.

Fixes #214 (no longer have interface upcasting issues)


encoding/meta.go
```diff
@@ -92,5 +92,6 @@ func getMetaService() *compile.ServiceSpec {
 }
 
 func getHealthSpec() (string, *compile.FunctionSpec) {
-	return metaService + "::" + healthMethod, getMetaService().Functions[healthMethod]
+	healthSpec := getMetaService().Functions[healthMethod]
+	return metaService + "::" + healthMethod, healthSpec
```
robbertvanginkel: There doesn't seem to be a functional change here, any reason these changes are included?

prashantv: An intermediate state had more returns here, but I actually simplifies it back, reverting, thanks.

PR #301, opened 2020-06-09
Add JSON Format Option to Benchmark Output

nishakhater:
These changes allow a user to specify a 'format' option when running yab. Currently, the two formats supported are 'json' and 'plaintext' with 'plaintext' being the default output that is currently outputted by yab. When a user specifies the 'json' output, fields such as benchmark parameters, latencies, and benchmarking summaries will be printed in JSON format. This output format has been tested in benchmark_test.go.

Example JSON Output
```
{
  "body": {
    "result": {
      "count": 2,
      "message": "hello"
    }
  },
  "headers": {
    "Content-Length": "38",
    "Content-Type": "application/vnd.apache.thrift.binary",
    "Date": "Thu, 18 Jun 2020 23:32:07 GMT",
    "Rpc-Service": "hello",
    "Rpc-Status": "success"
  },
  "statusCode": 200
}

{
  "benchmarkParameters": {
    "cpus": 8,
    "connections": 16,
    "concurrency": 1,
    "maxRequests": 0,
    "maxDuration": "1s",
    "maxRPS": 0
  },
  "latencies": {
    "0.5000": "224.56µs",
    "0.9000": "541.276µs",
    "0.9500": "780.953µs",
    "0.9900": "1.409557ms",
    "0.9990": "4.644993ms",
    "0.9995": "5.223061ms",
    "1.0000": "7.468056ms"
  },
  "summary": {
    "elapsedTimeSeconds": 1,
    "totalRequests": 51891,
    "rps": 51851.72
  }
}
```

### DIFF
--- a/bench_state.go
+++ b/bench_state.go
@@ -94,15 +94,6 @@ func (s *benchmarkState) getLatencies(out output, quantiles []string) map[string
 	return latencyValues
 }
 
-// Prints out latency values
-func (s *benchmarkState) printLatencies(out output) {
-	sort.Sort(byDuration(s.latencies))
-	out.Printf("Latencies:\n")
-	for _, quantile := range []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0} {
-		out.Printf("  %.4f: %v\n", quantile, s.getQuantile(quantile))
-	}
-}
-
 func (s *benchmarkState) printErrors(out output) {
 	if len(s.errors) == 0 {
 		return
--- a/bench_state.go
+++ b/bench_state.go
@@ -80,11 +80,11 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 }
 
 // Returns an array of latency values
-func (s *benchmarkState) getLatencies(out output) []string {
+func (s *benchmarkState) getLatencies(out output, quantiles []float64) map[string]string {
 	sort.Sort(byDuration(s.latencies))
-	latencyValues := make([]string, 7)
-	for i, quantile := range []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0} {
-		latencyValues[i] = s.getQuantile(quantile).String()
+	latencyValues := make(map[string]string)
+	for _, quantile := range quantiles {
+		latencyValues[fmt.Sprintf("%.4f", quantile)] = s.getQuantile(quantile).String()
 	}
 	return latencyValues
 }
--- a/bench_state.go
+++ b/bench_state.go
@@ -80,11 +80,11 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 	s.statter.Timing("latency", d)
 }
 
-// Returns an array of latency values
+// Returns a mapping of quantiles to latency values
 func (s *benchmarkState) getLatencies(out output) map[string]string {
 	sort.Sort(byDuration(s.latencies))
-	latencyValues := make(map[string]string, len(quantiles))
-	for _, quantile := range quantiles {
+	latencyValues := make(map[string]string, len(_quantiles))
+	for _, quantile := range _quantiles {
 		floatQuantile, err := strconv.ParseFloat(quantile, 64)
 		if err != nil {
 			continue
--- a/bench_state.go
+++ b/bench_state.go
@@ -81,7 +81,7 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 }
 
 // Returns an array of latency values
-func (s *benchmarkState) getLatencies(out output, quantiles []string) map[string]string {
+func (s *benchmarkState) getLatencies(out output) map[string]string {
 	sort.Sort(byDuration(s.latencies))
 	latencyValues := make(map[string]string, len(quantiles))
 	for _, quantile := range quantiles {
--- a/bench_state.go
+++ b/bench_state.go
@@ -85,7 +85,10 @@ func (s *benchmarkState) getLatencies(out output, quantiles []string) map[string
 	sort.Sort(byDuration(s.latencies))
 	latencyValues := make(map[string]string, len(quantiles))
 	for _, quantile := range quantiles {
-		floatQuantile, _ := strconv.ParseFloat(quantile, 64)
+		floatQuantile, err := strconv.ParseFloat(quantile, 64)
+		if err != nil {
+			continue
+		}
 		latencyValues[quantile] = s.getQuantile(floatQuantile).String()
 	}
 	return latencyValues
--- a/bench_state.go
+++ b/bench_state.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"sort"
 	"time"
+    "encoding/json"
 
 	"github.com/yarpc/yab/sorted"
 	"github.com/yarpc/yab/statsd"
@@ -82,10 +83,24 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 func (s *benchmarkState) printLatencies(out output) {
 	// TODO JSON output?
 	sort.Sort(byDuration(s.latencies))
-	out.Printf("Latencies:\n")
-	for _, quantile := range []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0} {
-		out.Printf("  %.4f: %v\n", quantile, s.getQuantile(quantile))
+    quantiles := []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0}
+    latency_values := make(map[string]string)
+    for _, quantile := range quantiles {
+        q := fmt.Sprintf("%f", quantile)
+        latency_values[q] = s.getQuantile(quantile).String()
+    }
+    latencies := make(map[string]interface{})
+    latencies["Latencies"] = latency_values
+    output, err := json.MarshalIndent(latencies, "", "  ")
+    if err != nil {
+		out.Fatalf("Failed to convert map to JSON")
 	}
+    out.Printf("%s\n\n", output)
+
+	// out.Printf("Latencies:\n")
+	// for _, quantile := range []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0} {
+	// 	out.Printf("  %.4f: %v\n", quantile, s.getQuantile(quantile))
+	// }
 }
 
 func (s *benchmarkState) printErrors(out output) {
--- a/bench_state.go
+++ b/bench_state.go
@@ -82,12 +82,11 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 // Returns an array of latency values
 func (s *benchmarkState) getLatencies(out output) []string {
 	sort.Sort(byDuration(s.latencies))
-    quantiles := []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0}
-    latencyValues := make([]string, 7)
-    for i, quantile := range quantiles {
-        latencyValues[i] = s.getQuantile(quantile).String()
-    }
-    return latencyValues
+	latencyValues := make([]string, 7)
+	for i, quantile := range []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0} {
+		latencyValues[i] = s.getQuantile(quantile).String()
+	}
+	return latencyValues
 }
 
 // Prints out latency values
--- a/bench_state.go
+++ b/bench_state.go
@@ -79,7 +79,7 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 	s.statter.Timing("latency", d)
 }
 
-func (s *benchmarkState) printLatencies(out output) []string {
+func (s *benchmarkState) getLatencies(out output) []string {
 	// TODO JSON output?
 	sort.Sort(byDuration(s.latencies))
     quantiles := []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0}
--- a/bench_state.go
+++ b/bench_state.go
@@ -23,6 +23,7 @@ package main
 import (
 	"fmt"
 	"sort"
+	"strconv"
 	"time"
 
 	"github.com/yarpc/yab/sorted"
@@ -80,11 +81,12 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 }
 
 // Returns an array of latency values
-func (s *benchmarkState) getLatencies(out output, quantiles []float64) map[string]string {
+func (s *benchmarkState) getLatencies(out output, quantiles []string) map[string]string {
 	sort.Sort(byDuration(s.latencies))
-	latencyValues := make(map[string]string)
+	latencyValues := make(map[string]string, len(quantiles))
 	for _, quantile := range quantiles {
-		latencyValues[fmt.Sprintf("%.4f", quantile)] = s.getQuantile(quantile).String()
+		floatQuantile, _ := strconv.ParseFloat(quantile, 64)
+		latencyValues[quantile] = s.getQuantile(floatQuantile).String()
 	}
 	return latencyValues
 }
--- a/bench_state.go
+++ b/bench_state.go
@@ -79,7 +79,7 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 	s.statter.Timing("latency", d)
 }
 
-// Prints an array of latency values
+// Returns an array of latency values
 func (s *benchmarkState) getLatencies(out output) []string {
 	sort.Sort(byDuration(s.latencies))
     quantiles := []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0}
@@ -88,6 +88,15 @@ func (s *benchmarkState) getLatencies(out output) []string {
         latency_values[i] = s.getQuantile(quantile).String()
     }
     return latency_values
+}
+
+// Prints out latency values
+func (s *benchmarkState) printLatencies(out output) {
+	sort.Sort(byDuration(s.latencies))
+	out.Printf("Latencies:\n")
+	for _, quantile := range []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0} {
+		out.Printf("  %.4f: %v\n", quantile, s.getQuantile(quantile))
+	}
 }
 
 func (s *benchmarkState) printErrors(out output) {
--- a/bench_state.go
+++ b/bench_state.go
@@ -23,7 +23,6 @@ package main
 import (
 	"fmt"
 	"sort"
-	"strconv"
 	"time"
 
 	"github.com/yarpc/yab/sorted"
@@ -81,15 +80,11 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 }
 
 // Returns a mapping of quantiles to latency values
-func (s *benchmarkState) getLatencies(out output) map[string]string {
+func (s *benchmarkState) getLatencies(out output) map[float64]time.Duration {
 	sort.Sort(byDuration(s.latencies))
-	latencyValues := make(map[string]string, len(_quantiles))
+	latencyValues := make(map[float64]time.Duration, len(_quantiles))
 	for _, quantile := range _quantiles {
-		floatQuantile, err := strconv.ParseFloat(quantile, 64)
-		if err != nil {
-			continue
-		}
-		latencyValues[quantile] = s.getQuantile(floatQuantile).String()
+		latencyValues[quantile] = s.getQuantile(quantile)
 	}
 	return latencyValues
 }
--- a/bench_state.go
+++ b/bench_state.go
@@ -80,7 +80,7 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 }
 
 // Returns a mapping of quantiles to latency values
-func (s *benchmarkState) getLatencies(out output) map[float64]time.Duration {
+func (s *benchmarkState) getLatencies() map[float64]time.Duration {
 	sort.Sort(byDuration(s.latencies))
 	latencyValues := make(map[float64]time.Duration, len(_quantiles))
 	for _, quantile := range _quantiles {
--- a/bench_state.go
+++ b/bench_state.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"sort"
 	"time"
-    "encoding/json"
 
 	"github.com/yarpc/yab/sorted"
 	"github.com/yarpc/yab/statsd"
@@ -80,27 +79,15 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 	s.statter.Timing("latency", d)
 }
 
-func (s *benchmarkState) printLatencies(out output) {
+func (s *benchmarkState) printLatencies(out output) []string {
 	// TODO JSON output?
 	sort.Sort(byDuration(s.latencies))
     quantiles := []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0}
-    latency_values := make(map[string]string)
-    for _, quantile := range quantiles {
-        q := fmt.Sprintf("%f", quantile)
-        latency_values[q] = s.getQuantile(quantile).String()
+    latency_values := make([]string, 7)
+    for i, quantile := range quantiles {
+        latency_values[i] = s.getQuantile(quantile).String()
     }
-    latencies := make(map[string]interface{})
-    latencies["Latencies"] = latency_values
-    output, err := json.MarshalIndent(latencies, "", "  ")
-    if err != nil {
-		out.Fatalf("Failed to convert map to JSON")
-	}
-    out.Printf("%s\n\n", output)
-
-	// out.Printf("Latencies:\n")
-	// for _, quantile := range []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0} {
-	// 	out.Printf("  %.4f: %v\n", quantile, s.getQuantile(quantile))
-	// }
+    return latency_values
 }
 
 func (s *benchmarkState) printErrors(out output) {
--- a/bench_state.go
+++ b/bench_state.go
@@ -83,11 +83,11 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 func (s *benchmarkState) getLatencies(out output) []string {
 	sort.Sort(byDuration(s.latencies))
     quantiles := []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0}
-    latency_values := make([]string, 7)
+    latencyValues := make([]string, 7)
     for i, quantile := range quantiles {
-        latency_values[i] = s.getQuantile(quantile).String()
+        latencyValues[i] = s.getQuantile(quantile).String()
     }
-    return latency_values
+    return latencyValues
 }
 
 // Prints out latency values
--- a/bench_state.go
+++ b/bench_state.go
@@ -79,8 +79,8 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 	s.statter.Timing("latency", d)
 }
 
+// Prints an array of latency values
 func (s *benchmarkState) getLatencies(out output) []string {
-	// TODO JSON output?
 	sort.Sort(byDuration(s.latencies))
     quantiles := []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0}
     latency_values := make([]string, 7)
--- a/benchmark.go
+++ b/benchmark.go
@@ -59,7 +59,7 @@ type Summary struct {
 	RPS           float64 `json:"rps"`
 }
 
-// BenchmarkOutput stores above defined structs
+// BenchmarkOutput stores benchmark settings and results
 type BenchmarkOutput struct {
 	BenchmarkParameters BenchmarkParameters `json:"benchmarkParameters"`
 	Latencies           map[string]string   `json:"latencies"`
@@ -206,7 +206,7 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
 	overall.printErrors(out)
 
 	// Create Quantiles array
-	quantiles := []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0}
+	quantiles := []string{"0.5000", "0.9000", "0.9500", "0.9900", "0.9990", "0.9995", "1.0000"}
 	latencyValues := overall.getLatencies(out, quantiles)
 
 	// Create BenchmarkParameters, Summary, and BenchmarkOutput structs
@@ -248,7 +248,7 @@ func outputJSON(overall *benchmarkState, out output, total time.Duration, benchm
 }
 
 // Plaintext output helper method
-func outputPlaintext(overall *benchmarkState, out output, total time.Duration, quantiles []float64, benchmarkOutput BenchmarkOutput) {
+func outputPlaintext(overall *benchmarkState, out output, total time.Duration, quantiles []string, benchmarkOutput BenchmarkOutput) {
 	// Print out benchmark parameters
 	out.Printf("Benchmark parameters:\n")
 	out.Printf("  CPUs:            %v\n", benchmarkOutput.BenchmarkParameters.CPUs)
@@ -261,7 +261,7 @@ func outputPlaintext(overall *benchmarkState, out output, total time.Duration, q
 	// Print out latencies
 	out.Printf("Latencies:\n")
 	for _, quantile := range quantiles {
-		out.Printf("  %.4f: %v\n", quantile, benchmarkOutput.Latencies[fmt.Sprintf("%.4f", quantile)])
+		out.Printf("  %s: %v\n", quantile, benchmarkOutput.Latencies[quantile])
 	}
 
 	// Print out summary
--- a/benchmark.go
+++ b/benchmark.go
@@ -25,6 +25,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math"
 	"os"
 	"os/signal"
 	"runtime"
@@ -59,8 +60,8 @@ var (
 // BenchmarkParameters holds values of all benchmark parameters
 type BenchmarkParameters struct {
 	CPUs        int    `json:"cpus"`
-	Concurrency int    `json:"concurrency"`
 	Connections int    `json:"connections"`
+	Concurrency int    `json:"concurrency"`
 	MaxRequests int    `json:"maxRequests"`
 	MaxDuration string `json:"maxDuration"`
 	MaxRPS      int    `json:"maxRPS"`
@@ -301,10 +302,15 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
 
 	latencyValues := overall.getLatencies()
 
+	// Rounding RPS value to the hundredths place
+	rps := float64(overall.totalRequests) / total.Seconds()
+	rps = math.Round(rps * 100)
+	rps = rps / 100
+
 	summary := Summary{
 		ElapsedTime:   (total / time.Millisecond * time.Millisecond).String(),
 		TotalRequests: overall.totalRequests,
-		RPS:           float64(overall.totalRequests) / total.Seconds(),
+		RPS:           rps,
 	}
 
 	if formatAsJSON {
--- a/benchmark.go
+++ b/benchmark.go
@@ -44,7 +44,7 @@ var (
 	// using a global quantiles array mainly for ease of testing, and not passing the same array
 	// around to multiple functions
 	// TODO: modify quantile parameters to remove "." character and increase ease of parsing
-	_quantiles = []string{"0.5000", "0.9000", "0.9500", "0.9900", "0.9990", "0.9995", "1.0000"}
+	_quantiles = []float64{0.5000, 0.9000, 0.9500, 0.9900, 0.9990, 0.9995, 1.0000}
 )
 
 // BenchmarkParameters holds values of all benchmark parameters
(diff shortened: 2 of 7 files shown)


bench_state.go
```diff
@@ -79,8 +79,19 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 	s.statter.Timing("latency", d)
 }
 
+// Returns an array of latency values
+func (s *benchmarkState) getLatencies(out output) []string {
+	sort.Sort(byDuration(s.latencies))
```
jaricftw, line 84: here and throughout the PR, indentation seems off

bench_state.go
```diff
@@ -79,8 +79,19 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
...
+    quantiles := []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0}
+    latencyValues := make([]string, 7)
+    for i, quantile := range quantiles {
```
jaricftw, line 87: you can simplify to 
```
for i, quantile := range []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0} {
}
```

benchmark.go
```diff
@@ -41,6 +42,40 @@ var (
 	errNegativeMaxReqs  = errors.New("max requests cannot be negative")
 )
 
+// Struct to hold values of benchmark parameters
```
jaricftw, line 45: nits: Comments for exposed structs normally starts with the struct name, e.g. 
```
BenchmarkParameters holds values....
```

same for other structs you defined

benchmark.go
```diff
@@ -41,6 +42,40 @@ var (
...
+type BenchmarkParameters struct {
+    CPUs int `json:"CPUs"`
```
jaricftw, line 47: for naming consistency with Go, can you use `camelCase` for the json tag names? e.g. `cpus`, `maxRequests`, etc.

ref: https://stackoverflow.com/questions/5543490/json-naming-convention

benchmark.go
```diff
@@ -188,12 +213,57 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
 		zap.Time("startTime", start),
 	)
 
+    if opts.Format == "json" || opts.Format == "JSON" {
+        // Calling JSON output helper method
```
jaricftw, line 217: comments are not giving extra useful information. Since the logic is also self-descriptive, consider remove the comments

benchmark.go
```diff
@@ -188,12 +213,57 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+        outputJSON(opts, overall, out, goMaxProcs, numConns, total)
+    } else {
+        // Calling plaintext output helper method
+        outputPlaintext(opts, overall, out, goMaxProcs, numConns, total)
+	}
+}
+
+// JSON output helper method
+func outputJSON(opts BenchmarkOptions, overall *benchmarkState, out output, goMaxProcs int, numConns int, total time.Duration) {
```
jaricftw, line 226: generally for readability, consider adding extra blank lines to separate logic inside a function. E.g. take a look a `getQuantile` function on how it does that.

benchmark.go
```diff
@@ -188,12 +213,57 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+	// Print out errors
 	overall.printErrors(out)
-	overall.printLatencies(out)
+	// Create Latencies, BenchmarkParameters, and Summary struct
+    latencyValues := overall.getLatencies(out)
+    benchmarkParams := BenchmarkParameters{CPUs: goMaxProcs, Connections: numConns, Concurrency: opts.Concurrency, MaxRequests: opts.MaxRequests, MaxDuration: opts.MaxDuration.String(), MaxRPS: opts.RPS}
```
jaricftw, line 231: readability: can you have a separate line for each field

benchmark.go
```diff
@@ -188,12 +213,57 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+	latenciesOutput := Latencies{P5000: latencyValues[0], P9000: latencyValues[1], P9500: latencyValues[2], P9900: latencyValues[3], P9990: latencyValues[4], P9995: latencyValues[5], P1000: latencyValues[6]}
+	summary := Summary{ElapsedTime: (total / time.Millisecond * time.Millisecond).String(), TotalRequests: overall.totalRequests, RPS: float64(overall.totalRequests)/total.Seconds()}
+	// Create output struct and format into JSON
+    benchmarkOutput := BenchmarkOutput{BenchmarkParameters: benchmarkParams, Latencies: latenciesOutput, Summary: summary}
+    functionOutput, err := json.MarshalIndent(&benchmarkOutput, "", "  ")
```
jaricftw, line 236: nits: avoid naked arguments, it's recommended to add a comment on what those means, e.g. 
```
json.MarshalIndent(&benchmarkOutput, "" /* prefix */, "  " /* indent */)
```
ref: https://github.com/uber-go/guide/blob/master/style.md#avoid-naked-parameters

benchmark.go
```diff
@@ -188,12 +213,57 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+    if err != nil {
+        out.Fatalf("Failed to convert map to JSON: %v\n", err)
```
jaricftw, line 238: error message isn't quite accurate, consider:
```out.Fatalf("Failed to marshal benchmark output: %v\n", err)

File: benchmark_test.go [hunk 10]
```diff
@@ -218,3 +220,106 @@ func TestBenchmarkStatsPerPeer(t *testing.T) {
 	})
 	assert.Equal(t, want, statsServer.Aggregated(), "unexpected stats")
 }
+
+// Method that tests the 'Format' option of Benchmark Options
+func TestBenchmarkOutput(t *testing.T) {
+	type BenchmarkParameters struct {
```
jaricftw, line 226: you don't need to re-define the structure here, should be able just use the ones you defined in Benchmark.go

benchmark.go
```diff
@@ -41,6 +42,41 @@ var (
...
+	CPUs        int    `json:"cpus"`
+	Concurrency int    `json:"concurrency"`
+	Connections int    `json:"connections"`
+	MaxRequests int    `json:"maxRequests"`
+	MaxDuration string `json:"maxDuration"`
+	MaxRPS      int    `json:"maxRPS"`
+}
+
+// Latencies stores the latencies output for different quantiles
+type Latencies struct {
```
jaricftw, line 56: related to a previous comment, instead of having a  separate struct for storing Latency values, can we simply use a `map[string]string` and use that in `Summary`

benchmark.go
```diff
@@ -188,9 +214,75 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+		P5000: latencyValues[0],
+		P9000: latencyValues[1],
+		P9500: latencyValues[2],
+		P9900: latencyValues[3],
+		P9990: latencyValues[4],
+		P9995: latencyValues[5],
+		P1000: latencyValues[6],
+	}
+	summary := Summary{
+		ElapsedTime:   (total / time.Millisecond * time.Millisecond).String(),
```
jaricftw, line 249: it looks like there are quite some duplicated logic in `outputJSON` and `outputPlainText` (cacluating ElapsedTime/RPS). An idea: we always populate the BenchmarkParams/BenchmarkOutput structs before checking for `Format`, and then passing those structs to `outputJSON` and `outputPlainText` so that they can just use the results as is, without having to do the calculating again.

benchmark.go
```diff
@@ -39,8 +40,35 @@ import (
 var (
 	errNegativeDuration = errors.New("duration cannot be negative")
 	errNegativeMaxReqs  = errors.New("max requests cannot be negative")
+
+	// Global quantiles array for obtaining latency values from getLatencies() method
```
jaricftw, line 44: since this is a global, we can be more explicit on why we are doing this. Looks like this is mainly so we don't create the same array in test, so along the lines of "using a global quantiles array mainly for ease of testing, and not passing the same arpary around to multiple functions".

also, a nit, we usually prefix global with `_` to infer it's package local: https://github.com/uber-go/guide/blob/master/style.md#prefix-unexported-globals-with-_

benchmark_test.go
```diff
@@ -218,3 +220,76 @@ func TestBenchmarkStatsPerPeer(t *testing.T) {
...
+		opts BenchmarkOptions
+	}{
+		{
+			opts: BenchmarkOptions{
+				Format: "json",
+			},
+		},
+		{
+			opts: BenchmarkOptions{
+				Format: "plaintext",
```
jaricftw, line 237: can we add a test case with a different format, to test the behavior that we default to plain text

benchmark_test.go
```diff
@@ -218,3 +220,76 @@ func TestBenchmarkStatsPerPeer(t *testing.T) {
 	})
 	assert.Equal(t, want, statsServer.Aggregated(), "unexpected stats")
 }
+
+// Method that tests the 'Format' option of Benchmark Options
+func TestBenchmarkOutput(t *testing.T) {
+	// Testing both plaintext and JSON output
+	tests := []struct {
+		opts BenchmarkOptions
```
jaricftw, line 228: a couple of things for writing a good table test:
- we can add a field to indicate the test case name, and use `t.Run` so that each test case can be run independently, and easily readable from the test output, an good example: https://github.com/yarpc/yab/blob/dev/template_test.go#L159
- we can add a filed in the test struct to indicate what are the expected output, and can just use that in the test, that way, you can get ride of the if-else check

benchmark.go
```diff
@@ -39,8 +40,36 @@ import (
 var (
 	errNegativeDuration = errors.New("duration cannot be negative")
 	errNegativeMaxReqs  = errors.New("max requests cannot be negative")
+
+	// using a global quantiles array mainly for ease of testing, and not passing the same array
+	// around to multiple functions
+	_quantiles = []string{"0.5000", "0.9000", "0.9500", "0.9900", "0.9990", "0.9995", "1.0000"}
```
jaricftw, line 46: as chatted offline: usually we'd want to avoid a mutable global if possible: https://github.com/uber-go/guide/blob/master/style.md#avoid-mutable-globals. But I see that it'll simplify code much, and avoid duplicating the same array in test to print latencies for validation. so I am not strongly against this.

benchmark_test.go
```diff
@@ -218,3 +219,90 @@ func TestBenchmarkStatsPerPeer(t *testing.T) {
...
+func TestBenchmarkOutput(t *testing.T) {
+	// Testing both plaintext and JSON output
+	tests := []struct {
+		name          string
+		opts          BenchmarkOptions
+		wantOutput    []string
+		notWantOutput []string
+	}{
+		{
+			name: "Test json output",
```
jaricftw, line 232: nit: name can just `json` for simplicity. similarly for other test case names

benchmark_test.go
```diff
@@ -218,3 +219,90 @@ func TestBenchmarkStatsPerPeer(t *testing.T) {
...
+				// Creating struct from string of JSON output
+				var benchmarkOutput BenchmarkOutput
+				b := []byte(bufStr)
+				err := json.Unmarshal(b, &benchmarkOutput)
+				if err != nil {
+					fmt.Println("error:", err)
+				}
+				assert.Equal(t, benchmarkOutput.BenchmarkParameters.MaxRPS, opts.BOpts.RPS)
+				assert.GreaterOrEqual(t, opts.BOpts.MaxRequests, benchmarkOutput.Summary.TotalRequests)
+				return
```
jaricftw, line 303: since there is now no more logic after this, you can drop the `return`

benchmark_test.go
```diff
@@ -218,3 +219,90 @@ func TestBenchmarkStatsPerPeer(t *testing.T) {
...
+				assert.NotContains(t, bufStr, notWant)
+			}
+
+			if opts.BOpts.Format == "json" {
+				// Creating struct from string of JSON output
+				var benchmarkOutput BenchmarkOutput
+				b := []byte(bufStr)
+				err := json.Unmarshal(b, &benchmarkOutput)
+				if err != nil {
+					fmt.Println("error:", err)
```
jaricftw, line 299: we should validate there is no error instead of simply printing it.

nishakhater: Adding return after to make sure no further code is executed if error is thrown.

benchmark.go
```diff
@@ -187,13 +206,77 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
 		zap.Int("totalRequests", overall.totalRequests),
 		zap.Time("startTime", start),
 	)
-
+	// Print out errors
```
jaricftw, line 209: just a note here to think about. we are printing errors in plaintext regardlesss of of the output format. That means we may have plaintext mixed with json output in case of error, and make the final output not parsable. That's not critical though, given this is just the error case. maybe leave a TODO here to consider move error into json field as well.

benchmark.go
jaricftw, line 46: another note, and doesn't need to be addressed in this PR, with `.` in the field, it makes json parsing (e.g. using `jq` a bit troublesome, but still possible, e.g. we'd `jq `.latencies[".0.9900"]`. Maybe leave a TODO so we make it easier to parse (i.e. output p50, p95, p99 instead)

bench_state.go
```diff
@@ -79,13 +80,18 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
 	s.statter.Timing("latency", d)
 }
 
-func (s *benchmarkState) printLatencies(out output) {
-	// TODO JSON output?
+// Returns a mapping of quantiles to latency values
+func (s *benchmarkState) getLatencies(out output) map[string]string {
```
prashantv, line 84: should we use `map[float64]time.Duration` so that the quantile conversion doesn't lose the type/precision?

bench_state.go
```diff
@@ -79,13 +80,18 @@ func (s *benchmarkState) recordLatency(d time.Duration) {
...
 	sort.Sort(byDuration(s.latencies))
-	out.Printf("Latencies:\n")
-	for _, quantile := range []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0} {
-		out.Printf("  %.4f: %v\n", quantile, s.getQuantile(quantile))
+	latencyValues := make(map[string]string, len(_quantiles))
+	for _, quantile := range _quantiles {
+		floatQuantile, err := strconv.ParseFloat(quantile, 64)
```
prashantv, line 88: use `[]float64` for `quantiles`, so we don't have to parse it here

benchmark.go
prashantv: good point. I think we should make the underlying quantile data agnostic to the output format. e.g., just use `[]float64` here, and then the output function can decide how best to format it

benchmark.go
```diff
@@ -112,13 +142,6 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
 
 	goMaxProcs := opts.setGoMaxProcs()
 	numConns := opts.getNumConnections(goMaxProcs)
-	out.Printf("Benchmark parameters:\n")
```
prashantv, line 115: the parameters are printed before we started the benchmark in the current state -- can we retain that?

for JSON, it's a little trickier, we may have to defer it to the end.

benchmark.go
```diff
@@ -187,13 +207,78 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+	} else {
+		if opts.Format != "plaintext" {
+			out.Printf("ERROR: Please specify <json> or <JSON> for JSON output. Printing plaintext output as default.\n\n")
+		}
+		outputPlaintext(overall, out, total, benchmarkOutput)
+	}
+}
+
+// JSON output helper method
+func outputJSON(overall *benchmarkState, out output, total time.Duration, benchmarkOutput BenchmarkOutput) {
```
prashantv, line 247: you can make this a function on the `BenchmarkOutput` struct

benchmark.go
```diff
@@ -187,13 +223,74 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
-	out.Printf("RPS:               %.2f\n", float64(overall.totalRequests)/total.Seconds())
+	latencyValues := overall.getLatencies(out)
+
+	summary := Summary{
+		ElapsedTime:   (total / time.Millisecond * time.Millisecond).String(),
+		TotalRequests: overall.totalRequests,
+		RPS:           float64(overall.totalRequests) / total.Seconds(),
+	}
+
+	if opts.Format == "json" || opts.Format == "JSON" {
```
jaricftw, line 238: shouldn't have to check the Format multiple times, maybe just cache the result as a boolean `formatAsJson` when you did the check above and use it here

benchmark.go
```diff
@@ -39,8 +40,37 @@ import (
 var (
 	errNegativeDuration = errors.New("duration cannot be negative")
 	errNegativeMaxReqs  = errors.New("max requests cannot be negative")
+
+	// using a global quantiles array mainly for ease of testing, and not passing the same array
+	// around to multiple functions
+	// TODO: modify quantile parameters to remove "." character and increase ease of parsing
+	_quantiles = []float64{0.5000, 0.9000, 0.9500, 0.9900, 0.9990, 0.9995, 1.0000}
```
jaricftw, line 47: this slice can just be the old `[]float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0}` right?

benchmark.go
nishakhater: Changed to print the parameters before starting benchmark for plaintext.

benchmark.go
```diff
@@ -112,13 +142,22 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+		CPUs:        goMaxProcs,
+		Connections: numConns,
+		Concurrency: opts.Concurrency,
+		MaxRequests: opts.MaxRequests,
+		MaxDuration: opts.MaxDuration.String(),
+		MaxRPS:      opts.RPS,
+	}
+
+	if opts.Format != "json" && opts.Format != "JSON" {
+		if opts.Format != "plaintext" {
```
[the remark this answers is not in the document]
nishakhater: This is the case we discussed offline where the format specified by user is "csv" or some unrecognized format. So the checks are to see if a format is specified that is not json or plaintext.

benchmark.go
```diff
@@ -188,12 +247,120 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+func (latencyMap LatencyMap) UnmarshalJSON(b []byte) error {
+	var unmarshalOutput map[string]string
+	err := json.Unmarshal(b, &unmarshalOutput)
+	if err != nil {
+		return err
+	}
+
+	for k, v := range unmarshalOutput {
+		kv := KeyVal{Key: k, Val: v}
+		latencyMap = append(latencyMap, kv)
```
jaricftw, line 360: will need to pointer receiver to actually mutate the result

benchmark.go
```diff
@@ -39,8 +41,54 @@ import (
...
+type BenchmarkParameters struct {
+	CPUs        int    `json:"cpus"`
+	Concurrency int    `json:"concurrency"`
+	Connections int    `json:"connections"`
+	MaxRequests int    `json:"maxRequests"`
+	MaxDuration string `json:"maxDuration"`
+	MaxRPS      int    `json:"maxRPS"`
+}
+
+// KeyVal stores individual quantile -> latency value mapping
```
jaricftw, line 69: name can be more descriptive, and probably don't need to be exported

benchmark.go
```diff
@@ -39,8 +41,102 @@ import (
...
```
jaricftw, line 69: s/KeyVal/latencyElement/

benchmark.go
```diff
@@ -39,8 +41,102 @@ import (
...
+type latencyElement struct {
+	quantile string
+	latency  string
+}
+
+// LatencyMap stores multiple key-value pairs for latency value output to preserve specific ordering of elements.
```
jaricftw, line 75: // LatencyMap stores latency quantile to value mappings, using a slice to preserve specific ordering of elements for serialization

benchmark.go
```diff
@@ -39,8 +41,102 @@ import (
...
+}
+
+// BenchmarkOutput stores benchmark settings and results for JSON output
+type BenchmarkOutput struct {
+	BenchmarkParameters BenchmarkParameters `json:"benchmarkParameters"`
+	Latencies           LatencyMap          `json:"latencies"`
+	Summary             Summary             `json:"summary"`
+}
+
+// MarshalJSON implements the json.Marshaler interface for the custom latencyMap
```
jaricftw, line 92: comments: s/latencyMap/LatencyMap

benchmark.go
```diff
@@ -112,13 +208,27 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+		MaxRequests: opts.MaxRequests,
+		MaxDuration: opts.MaxDuration.String(),
+		MaxRPS:      opts.RPS,
+	}
+
+	formatAsJSON := false
+	if opts.Format == "json" || opts.Format == "JSON" {
+		formatAsJSON = true
+	} else if opts.Format != "" {
+		out.Printf("Unrecognized format option <%s>, please specify <json> or <JSON> for JSON output. Printing plaintext output as default.\n\n", opts.Format)
```
jaricftw, line 225: you can use `%q` to quote, instead of doing `<%s>`. And if we change that, replace other `<>` with quotes as well

jaricftw:
LGTM, @prashantv can you also take another look

benchmark.go
```diff
@@ -39,8 +42,102 @@ import (
...
+	_quantiles     = []float64{0.5, 0.9, 0.95, 0.99, 0.999, 0.9995, 1.0}
+	_percentileMap = map[float64]string{
+		0.5:    "P50",
+		0.9:    "P90",
+		0.95:   "P95",
+		0.99:   "P99",
+		0.999:  "P999",
+		0.9995: "P9995",
+		1.0:    "P100",
+	}
```
prashantv, line 57: Since the map and the slice are closely related, I'd suggest having a single source of truth (probably the map), and generate the slice based on that.
E.g.,
```go
_percentileMap = map[float64]string{
  [...]
}
_quantiles = mapKeysPercentileMap(_percentileMap)

[.. later in the file ..]
func mapKeysPercentileMap(m map[float64]string) []float64 {
  keys := make([]float64, len(m))
  for k := range m {
    keys = append(keys, k)
  }
  sort.Float64s(keys)
  return keys
}
```

This avoids the data getting out of sync

benchmark.go
```diff
@@ -39,8 +42,102 @@ import (
 var (
 	errNegativeDuration = errors.New("duration cannot be negative")
 	errNegativeMaxReqs  = errors.New("max requests cannot be negative")
+
+	// using a global _quantiles slice and _percentileMap mainly for ease of testing, and not passing the same array
+	// around to multiple functions
```
prashantv, line 47: while passing around the globals may be easier, we likely will want to pass in the quantiles as a flag in future.

Although that makes me wonder, do we really want the hardcoded mapping from quantile to a name like P999? Can we either:
 - generate them
 - or just keep it simple and use the percentile value (perhaps with a fixed number of decimal places as the key)

cc @jaricftw

benchmark.go
```diff
@@ -39,8 +42,102 @@ import (
...
+// Summary stores the benchmarking summary
+type Summary struct {
+	ElapsedTime   string  `json:"elapsedTime"`
+	TotalRequests int     `json:"totalRequests"`
+	RPS           float64 `json:"rps"`
+}
+
+// BenchmarkOutput stores benchmark settings and results for JSON output
+type BenchmarkOutput struct {
+	BenchmarkParameters BenchmarkParameters `json:"benchmarkParameters"`
```
prashantv, line 88: nit: could this be `parameters` instead of `benchmarkParameters`

benchmark.go
```diff
@@ -39,8 +42,102 @@ import (
...
+	Summary             Summary             `json:"summary"`
+}
+
+// MarshalJSON implements the json.Marshaler interface for the custom LatencyMap
+func (latencyMap LatencyMap) MarshalJSON() ([]byte, error) {
+	var buf bytes.Buffer
+
+	buf.WriteString("{")
+	for i, kv := range latencyMap {
+		if i != 0 {
```
prashantv, line 99: nit: prefer `i > 0` since we know it's only ever positive and makes it a little cleaner to read

benchmark.go
```diff
@@ -39,8 +42,102 @@ import (
...
+	Latencies           LatencyMap          `json:"latencies"`
+	Summary             Summary             `json:"summary"`
+}
+
+// MarshalJSON implements the json.Marshaler interface for the custom LatencyMap
+func (latencyMap LatencyMap) MarshalJSON() ([]byte, error) {
+	var buf bytes.Buffer
+
+	buf.WriteString("{")
+	for i, kv := range latencyMap {
```
prashantv, line 98: JSON serialization keeps the output deterministic by sorting the keys, so if we had key names that were sorted (e.g., `0.5000` sorts correctly against `0.9995`, we wouldn't need this custom marshalling

cc @jaricftw

benchmark.go
```diff
@@ -112,13 +209,27 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+		CPUs:        goMaxProcs,
+		Connections: numConns,
+		Concurrency: opts.Concurrency,
+		MaxRequests: opts.MaxRequests,
+		MaxDuration: opts.MaxDuration.String(),
+		MaxRPS:      opts.RPS,
+	}
+
+	formatAsJSON := false
+	if opts.Format == "json" || opts.Format == "JSON" {
```
prashantv, line 223: alternatively, `if strings.ToLower(opts.Format) == "json"` which also will allow `Json`

benchmark.go
```diff
@@ -112,13 +209,27 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+		formatAsJSON = true
+	} else if opts.Format != "" {
+		out.Printf("Unrecognized format option %q, please specify 'json' or 'JSON' for JSON output. Printing plaintext output as default.\n\n", opts.Format)
+	}
+
+	if !formatAsJSON {
+		printBenchmarkParameters(out, benchmarkParameters)
+	}
+	// If format is JSON, benchmark parameters are printed after benchmark is run to maintain a single JSON blob
```
prashantv, line 232: can you move this comment to above the `if !formatAsJSON` line

benchmark.go
```diff
@@ -188,12 +296,82 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+	if formatAsJSON {
+		outputJSON(out, benchmarkParameters, latencyValues, summary)
+	} else {
+		outputPlaintext(out, latencyValues, summary)
+	}
+}
+
+// JSON output helper method
+func outputJSON(out output, benchmarkParameters BenchmarkParameters, latencyValues map[float64]time.Duration, summary Summary) {
+
```
prashantv, line 324: nit: delete blank line

benchmark.go
```diff
@@ -39,8 +42,102 @@ import (
...
+	quantile string
+	latency  string
+}
+
+// LatencyMap stores latency quantile to value mappings, using a slice to preserve specific ordering of elements for serialization.
+type LatencyMap []latencyElement
+
+// Summary stores the benchmarking summary
+type Summary struct {
+	ElapsedTime   string  `json:"elapsedTime"`
```
prashantv, line 81: For JSON, It likely makes more sense to emit something like `ElapsedTimeSeconds float64` rather than a string, which is harder to parse by tools

benchmark.go
```diff
@@ -188,12 +296,82 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+	}
+
+	if formatAsJSON {
+		outputJSON(out, benchmarkParameters, latencyValues, summary)
+	} else {
+		outputPlaintext(out, latencyValues, summary)
+	}
+}
+
+// JSON output helper method
```
prashantv, line 322: nit: this comment doesn't provide much additional context over the method name (which is pretty self-describing), so I think it can be removed. same below for the plain text

benchmark.go
```diff
@@ -188,12 +296,82 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+	if err != nil {
+		out.Fatalf("Failed to marshal benchmark output: %v\n", err)
+	}
+
+	out.Printf("%s\n\n", jsonOutput)
+}
+
+// Plaintext output helper method
+func outputPlaintext(out output, latencyValues map[float64]time.Duration, summary Summary) {
+	// Print out latencies
```
prashantv, line 351: remove comment since the code below is self-describing

benchmark.go
jaricftw: +1, this may worth doing, if we do, we should change the output error message to state the its case-insensitive, maybe `please specify 'json' (case-insensitive) ...`

benchmark.go
```diff
@@ -170,9 +214,6 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
 		}
 	}
 
-	// TODO: Support streaming updates.
-	// TOOD: Aggregate per-backend state for JSON output in future.
```
jaricftw, line 174: I don't have context on these TODOs, but seems they should stay as is?

benchmark.go
```diff
@@ -188,12 +229,75 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+		Latencies:  latencies,
+		Summary:    summary,
+	}
+
+	jsonOutput, err := json.MarshalIndent(&benchmarkOutput, "" /* prefix */, "  " /* indent */)
+	if err != nil {
+		out.Fatalf("Failed to marshal benchmark output: %v\n", err)
+	}
+
+	out.Printf("%s\n\n", jsonOutput)
```
jaricftw, line 272: do we need two newlines

benchmark.go
```diff
@@ -188,12 +229,75 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+}
+
+// Plaintext output helper method
```
jaricftw, line 275: comment isn't quite necessary since the method name is self-descripive

benchmark_test.go
```diff
@@ -218,3 +220,88 @@ func TestBenchmarkStatsPerPeer(t *testing.T) {
 	})
 	assert.Equal(t, want, statsServer.Aggregated(), "unexpected stats")
 }
+
+func TestBenchmarkOutput(t *testing.T) {
+	// Testing both plaintext and JSON output
+	tests := []struct {
+		name          string
+		opts          BenchmarkOptions
```
jaricftw, line 228: do you need the entire `BenchmarkOptions` in the input, or you can just use `format string` here, and use it to construct the option in your test logic

benchmark_test.go
```diff
@@ -218,3 +220,88 @@ func TestBenchmarkStatsPerPeer(t *testing.T) {
...
+			bufStr := buf.String()
+
+			for _, want := range tt.wantOutput {
+				assert.Contains(t, bufStr, want)
+			}
+			for _, notWant := range tt.notWantOutput {
+				assert.NotContains(t, bufStr, notWant)
+			}
+
+			if opts.BOpts.Format == "json" {
```
jaricftw, line 294: optional: if you want to test all cases of json, e.g. (json, JSON, Json, etc), you can add an additional field in the test struct, e.g. `wantJSON bool`, and check it here

benchmark_test.go
```diff
@@ -218,3 +220,88 @@ func TestBenchmarkStatsPerPeer(t *testing.T) {
...
+		{
+			name: "json",
+			opts: BenchmarkOptions{
+				Format: "json",
+			},
+			wantOutput:    []string{"summary", "benchmarkParameters", "maxRPS", "latencies"},
+			notWantOutput: []string{"Errors", "Benchmark parameters", "Max RPS"},
+		},
+		{
+			name: "plaintext",
```
jaricftw, line 241: given plaintext or the following csv isn't a format option actually, so they all fall into the default "plaintext". That means, the two test cases are the same, and you have duplicated code

one idea is to use a slice of options in the test struct. so you would have
```
tests := []struct{
   name string
   formats []string 
   /// other fields
} {
  {
   name: "json"
   formats: []string{"json", "JSON", "Json", "jSON"},
  },
   {
   name: "others"
   formats: []string{"plaintext", "csv", "blah"},
  },
}
// and in the test logic, use a loop to loop through the slice of options
for _, tt := range tests {
  for _, format := range tt.formats {
    t.Run(fmt.Sprintf("%s-%s", tt.name, format), func(t *testinglT) {
       // your current test logic
    })
  }
}
```

benchmark.go
prashantv: OK with keeping the error simple, and just saying "please specify 'json'", we just want to support other cases to follow the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle),
> Be conservative in what you send, be liberal in what you accept

benchmark.go
prashantv: Yes, I think so, let's leave these TODOs

benchmark_test.go
```diff
@@ -218,3 +220,88 @@ func TestBenchmarkStatsPerPeer(t *testing.T) {
...
+	}{
+		{
+			name:          "json",
+			format:        []string{"json", "JSON", "Json", "jsON", "jsoN"},
+			wantJSON:      true,
+			wantOutput:    []string{"summary", "benchmarkParameters", "maxRPS", "latencies"},
+			notWantOutput: []string{"Errors", "Benchmark parameters", "Max RPS", "Unrecognized format option"},
+		},
+		{
+			name:          "plaintext",
```
jaricftw, line 241: maybe name should say "unspecified"

benchmark_test.go
```diff
@@ -218,3 +220,88 @@ func TestBenchmarkStatsPerPeer(t *testing.T) {
...
+					// Creating struct from string of JSON output
+					var benchmarkOutput BenchmarkOutput
+					b := []byte(bufStr)
+					err := json.Unmarshal(b, &benchmarkOutput)
+					require.NoError(t, err)
+					assert.Equal(t, benchmarkOutput.Parameters.MaxRPS, opts.BOpts.RPS)
+					assert.GreaterOrEqual(t, opts.BOpts.MaxRequests, benchmarkOutput.Summary.TotalRequests)
+					assert.Equal(t, len(_quantiles), len(benchmarkOutput.Latencies))
+				}
+
```
jaricftw, line 303: extra line not needed

benchmark.go
```diff
@@ -112,13 +143,28 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+	}
+
+	// If format is JSON, benchmark parameters are printed after benchmark is run to maintain a single JSON blob
+	formatAsJSON := false
+	opts.Format = strings.ToLower(opts.Format)
+	if opts.Format == "json" {
+		formatAsJSON = true
+	} else if opts.Format != "" {
+		out.Printf("Unrecognized format option %q, please specify 'json' for JSON output. Printing plaintext output as default.\n\n", opts.Format)
+	}
```
prashantv, line 163: optional: can use a `switch` case to avoid mutating the `opts.Format` with a variable that's only in-scope for the `switch`, and is a little easier to read:

```go
switch format := strings.ToLower(opts.Format); format {
case "":
  // plain text
case "json":
  formatAsJSON = true
default:
  // print error
}
```

benchmark_test.go
```diff
@@ -218,3 +220,87 @@ func TestBenchmarkStatsPerPeer(t *testing.T) {
...
+				requests.Store(0)
+				buf, _, out := getOutput(t)
+				opts := Options{
+					BOpts: BenchmarkOptions{
+						MaxRequests: 10,
+						MaxDuration: 200 * time.Millisecond,
+						RPS:         120,
+						Connections: 50,
+						Concurrency: 2,
+						Format:      format,
```
prashantv, line 278: since we want a really quick benchmark (we just want to validate the output rather than the benchmark run), suggest tweaking the parameters to make it run even quicker:
 - max requests is fine, but could drop to 1
 - max duration of like 100ms
 - no RPS specified (means run as fast as possible)
 - warmup = 0
 - Connections = 1, concurrency = max requests (to make sure they all start at once)

options.go
```diff
@@ -109,6 +109,7 @@ type BenchmarkOptions struct {
 	// Benchmark metrics can optionally be reported via statsd.
 	StatsdHostPort string `long:"statsd" description:"Optional host:port of a StatsD server to report metrics"`
 	PerPeerStats   bool   `long:"per-peer-stats" description:"Whether to emit stats by peer rather than aggregated"`
+	Format         string `long:"format" description:"Prints benchmark output in either plaintext or JSON format. Default is plaintext."`
```
prashantv, line 112: Couple of suggestions here:
 - we should use the `choice:` struct tag so the help page lists all the allowed options
 - we can use `default:` struct tag to indicate the default rather than calling it out in the description
 - we should probably support "text" or "plaintext" or something in addition to empty string to mean the default plaintext outpu.

something like:
```
Format string `long:"format" choice:"text" choice:"json" default:"text" description:"Output format for benchmark results"`
```

We also should update the test to validate that `"text"` works as well as `""`.

Once you make these changes, test that the `--help` output looks good.

benchmark.go
```diff
@@ -112,13 +143,27 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
...
+
+	// If format is JSON, benchmark parameters are printed after benchmark is run to maintain a single JSON blob
+	formatAsJSON := false
+	switch format := strings.ToLower(opts.Format); format {
+	case "text", "":
+		printParameters(out, parameters)
+	case "json":
+		formatAsJSON = true
+	default:
+		out.Printf("Unrecognized format option %q, please specify 'json' for JSON output. Printing plaintext output as default.\n\n", opts.Format)
```
prashantv, line 164: this should use `out.Warnf` (so it goes to stderr instead of stdout)

PR #303, opened 2020-06-26
Include missed changelog in 0.18.0

jaricftw:


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 # 0.18.0 (2020-06-26)
 * Add support for benchmark output in JSON format.
+* Introduce yamlalias to simplify aliases in templates.
```
prashantv, line 6: Since this is internal and doesn't affect users, I think it's OK to skip

jaricftw: removed

PR #306, opened 2020-09-08
Provide proper support for responses that contain protobuf.Any

nikolavp:
By default protoreflect fails when there is an any type and it's type
url is unkonwn to in the extension registry. This change tries to use
the descriptor providers that we have registered in yab (if provided)
otherwise it will just use a simple bytes wrapper to represent the any
type. This results in a base64 encoding in yab requests.

See https://github.com/golang/protobuf/issues/747 and
https://github.com/protocolbuffers/protobuf/issues/5226 for more context
on what this task is fixing


rabbbit:
Hey @robbertvanginkel or @prashantv - I was asked to review this. I can try but I imagine you can do it 10/100 times faster? Would need to learn a bit about proto.Any :)

encoding/protobuf.go
```diff
@@ -3,6 +3,7 @@ package encoding
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/golang/protobuf/jsonpb"
```
robbertvanginkel, line 6: Nit: group imports

encoding/protobuf.go
```diff
@@ -20,6 +21,53 @@ type protoSerializer struct {
 	serviceName string
 	methodName  string
 	method      *desc.MethodDescriptor
+	anyResolver jsonpb.AnyResolver
+}
+
+// a type to wrap a raw byte slice. Especially useful for any
+// types where we can't find the value in the registry.
```
robbertvanginkel, line 28: Nit: please follow doc comment style from https://golang.org/doc/effective_go.html#commentary:
> Doc comments work best as complete sentences, which allow a wide variety of automated presentations. The first sentence should be a one-sentence summary that starts with the name being declared.

encoding/protobuf_test.go
```diff
@@ -3,6 +3,7 @@ package encoding
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/jhump/protoreflect/dynamic"
```
robbertvanginkel, line 6: Nit: import grouping.

protobuf/source.go
```diff
@@ -1,12 +1,16 @@
...
+	"github.com/jhump/protoreflect/desc"
+)
 
 // DescriptorProvider is a source of protobuf descriptor information. It can be backed by a FileDescriptorSet
 // proto (like a file generated by protoc).
 type DescriptorProvider interface {
 	// FindService returns a service descriptor for the given fully-qualified symbol name.
 	FindService(fullyQualifiedName string) (*desc.ServiceDescriptor, error)
 
+	FindMessage(messageType string) (*desc.MessageDescriptor, error)
```
robbertvanginkel, line 13: It would be good to document valid inputs for `messageType`.

encoding/protobuf.go
```diff
@@ -20,6 +21,53 @@ type protoSerializer struct {
...
+func (m *bytesMsg) Unmarshal(b []byte) error {
+	m.V = append([]byte(nil), b...)
+	return nil
+}
+
+func (r anyResolver) Resolve(typeUrl string) (proto.Message, error) {
+	mname := typeUrl
+	if slash := strings.LastIndex(mname, "/"); slash >= 0 {
+		mname = mname[slash+1:]
+	}
```
robbertvanginkel, line 57: Looking at the testcases, maps the typeurls like
`schemas.test.proto/3ae3e282-1a1f-4921-91b4-12369cfc6036/Foo` to a message. This seems like a different pattern than the default according to https://developers.google.com/protocol-buffers/docs/proto3#any, 
> The default type URL for a given message type is type.googleapis.com/_packagename_._messagename_.

I think it would be good to add a comment here indicating what type of typeurl patterns this resolver intends to support.

encoding/protobuf_test.go
```diff
@@ -188,3 +189,40 @@ func TestProtobufResponse(t *testing.T) {
...
 }
+
+func Test_anyResolver_Resolve(t *testing.T) {
+	source, err := protobuf.NewDescriptorProviderFileDescriptorSetBins("../testdata/protobuf/simple/simple.proto.bin")
+	assert.NoError(t, err)
+	tests := []struct {
+		name    string
+		typeUrl string
+		resolveType interface{}
+		wantErr bool
```
robbertvanginkel, line 200: There is no error testcase, this should either be removed or preferably an error case should be added.

encoding/protobuf.go
```diff
@@ -20,6 +21,53 @@ type protoSerializer struct {
...
+type bytesMsg struct {
+	V []byte
+}
+
+
+// A custom any resolver which will use the descriptor provider
+// and fallback to a simple byte array structure in the json.
+// This is the needed otherwise the default behaviour of protoreflect
+// is to fail with an error when it can't find a type.
+type anyResolver struct {
```
robbertvanginkel, line 38: Nit: doc comment style.

robbertvanginkel (commented):
Overall the direction looks reasonable to me. I think it would be good to declare clearly which typeurl's are supported as the testcases seem to deviate from Protobuf's default.

Please also add testcases for the new `FindMessage` in `grpcreflectSources` and usage of `bytesMsg`.

nikolavp (commented):
> Please also add testcases for the new `FindMessage` in `grpcreflectSources` and usage of `bytesMsg`.

Added more tests as well (saw the comment at a later stage. Please have a look at them, I needed to add new testdata files that contain the protobuf any.

nikolavp:
@robbertvanginkel can you please have another look at this. Feel free to add more comments as they were very helpful.

encoding/protobuf_test.go
```diff
@@ -137,34 +147,63 @@ func TestProtobufRequest(t *testing.T) {
 }
 
 func TestProtobufResponse(t *testing.T) {
+	source, err := protobuf.NewDescriptorProviderFileDescriptorSetBins("../testdata/protobuf/simple/simple.proto.bin")
+	require.Nil(t, err)
```
robbertvanginkel, line 151:
```suggestion
	require.NoError(t, err)
```

encoding/protobuf_test.go
```diff
@@ -137,34 +147,63 @@ func TestProtobufRequest(t *testing.T) {
...
+	anySource, err := protobuf.NewDescriptorProviderFileDescriptorSetBins("../testdata/protobuf/any/any.proto.bin")
+	require.Nil(t, err)
```
robbertvanginkel, line 153:
```suggestion
	require.NoError(t, err)
```

protobuf/source_reflection.go
```diff
@@ -74,6 +74,18 @@ type grpcreflectSource struct {
...
+
+	if grpcreflect.IsElementNotFoundError(err) {
+		// If we couldn't find the message through the client,
+		// return nil instead to follow the contract
+		return nil, nil
+	}
+
+	return msg, err
+}
+
```
robbertvanginkel, line 88: Please add testcases covering this method, specifically the bit that returns nil for not founds.

encoding/protobuf.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/golang/protobuf/jsonpb"
```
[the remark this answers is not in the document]
nikolavp: Fixed

encoding/protobuf_test.go
```diff
@@ -2,12 +2,22 @@ package encoding
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"testing"
 
+	"github.com/golang/protobuf/proto"
```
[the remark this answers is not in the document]
nikolavp: Fixed

Issue #311, opened 2020-11-19
Support setting maxResponseSize for grpc


rabbbit:
I believe that yab (through) yarpc has a limit of 4mb per GRPC response - for anything above that we get:

```
Failed while making call: code:resource-exhausted message:grpc: received message larger than max (4302300 vs. 4194304)
```

jronak:
This was released in [v0.19.0](https://github.com/yarpc/yab/releases/tag/v0.19.0)

PR #312, opened 2020-12-11
gRPC: Add max-response-size option to support large responses

jronak:
This change addresses #311, adds new option `max-response-size` (applicable only for gRPC requests)

### DIFF
--- a/options.go
+++ b/options.go
@@ -74,17 +74,17 @@ type RequestOptions struct {
 
 // TransportOptions are transport related options.
 type TransportOptions struct {
-	ServiceName      string            `short:"s" long:"service" description:"The TChannel/Hyperbahn service name"`
-	Peers            []string          `short:"p" long:"peer" description:"The host:port of the service to call"`
-	PeerList         string            `short:"P" long:"peer-list" description:"Path or URL of a JSON, YAML, or flat file containing a list of host:ports. -P? for supported protocols."`
-	CallerName       string            `long:"caller" description:"Caller will override the default caller name (which is yab-$USER)."`
-	RoutingKey       string            `long:"rk" description:"The routing key overrides the service name traffic group for proxies."`
-	RoutingDelegate  string            `long:"rd" description:"The routing delegate overrides the routing key traffic group for proxies."`
-	ShardKey         string            `long:"sk" description:"The shard key is a transport header that clues where to send a request within a clustered traffic group."`
-	Jaeger           bool              `long:"jaeger" description:"Use the Jaeger tracing client to send Uber style traces and baggage headers"`
-	TransportHeaders map[string]string `short:"T" long:"topt" description:"Transport options for TChannel, protocol headers for HTTP"`
-	HTTPMethod       string            `long:"http-method" description:"The HTTP method to use"`
-	MaxResponseSize  int               `long:"max-response-size" description:"Maximum response size, applicable for gRPC requests only. Default value is 4MB"`
+	ServiceName         string            `short:"s" long:"service" description:"The TChannel/Hyperbahn service name"`
+	Peers               []string          `short:"p" long:"peer" description:"The host:port of the service to call"`
+	PeerList            string            `short:"P" long:"peer-list" description:"Path or URL of a JSON, YAML, or flat file containing a list of host:ports. -P? for supported protocols."`
+	CallerName          string            `long:"caller" description:"Caller will override the default caller name (which is yab-$USER)."`
+	RoutingKey          string            `long:"rk" description:"The routing key overrides the service name traffic group for proxies."`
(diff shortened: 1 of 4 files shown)


transport/grpc.go
```diff
@@ -81,7 +82,12 @@ func newGRPC(options GRPCOptions) (*grpcTransport, error) {
 		return nil, errGRPCNoCaller
 	}
 
-	transport := grpc.NewTransport(grpc.Tracer(options.Tracer))
+	transportOptions := []grpc.TransportOption{grpc.Tracer(options.Tracer)}
+	if options.MaxResponseSize != 0 {
```
AllenLuUber, line 86: `> 0` ? Do we want to allow negative?

transport/grpc_test.go
```diff
@@ -139,6 +164,17 @@ func testBar(ctx context.Context, request *testBarRequest) (*testBarResponse, er
 	}, nil
 }
 
+func testLargeResponse(ctx context.Context, request *testBarRequest) (*testBarResponse, error) {
+	body := make([]byte, request.Size)
+	// filling body with non-zero to avoid html-escape in json encoder
+	for i := range body {
+		body[i] = 'a'
```
AllenLuUber, line 171: do you think  `bytes.Repeat([]byte{'a'},len(request.Size))` is cleaner?

transport/grpc_test.go
```diff
@@ -115,12 +115,37 @@ func TestGRPCError(t *testing.T) {
...
+func TestGRPCMaxResponseSize(t *testing.T) {
+	t.Run("With default max response size", func(t *testing.T) {
+		doWithGRPCTestEnv(t, "example-caller", 1, []transport.Procedure{
+			newTestJSONProcedure("example", "Foo::Bar", testLargeResponse),
+		}, func(t *testing.T, grpcTestEnv *grpcTestEnv) {
+			request, err := newTestJSONRequest("example", "Foo::Bar", &testBarRequest{One: "hello", Size: 1024 * 1024 * 4})
+			require.NoError(t, err)
+			_, err = grpcTestEnv.Transport.Call(context.Background(), request)
+			require.Error(t, err)
+			require.Equal(t, yarpcerrors.CodeResourceExhausted, yarpcerrors.FromError(err).Code())
```
AllenLuUber, line 130: can folder together as `require.EqualError`. see https://godoc.org/github.com/stretchr/testify/require#EqualError

transport/grpc.go
jronak: Nope, fixed it

AllenLuUber (approved):

transport/grpc_test.go
```diff
@@ -183,8 +217,13 @@ func newGRPCTestEnv(
 	caller string,
 	numInbounds int,
 	procedures []transport.Procedure,
+	maxResponseSize int,
 ) (_ *grpcTestEnv, err error) {
-	yarpcTransport := grpc.NewTransport()
+	var options []grpc.TransportOption
+	if maxResponseSize > 0 {
+		options = append(options, grpc.ServerMaxSendMsgSize(maxResponseSize))
+	}
```
prashantv, line 225: why not always set this to a larger size? for the test, right now we're tweaking 2 parameters together, so for the large size, we're tweaking max response size on the client + server max send size on the server. Ideally we only change one of the variables (client side max response size) instead of both for the test

options.go
```diff
@@ -84,6 +84,7 @@ type TransportOptions struct {
 	Jaeger           bool              `long:"jaeger" description:"Use the Jaeger tracing client to send Uber style traces and baggage headers"`
 	TransportHeaders map[string]string `short:"T" long:"topt" description:"Transport options for TChannel, protocol headers for HTTP"`
 	HTTPMethod       string            `long:"http-method" description:"The HTTP method to use"`
+	MaxResponseSize  int               `long:"max-response-size" description:"Maximum response size, applicable for gRPC requests only. Default value is 4MB"`
```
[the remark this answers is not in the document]
jronak: Not sure if we want to restrict response sizes for any other transport, `grpc-max-response-size` sounds good to me

transport/grpc_test.go
jronak: Makes sense, I have set it to 10MB

options.go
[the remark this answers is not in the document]
prashantv: yeah, I think that might be better though to be explicit. E.g., similarly above with `HTTPMethod`, it's very bound to HTTP only, but the name makes that clear.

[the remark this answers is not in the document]
AllenLuUber: yea I agree with @prashantv here. We want to be consistent with other field naming here.

rabbbit:
Thanks @jronak, I owe you a beer;)

PR #313, opened 2021-01-06
Add multiple JSON/YAML request reader

jronak:
* Auto detects the encoding of the input request
* JSON requests can be delimited by space or new-line
* YAML requests must be delimited by `---` on a newline

This change is required for supporting client gRPC streaming calls


abhinav (changes requested):
I may be misunderstanding how the pieces fit together, but from what I can tell, currently:

- We'll decode JSON to determine whether it's valid
- Decode it again, and then encode it back in the requestReader
- Decode it again in serializer.Request

For YAML,

- We'll attempt to decode JSON
- Decode YAML, then encode it into JSON
- Decode the serialized JSON in serializer.Request

Can we find a better way of plumbing information through? It's been a while since I looked at how yab works but I think the serializers are already encoding-specific. Depending on what this change intends to enable, we may be able to add support for streaming/multiple objects to a subset of the serializers.

request_reader.go
```diff
@@ -0,0 +1,131 @@
...
+	yamlDecoder *yaml.Decoder
+	jsonDecoder *json.Decoder
+}
+
+// next decodes the request from the reader as YAML or JSON and returns
+// JSON encoded byte
+func (r *requestReader) next() ([]byte, error) {
+	var v interface{}
+	if r.jsonDecoder != nil {
+		err := r.jsonDecoder.Decode(&v)
```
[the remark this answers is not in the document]
jronak: This totally makes sense and I have made the changes to use `json.RawMessage` and explicit EOF check in JSON decoder.
I was depending on reader to internally return `EOF` during decode than checking `More` in JSON decoder as YAML decoder doesn't support any such method.

request_reader.go
```diff
@@ -0,0 +1,131 @@
...
+		return nil, err
+	}
+	return gyaml.YAMLToJSON(bytes)
+}
+
+// isJSONInput returns true if input reader can be parsed into JSON correctly
+func isJSONInput(reader io.Reader) bool {
+	var v interface{}
+	jsonDecoder := json.NewDecoder(reader)
+	err := jsonDecoder.Decode(&v)
```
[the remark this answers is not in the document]
jronak: Updated it to use rawMessage, we can cache the initial decode but the overhead is for the initial phase only before benchmark or initial request

jronak:
Things you have listed are correct, this PR aims to allow parsing multiple yaml/json requests lazily from the stdin to provide interactive mode while calling bidirectional/client-side streams. Also it must work with requests provided via file/inline requests

The initial phase where attempting to find JSON compatibility happens only for once and following requests are decoded using relevant decoder
```
1 * attempt to find JSON compatibility
n * decode and encode JSON/YAML request
```
> Decode it again in serializer.Request

Nice catch, this additional decode+encode happen for protobuf and json requests, will remove this additional step in a  consecutive PR

request_reader.go
```diff
@@ -0,0 +1,134 @@
...
+	"encoding/json"
+	"io"
+
+	gyaml "github.com/ghodss/yaml"
+	"gopkg.in/yaml.v2"
+)
+
+type requestReader interface {
+	next() ([]byte, error)
+}
```
abhinav, line 33: Can we document the contract here? Things that should be answered with docs:

- what does next return
- what does next do when there are no more requests to produce

request_reader.go
```diff
@@ -0,0 +1,134 @@
...
+
+// jsonRequestReader parses multiple JSON objects from given reader
+// JSON objects can be delimited by space or newline
+type jsonRequestReader struct{ *json.Decoder }
```
abhinav, line 37: I would avoid embedding this here. The methods of `json.Decoder` should not become methods of the jsonRequestReader:

```suggestion
type jsonRequestReader struct{ dec *json.Decoder }
```

request_reader.go
```diff
@@ -0,0 +1,134 @@
...
+func (r *jsonRequestReader) next() ([]byte, error) {
+	if !r.More() {
+		return nil, io.EOF
+	}
+	var v json.RawMessage
+	err := r.Decode(&v)
+	if err != nil {
+		return nil, err
+	}
+	return []byte(v), nil
```
abhinav, line 48: This can be shortened to a single line:
```suggestion
	return []byte(v), err
```

The contract for most Go functions is that if `err != nil`, the other value is invalid/can be ignored.

request_reader.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// yamlRequestReader parses multiple YAML objects from given reader
+// YAML objects must be delimited by `---`
+type yamlRequestReader struct{ *yaml.Decoder }
```
abhinav, line 53: Same, let's not embed this.
```suggestion
type yamlRequestReader struct{ dec *yaml.Decoder }
```

request_reader.go
```diff
@@ -0,0 +1,134 @@
...
+func (r *yamlRequestReader) next() ([]byte, error) {
+	var v interface{}
+	err := r.Decode(&v)
+	if err != nil {
+		return nil, err
+	}
+	bytes, err := yaml.Marshal(v)
+	if err != nil {
+		return nil, err
+	}
```
abhinav, line 66: This is unfortunate but I suppose it'll have to do for now.

request_reader.go
```diff
@@ -0,0 +1,134 @@
...
+	return gyaml.YAMLToJSON(bytes)
```
abhinav, line 67: Are you sure we need to do this? Both, the Thrift and Proto serializers expect YAML input. (JSON is valid YAML so JSON works too, but in this case, just the raw YAML would work.)
```suggestion
	return bytes
```

request_reader.go
```diff
@@ -0,0 +1,134 @@
...
+	err := jsonDecoder.Decode(&v)
+	return err == nil
+}
+
+// newRequestReader detects the input encoding type, returns either
+// json or yaml request decoder
+// note: wraps reader under buffer to seek to the beginning for readers
+// which do not support seeking such as io.STDIN
+func newRequestReader(reader io.Reader) requestReader {
+	bufReader := newBufferReader(reader)
```
abhinav, line 83: This method of determining whether the input is JSON is a bit expensive and requires maintaining a custom bufferReader (which has unbounded memory growth for streaming inputs).

We can perhaps simplify this all significantly with a simpler heuristic and with use of [bufio.Reader](https://golang.org/pkg/bufio/#Reader), which supports [`Peek`](https://golang.org/pkg/bufio/#Reader.Peek) and [`UnreadByte`](https://golang.org/pkg/bufio/#Reader.UnreadByte).

The heuristic can be: does the input start with `{`? It's not perfect but it's good enough. If we want something smarter later, we can build something that accepts both, JSON and YAML on the same stream. But for now, we can do:

```
func isJSONInput(r *bufio.Reader) (bool, error) {
  b, err := r.ReadByte()
  ...
  isJSON := b == '{'
  err := r.UnreadByte()
  ...
```

That way we're not double decoding, nor are do we have unbounded memory growth on a custom buffered reader.

request.go
```diff
@@ -41,6 +43,25 @@ var (
...
 )
 
+// getRequestReader create request reader from inline bytes passed, through a file or stdin
+func getRequestReader(inline, file string) (requestReader, error) {
+	var reader io.Reader
+	if file == "-" || inline == "-" {
+		reader = os.Stdin
+	} else if file != "" {
+		f, err := os.Open(file)
+		if err != nil {
```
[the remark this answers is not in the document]
jronak: Seems to be leaking in the header read function as well. Managed to keep it simple by closing in the main/runWithOptions method (currently in the refactored branch)
https://github.com/yarpc/yab/blob/f98df5457749323626da58c9556f78d93ee4ee1b/request.go#L47
https://github.com/yarpc/yab/blob/f98df5457749323626da58c9556f78d93ee4ee1b/main.go#L294

request_reader.go
jronak: True, this would work in the current state. I wanted to make the contract in serializer to expect JSON input only to avoid redundant (un)marshals. This has been mostly addressed in the refactored branch

request_reader.go
jronak: +1 unbounded memory growth is concerning, this would break for many YAML inputs like `{a:b}` as decoder will consider it as JSON. Since the buffer capture is required only on the initial object parsing, I could parse it using the custom reader and post encoding detection it could stop caching the data and work only as pass through cache, this should avoid unbounded memory growth. Thoughts?

request.go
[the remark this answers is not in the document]
abhinav: Sorry, can you clarify? getHeaders doesn't leak; it uses getRequestInput, which uses ioutil.ReadFile, which cleans up after itself.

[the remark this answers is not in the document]
jronak: My bad, you are right it doesn't leak

request_reader.go
abhinav: I'm concerned that short of something super custom and complicated, any version that doesn't attempt full decoding is not going to be perfect because all JSON is valid YAML.

I would maybe opt for a simple-ish heuristic at the cost of imperfect detection for now.

jronak: Makes sense, will go ahead with simple decoder for now :-)

jronak:
@abhinav closing this PR in favour of another set of stacked PRs

PR #314, opened 2021-01-08
gRPC streams support - 1

jronak:
## Aim
Support gRPC streams in Yab

## Changes
* Exposes a new interface in serializer for creating stream request and multiple stream message body reader
* Refactors `getRequestInput` to provider a reader interface which would be used for reading multiple stream message body

Next Stack: #315


encoding/encoding.go
```diff
@@ -51,6 +52,20 @@ type Serializer interface {
...
+	// NextBody returns the encoded request body if available, and if not, returns an
+	// io.EOF error
+	NextBody() ([]byte, error)
+}
+
+// StreamSerializer interface exposes additional methods to handle stream requests
+type StreamSerializer interface {
+	// StreamRequest creates a root stream request, a stream request reader using
+	// body reader provided
+	StreamRequest(body io.Reader) (*transport.Request, StreamRequestReader, error)
```
Dogild, line 66: One improvement we could do with this method is to have a `*transport.StreamRequest`, this will give the following benefits:
- No confusion on when to use the method `Request` or `StreamRequest`
- No confusion if `*transport.Request` is being used manually. You will one type for streaming and another for unary. The big benefits is that `*transport.StreamRequest` will not have a Body field, this will remove confusion about using `*transport.Request` with a stream request.
- No need to do the extra check in the methods `Request` and `StreamRequest`. You will not be able to use request with streaming anyway.

Thoughts?

encoding/inputdecoder/decoder.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// jsonInputDecoder parses multiple JSON objects from given reader
+// JSON objects can be delimited by space or newline
+type jsonInputDecoder struct{ dec *json.Decoder }
+
+func (r *jsonInputDecoder) Next() ([]byte, error) {
+	if !r.dec.More() {
+		return nil, io.EOF
+	}
```
Dogild, line 46: nit: missing break line after the if statement, this helps for readability

for instance it should be:

```
if !r.dec.More() {
 		return nil, io.EOF
 }
 
 var v json.RawMessage
 err := r.dec.Decode(&v)
 return []byte(v), err
```

Note: there are a lot of locations where this is style things, this is a comment for all of them

encoding/encoding.go
jronak: StreamRequest does make sense, but it's exactly replica of `transport.Request` except the body which would be nil in case of a streaming request. Also, it would again need bunch of duplicate implementation of certain request enricher and middleware watchers
https://github.com/yarpc/yab/blob/8f466890aa39477328191e0eccfdef4d04190eb2/request.go#L178

Transport reuses the `transport.Request` to transport to YARPC stream cleanly
https://github.com/yarpc/yab/blob/f84a2941e8375ef8f980a41aeefa650059af1b0d/transport/grpc.go#L146

I believe this would overwhelm the benefits

encoding/protobuf.go
```diff
@@ -144,11 +141,53 @@ func (p protoSerializer) Response(body *transport.Response) (interface{}, error)
 	return unmarshaledJSON, nil
 }
 
+func (p protoSerializer) StreamRequest(body io.Reader) (*transport.StreamRequest, StreamRequestReader, error) {
+	decoder, err := inputdecoder.New(body)
+	if err != nil {
+		return nil, nil, err
+	}
+	reader := protoStreamRequestReader{
```
Dogild, line 149: nit: missing break line

encoding/protobuf.go
```diff
@@ -144,11 +141,53 @@ func (p protoSerializer) Response(body *transport.Response) (interface{}, error)
...
+		decoder: decoder,
+		proto:   p,
+	}
+	streamReq := &transport.StreamRequest{
+		Request: &transport.Request{
+			Method: procedure.ToName(p.serviceName, p.methodName),
+		},
+	}
+	return streamReq, reader, nil
```
Dogild, line 158: nit: missing break line

encoding/protobuf.go
```diff
@@ -144,11 +141,53 @@ func (p protoSerializer) Response(body *transport.Response) (interface{}, error)
...
 	_, err := p.Response(body)
 	return err
 }
 
+func (p protoSerializer) encode(jsonBytes []byte) ([]byte, error) {
+	req := dynamic.NewMessage(p.method.GetInputType())
+	if err := req.UnmarshalJSON(jsonBytes); err != nil {
+		return nil, fmt.Errorf("could not parse given request body as message of type %q: %v", p.method.GetInputType().GetFullyQualifiedName(), err)
+	}
+	bytes, err := proto.Marshal(req)
```
Dogild, line 171: nit: missing break line

encoding/protobuf.go
```diff
@@ -144,11 +141,53 @@ func (p protoSerializer) Response(body *transport.Response) (interface{}, error)
...
+	if err != nil {
+		return nil, fmt.Errorf("could marshal message of type %q: %v", p.method.GetInputType().GetFullyQualifiedName(), err)
+	}
+	return bytes, nil
```
Dogild, line 175: nit: missing break line

encoding/protobuf.go
```diff
@@ -144,11 +141,53 @@ func (p protoSerializer) Response(body *transport.Response) (interface{}, error)
...
+	}
+	return streamReq, reader, nil
+}
+
 func (p protoSerializer) CheckSuccess(body *transport.Response) error {
 	_, err := p.Response(body)
 	return err
 }
 
+func (p protoSerializer) encode(jsonBytes []byte) ([]byte, error) {
```
Dogild, line 166: love it <3

encoding/protobuf.go
```diff
@@ -144,11 +141,53 @@ func (p protoSerializer) Response(body *transport.Response) (interface{}, error)
...
+	decoder inputdecoder.Decoder
+	proto   protoSerializer
+}
+
+func (p protoStreamRequestReader) NextBody() ([]byte, error) {
+	body, err := p.decoder.Next()
+	if err != nil {
+		return nil, err
+	}
+	return p.proto.encode(body)
```
Dogild, line 188: nit: missing break line

transport/interface.go
```diff
@@ -40,6 +40,11 @@ type Request struct {
 	Body             []byte
 }
 
+// StreamRequest is a wrapper of Request, to be used for streaming RPC
+type StreamRequest struct {
+	Request *Request
+}
+
```
Dogild, line 47: We can even do better with `type StreamRequest Request`

Dogild (approved):
Added one additional change we should do.

Lgtm otherwise! @AllenLuUber please take a look

encoding/encoding.go
```diff
@@ -51,6 +52,20 @@ type Serializer interface {
 	CheckSuccess(body *transport.Response) error
 }
 
+// StreamRequestReader interface exposes method to read multiple request body
+type StreamRequestReader interface {
+	// NextBody returns the encoded request body if available, and if not, returns an
+	// io.EOF error
```
AllenLuUber, line 58: The comment here is a bit weird. Interface definition should not mention it is a `io.EOF` error since this sounds like an implementation details.
It should just say, `if not, returns error why it is unavailable`

encoding/encoding.go
```diff
@@ -51,6 +52,20 @@ type Serializer interface {
 	CheckSuccess(body *transport.Response) error
 }
 
+// StreamRequestReader interface exposes method to read multiple request body
```
AllenLuUber, line 55: `// StreamRequestReader reads multi-part request body`

request_test.go
```diff
@@ -114,7 +114,9 @@ func TestGetRequestInput(t *testing.T) {
 		}
 
 		if assert.NoError(t, err, "getRequestInput(%v, %v) should not fail", tt.inline, tt.file) {
-			assert.Equal(t, tt.want, got, "getRequestInput(%v, %v) mismatch", tt.inline, tt.file)
+			contents, err := ioutil.ReadAll(got)
+			assert.NoError(t, err)
```
AllenLuUber, line 118: nit: `require.NoError(t, err)`

If this check fails, the following `assert.Equal` will likely fail as well. To avoid having multiple consecutive failures, use `require` here to bail out early.

encoding/encoding.go
```diff
@@ -51,6 +52,20 @@ type Serializer interface {
...
+type StreamRequestReader interface {
+	// NextBody returns the encoded request body if available, and if not, returns an
+	// io.EOF error
+	NextBody() ([]byte, error)
+}
+
+// StreamSerializer interface exposes additional methods to handle stream requests
```
AllenLuUber, line 62: for consistency, the current `Serializer` interface comment reads as:
`Serializer serializes and deserializes data for a specific encoding and method.`
You should probably do `StreamSerializer serializes and deserializes data for a stream request` ?

encoding/inputdecoder/decoder_test.go
```diff
@@ -0,0 +1,133 @@
...
+		buf := bufio.NewReader(bytes.NewReader([]byte(`{}`)))
+		isJSON, err := isJSONInput(buf)
+		assert.True(t, isJSON, "expected true")
+		assert.NoError(t, err, "unexpected error")
+	})
+	t.Run("invalid json", func(t *testing.T) {
+		buf := bufio.NewReader(bytes.NewReader([]byte(`test: json`)))
+		isJSON, err := isJSONInput(buf)
+		assert.False(t, isJSON, "expected false")
+		assert.NoError(t, err, "exepected error")
```
AllenLuUber, line 43: typo

AllenLuUber (approved):

encoding/encoding.go
Dogild: sync offline - @jronak agreed to have StreamRequest

transport/interface.go
Dogild: sync offline.

We decided to keep it like this to have a more flexible way of enhancing `StreamRequest` in the future if needed

Dogild (approved):

encoding/inputdecoder/decoder.go
```diff
@@ -0,0 +1,96 @@
...
+
+	gyaml "github.com/ghodss/yaml"
+	"gopkg.in/yaml.v2"
+)
+
+// Decoder interface exposes method for reading multiple input requests
+type Decoder interface {
+	// NextJSONBytes returns JSON marshaled bytes of the next request body
+	// It returns io.EOF when there are no more requests
+	NextJSONBytes() ([]byte, error)
```
prashantv, line 36: Why are we using JSON here? The current serializer interface uses YAML as far as I can tell (and the specific encoding converts YAML to JSON if required, e.g., for Proto)
https://github.com/yarpc/yab/blob/dev/encoding/protobuf.go#L110

Are we planning to change the `Serializer` interface to work with JSON? If not, then we probably should use YAML here as well for consistency, and to avoid unnecessarily converting YAML to JSON if the underlying encoder (E.g., thrift) can work fine with YAML types.

YAML may allow us to express types that JSON cannot, so the conversion may not be lossless.

jronak: `Serializer` will continue to use YAML input, I thought of Protobuf encoder while coming up with input-decoder interface and used JSON since they eventually convert YAML to JSON like you pointed out. 

I agree with the lossless conversion point, but given this decoder would be used only with protobuf encoder for streaming requests can we assume it would be safe to have this interface?

prashantv: True that it'll only be used in protobuf, but since this is defined as a more general interface, I think it's preferred to use YAML here (for consistency), and push the lossy conversion closer to where it's required. I'd like to convert YAML to protobuf directly too one day..

jronak: Addressed this in #321

PR #315, opened 2021-01-08
gRPC streams support - 2

jronak:
## Aim
Support gRPC streams in Yab

## Changes
* Extend `StreamSerializer` to expose type of streaming method information
* Add guards in `Serializer.Request` and `StreamRequestReader.NextBody` to return error if invoked with incorrect rpc method

Next Stack: #316


encoding/protobuf.go
```diff
@@ -112,19 +113,42 @@ func (p protoSerializer) Encoding() Encoding {
...
 
+func (p protoSerializer) IsClientStreaming() bool {
+	return p.method.IsClientStreaming()
+}
+
+func (p protoSerializer) IsServerStreaming() bool {
+	return p.method.IsServerStreaming()
+}
+
+func (p protoSerializer) StreamRequest() ([]byte, error) {
```
Dogild, line 124: The name of this API is odd.

We have `Request()` and `StreamRequest()`, one is returning a `*transport.Request` and the other one returns a `[]byte`. We should rename this method IMO, maybe `NextStreamRequestBody()`

encoding/protobuf.go
```diff
@@ -112,19 +113,42 @@ func (p protoSerializer) Encoding() Encoding {
...
+	return p.encode(jsonBytes)
+}
+
 func (p protoSerializer) Request() (*transport.Request, error) {
+	if p.IsClientStreaming() || p.IsServerStreaming() {
+		return &transport.Request{
+			Method: procedure.ToName(p.serviceName, p.methodName),
+		}, nil
+	}
+
```
Dogild, line 141: This is a review not only for this PR but for https://github.com/yarpc/yab/pull/314, https://github.com/yarpc/yab/pull/316 and https://github.com/yarpc/yab/pull/317. This change is the cause of everything :)

What is the real reason of changing `Request(input []byte) (*transport.Request, error)` to `Request() (*transport.Request, error)` ? If we create a new method `StreamRequest(io.Reader)` and `NextStreamRequestBody()` would we need all of the changes with `Request` and the constructor?
With this proposal we are in a weird state where `Request()` does a lot of things that nothing the API indicates (encoding, special case for streaming)

Dogild: Below a solution we could have to avoid to break APIs and have consistent/logic APIs naming:

First, we could create a new StreamRequestReader for the stream request:

```
type StreamRequestReader interface {
	NextBody() ([]byte, error)
}
type protoStreamerSerializer struct {
	reader io.Reader
	s protoSerializer
}
func (p protoStreamerSerializer)NextBody() (body []byte, error){
        // private encode method to encode a grpc body
	return p.s.encode(p.reader.Next())
}

func (p protoSerializer) StreamRequest(body io.Reader) (*transport.Request, StreamRequestReader, error) {
....
}
```

Serialize and StreamSerializer  interface:
```
// Serializer serializes and deserializes data for a specific encoding and method.
type Serializer interface {
	// Encoding returns the encoding for this serializer.
	Encoding() Encoding
	// Request creates a transport.Request from the given []byte input.
        // returns an error if reader has several entries
	Request(body io.Reader) (*transport.Request, error)
	// Response converts a transport.Response into something that can be displayed to a user.
	// For non-raw encodings, this is typically a map[string]interface{}.
	Response(body *transport.Response) (response interface{}, err error)
	// CheckSuccess checks whether the response body is a success, and if not, returns an
	// error with the failure reason.
	CheckSuccess(body *transport.Response) error
}

type StreamSerializer interface {
   StreamRequest(body io.Reader) (*transport.Request, StreamRequestReader, error)
}
```
Here we have `StreamRequest` and `Request` which both take a io.reader param (so it is consistent in the API).

What I would also recommend here:
- Request should call a private method `encode` to encode the body
- Request return an error if being used with a streaming API
- Request could return an error if several entries in the io.reader? (I am still thinking of the value of this thought, what do you think?)

Thoughts?

jronak: This looks great, something on the lines of how I am refactoring it currently

>Request could return an error if several entries in the io.reader? (I am still thinking of the value of this thought, what do you think?)

I would avoid doing any additional input decoding to find if there are more than one request since this issue exists currently, will fix this in a separate PR

encoding/encoding.go
```diff
@@ -64,6 +64,12 @@ type StreamSerializer interface {
 	// StreamRequest creates a root stream request, a stream request reader using
 	// body reader provided
 	StreamRequest(body io.Reader) (*transport.Request, StreamRequestReader, error)
+
+	// IsClientStreaming returns true for client streaming methods
+	IsClientStreaming() bool
+
+	// IsServerStreaming returns true for server streaming methods
+	IsServerStreaming() bool
```
Dogild, line 72: I like this! Way simpler and invasive!

Did you think of having a a method `Type() type` in the Serializer interface?
We have something very similar in yarpc-go (https://github.com/yarpc/yarpc-go/blob/dev/api/transport/handler.go#L31). I was wondering if we could do something similar

jronak: We could add `MethodType() type` which possibly enumerate to `unary`, `clientStream`, `serverStream` or `bidiStream`

encoding/protobuf.go
jronak: I haven't modified `serializer.Request([]byte)` to use `body io.Reader` in this PR, this is to keep the change very specific to streaming methods only.

IMO, since unary requests have single request having a body io.Reader might be an overkill. This method could continue expecting the bytes from caller.

encoding/encoding.go
```diff
@@ -32,6 +32,19 @@ import (
 	"github.com/yarpc/yab/unmarshal"
 )
 
+type methodType int
```
Dogild, line 35: `methodType` should be a public type

encoding/protobuf.go
```diff
@@ -106,7 +106,19 @@ func (p protoSerializer) Encoding() Encoding {
 	return Protobuf
 }
 
+func (p protoSerializer) IsClientStreaming() bool {
+	return p.method.IsClientStreaming()
+}
+
+func (p protoSerializer) IsServerStreaming() bool {
+	return p.method.IsServerStreaming()
```
Dogild, line 114: No need of the methods `IsServerStreaming` and `IsClientStreaming` anymore

encoding/protobuf.go
```diff
@@ -106,7 +106,19 @@ func (p protoSerializer) Encoding() Encoding {
...
+}
+
 func (p protoSerializer) Request(body []byte) (*transport.Request, error) {
+	if p.MethodType() != Unary {
+		return nil, fmt.Errorf("request method must not be invoked for a streaming rpc method: %q", p.method.GetInputType().GetFullyQualifiedName())
```
Dogild, line 119: nit: the error message should be that request can only be used with unary method. (this error message would become invalid if another type would be added)

encoding/protobuf.go
```diff
@@ -163,6 +180,17 @@ func (p protoSerializer) CheckSuccess(body *transport.Response) error {
...
 }
 
+func (p protoSerializer) MethodType() methodType {
+	if p.method.IsClientStreaming() && p.method.IsServerStreaming() {
+		return BidirectionalStream
+	} else if p.method.IsClientStreaming() {
+		return ClientStream
+	} else if p.method.IsServerStreaming() {
+		return ServerStream
+	}
```
Dogild, line 190: nit: for better code readability

```
if p.method.IsClientStreaming() && p.method.IsServerStreaming() {
   return BidirectionalStream
}

if p.method.IsClientStreaming() {
   return ClientStream
}
 
if p.method.IsServerStreaming() {
   return ServerStream
}
 
return Unary
```

encoding/encoding.go
```diff
@@ -32,6 +32,20 @@ import (
 	"github.com/yarpc/yab/unmarshal"
 )
 
+// MethodType is type of the RPC method
+type MethodType int
+
+const (
+	// Unary method type is a traditional RPC
+	Unary MethodType = iota
```
Dogild, line 40: nit: `Unary MethodType = iota + 1`

encoding/encoding.go
```diff
@@ -32,6 +32,20 @@ import (
...
+	// ClientStream method type RPC sends multiple messages
+	ClientStream
+	// ServerStream method type RPC receives multiple messages
+	ServerStream
+	// BidirectionalStream type RPC can send and receive multiple messages
+	BidirectionalStream
+)
```
Dogild, line 47: nit: to be consistent, let's use the same semantic as the yarpc now for the naming and comment. See https://github.com/yarpc/yarpc-go/blob/dev/api/transport/handler.go#L40

encoding/encoding.go
```diff
@@ -64,6 +78,9 @@ type StreamSerializer interface {
 	// StreamRequest creates a root stream request, a stream request reader using
 	// body reader provided
 	StreamRequest(body io.Reader) (*transport.StreamRequest, StreamRequestReader, error)
+
+	// MethodType returns the type of RPC method
+	MethodType() MethodType
```
Dogild, line 83: IMO we should add this methodic the `Encoding` interface. I understand that you only added here since only protobuf needs it. However, this method is valid to have for other encoding (in comparaison `StreamRequest` is not since it can only be used with proto)

encoding/encoding.go
jronak: The use-case here is quite different from YARPC, `MethodType` is verbose to identify different types of streaming RPCs as well which requires `client`, `server` and `bidirectional` stream

encoding/encoding.go
jronak: I'm in favour of keeping this method in `StreamSerializer` since other serializers do not support anything other than Unary which adds unnecessary boilerplate just like we found during the review initially and while only the streaming serializers have different RPC types (listed in MethodType)

Dogild: This is different than the initial review and the extra boilerplate.

In the first proposal, you needed to add extra methods for encoding that were not supported streaming. This was adding complexity, verbosity and unnecessary codes.

In this case, `MethodType` return some informations of the streaming, like `Unary`, `Bidi` etc..If you add this method to another encoding than protobuf, the method can be used and will return a valid result. `Unary` can be used for something else than just protobuf

You are actually using this method for other type of encoding, see https://github.com/yarpc/yab/pull/317/files 
For instance here:
```
func isStreamingMethod(serializer encoding.Serializer) bool {
	streamSerializer, ok := serializer.(encoding.StreamSerializer)
	if !ok {
		return false
	}

	return streamSerializer.MethodType() != encoding.Unary
```
Here we could avoid the cast to `encoding.StreamSerializer` and only use the `MethodType`.

jronak: Non-streaming encoders are not going to have any other type of RPC method apart from Unary, IMO MethodType in streamSerializer should be okay. And I agree something meaningful can be returned from other encoders though all of them would return `Unary` (except proto-encoder)

Dogild (approved):

AllenLuUber (approved):

PR #316, opened 2021-01-09
gRPC streams support - 3

jronak:
## Aim
Support gRPC streams in Yab

## Changes
* Add new interface `StreamTransport` which exposes method to call streaming requests
* Add `CallStream` method in gRPC transport 

Next Stack: #317


Dogild (approved):

Dogild (approved):

Dogild (approved):

transport/grpc_test.go
```diff
@@ -21,17 +21,25 @@
...
 	"strings"
 	"testing"
+	"time"
 
+	"github.com/golang/protobuf/proto"
 	"github.com/opentracing/opentracing-go"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/yarpc/yab/testdata/protobuf/simple"
+	googleGrpc "google.golang.org/grpc"
```
AllenLuUber, line 41: we don't use camelCase. Stick with `googlegrpc` or `google_grpc`

transport/grpc_test.go
```diff
@@ -216,6 +301,7 @@ func newGRPCTestEnv(
 	caller string,
 	numInbounds int,
 	procedures []transport.Procedure,
+	encoding string,
```
AllenLuUber, line 304: why do we need to pipe in this info? It is not getting used?

transport/interface.go
```diff
@@ -73,6 +74,12 @@ type Transport interface {
 	Tracer() opentracing.Tracer
 }
```
AllenLuUber, line 75: Can we improve this comment with s/calls/unary calls?

transport/grpc.go
```diff
@@ -133,10 +135,29 @@ func (t *grpcTransport) Call(ctx context.Context, request *Request) (*Response,
...
 
+func (t *grpcTransport) CallStream(ctx context.Context, request *StreamRequest) (*transport.ClientStream, error) {
+	return t.StreamOutbound.CallStream(ctx, t.requestToYARPCStreamRequest(request))
+}
+
 func (t *grpcTransport) Close() error {
 	return multierr.Combine(t.Transport.Stop(), t.Outbound.Stop())
 }
 
+func (t *grpcTransport) requestToYARPCStreamRequest(streamRequest *StreamRequest) *transport.StreamRequest {
```
AllenLuUber, line 146: seeing a bunch of `streamRequest.Request.Field` here. Do we need to check nil on `streamRequest.Request` before we use it? Actually we might even need to check nil on `streamRequest` itself as well?

transport/grpc_test.go
```diff
@@ -95,23 +103,99 @@ func TestGRPCConstructor(t *testing.T) {
...
+		Tracer:    opentracing.NoopTracer{},
+		Caller:    "test",
+		Encoding:  "proto",
+	})
+	assert.NoError(t, err)
+
+	streamClient, ok := client.(StreamTransport)
+	assert.True(t, ok)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
AllenLuUber, line 173: is `time.Minute` too long? Don't want make the CI/local test too long

jronak: Ah my bad, didn't replace it with second after I finished local debugging

AllenLuUber (approved):

PR #317, opened 2021-01-09
gRPC streams support - 4

jronak:
## Aim
Support gRPC streams in Yab

## Changes
* Add stream request dispatcher for streaming RPC methods
* Supports interactive IO for streaming methods, i.e. for bi-directional streams you could input the stream message in STDIN and see the stream response
* Concurrently send and receive stream messages for Bidirectional streaming RPC

Next Stack: #319

## How to test?
Download [grpc-stream-test-server.zip](https://github.com/yarpc/yab/files/5826809/grpc-stream-test-server.zip), follow the `readme.md` to run the test server and dispatch streaming yab queries

### Client Stream
```
yab  -F testprotoset --service "test-server" --method uber.yarpc.encoding.protobuf.Test/ClientStream -p grpc://localhost:55555 -r '{"value": "hi"}{"value": "world"}'

{
    "value" "2" // count of the number of requests sent
}
```
STDIN mode
```
yab  -F testprotoset --service "test-server" --method uber.yarpc.encoding.protobuf.Test/ClientStream -p grpc://localhost:55555 -r '-'

{"value": "req-msg-1"}
{"value": "req-msg-2"}

{
    "value" "2" // count of the number of requests sent
}
```

### Server Stream
```
yab  -F testprotoset --service "test-server" --method uber.yarpc.encoding.protobuf.Test/ServerStream -p grpc://localhost:55555 -r '{"value": "hi"}'

// Responds with n messages, where n = len(req.value)
{
  "value": "hi"
}

{
  "value": "hi"
}
```

### Bidirectional Stream
```
yab -F testprotoset --service "test-server" --method uber.yarpc.encoding.protobuf.Test/Stream -p grpc://localhost:55555 -r '{"value": "h"}{"value": "i"}{"value": " world"}'
{
  "value": "h"
}

{
  "value": "hi"
}

{
  "value": "hi world"
}
```

STDIN mode

```
yab -F testprotoset --service "test-server" --method uber.yarpc.encoding.protobuf.Test/Stream -p grpc://localhost:55555 -r '-'
{"value": "req-1"}
{
  "value": "req-1"
}

{"value": "req-2"}
{
  "value": "req-1req-2"
}

```


main.go
```diff
@@ -349,9 +353,16 @@ func runWithOptions(opts Options, out output, logger *zap.Logger) {
 		out.Fatalf("Failed while preparing the request: %v\n", err)
 	}
```
Dogild, line 355: I can not comment on the line I would like, but line:342.

Why are we calling `serializer.Request()` for a streaming call? Should we call `serializer.StreamRequest()`?
This will help to not have a mixt usage of stream and unary request with the method `Request`

main.go
```diff
@@ -349,9 +353,16 @@ func runWithOptions(opts Options, out output, logger *zap.Logger) {
...
-		makeInitialRequest(out, transport, serializer, req)
+		if isStreamingCall {
+			makeInitialStreamRequest(out, transport, serializer, req)
+		} else {
+			makeInitialRequest(out, transport, serializer, req)
+		}
+
 	}
 
 	runBenchmark(out, logger, opts, resolved, benchmarkMethod{
```
Dogild, line 368: Based on the design, benchmark with streaming should be handle this:

```
-r '{"value": "message-1"}{"value": "message-2"}'
```

How does the benchmark will work here? Based on the code, benchmark will call the method `call` on the transport and only the same message will be used. How is the next message will be useed? Did I miss something?

Dogild (commented):
Please update the summary of the commit with example of benchamarks and streaming request working. This should match with what was defined in the ERD

main.go
jronak: `serializer.Request()` is needed for both unary and stream requests. It holds the metadata like procedure-name, shard-key, headers, timeout, etc. (but not the request-body) which is needed to establish the stream

While `serializer.StreamRequest()` fetches request to be dispatched once the stream has been established

main.go
jronak: We are missing an additional PR (#5) here, PR#5 would address the benchmarking of the streams

main.go
```diff
@@ -474,8 +520,111 @@ func makeRequestWithTracePriority(t transport.Transport, request *transport.Requ
...
+		clientStreaming := streamSerializer.IsClientStreaming()
+		for {
+			eof := sendStreamMessage(ctx, stream, streamMsgReader, out)
+			// send message only once if the method is server streaming
+			// since server streaming method expects only one request
+			if eof || serverStreaming {
+				break
+			}
+		}
+		closeStream(ctx, stream, out)
```
Dogild, line 618: Something I am not sure to understand here. In this loop tou will first send all of the messages, close the stream and then read the response? How this is working if the stream is closed?

main.go
```diff
@@ -474,8 +520,111 @@ func makeRequestWithTracePriority(t transport.Transport, request *transport.Requ
...
+		for {
+			eof := receiveStreamMessage(ctx, stream, serializer, out)
+			// receive message only once if the method is client streaming
+			// since client streaming method expects only one response
+			if eof || clientStreaming {
+				break
+			}
+		}
+	}
```
Dogild, line 627: I have few questions about this block of codes:

- In the bi-directional stream mode, we will be in a mode where we send_a_message/get_a_message/send_a_message/get_a_message etc. Is it what we want? This is a very predictable way of streaming but this is not always the case. Should we have concurrent read and write here?
- In the design part of this feature, we had an option called `stream-interval` for sending messages from the client. Where are we going to use this one?

main.go
```diff
@@ -366,6 +379,34 @@ func runWithOptions(opts Options, out output, logger *zap.Logger) {
...
+	return streamSerializer.StreamRequest(body)
+}
+
+func isStreamingMethod(serializer encoding.Serializer) bool {
+	streamSerializer, ok := serializer.(encoding.StreamSerializer)
+	if !ok {
+		return false
+	}
+	return streamSerializer.IsClientStreaming() || streamSerializer.IsServerStreaming()
```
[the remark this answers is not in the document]
jronak: Agree, it is the responsibility of the encoding layer while naming it serializer is unfortunate. I think moving proto-method descriptor could be tackled in another PR, but that would require overlap between transport and proto-serializer. 

IMO, transport shouldn't know about the type of RPC, and it must just provide an interface to send and receive messages while the number of messages to be sent to controlled by the encoding layer (bidi-stream, client-stream, etc.)

main.go
jronak: Naming of `stream.Close` in YARPC is wrong 😐
Close actually closes the stream from the client side while server can still send the messages

https://github.com/yarpc/yarpc-go/blob/2cb4fe14284bfbf430f9846075fb1c61792a2527/transport/grpc/stream.go#L158

[CloseSend gRPC doc](https://godoc.org/google.golang.org/grpc#ClientStream)

main.go
jronak: * Assumption to have predictable stream in initial request was to enable simple interactive bidirectional streaming i.e.
```
{req-1} // user-input
{res-1} // server-response
{req-2} // user-input
{res-2} // server-response
...
```
Your point is valid though, we might have bidi-streams sending non-uniform responses.

* `stream-interval` will be used for non-interactive mode of request input, I'd prefer adding additional options in a separate PR once these changes are good to merge

main.go
```diff
@@ -335,29 +338,39 @@ func runWithOptions(opts Options, out output, logger *zap.Logger) {
...
 	}
 
 	// Only make the request if the user hasn't specified 0 warmup.
 	if !(opts.BOpts.enabled() && opts.BOpts.WarmupRequests == 0) {
-		makeInitialRequest(out, transport, serializer, req)
+		if streamMsgReader != nil {
+			makeInitialStreamRequest(out, t, serializer, streamMsgReader, req)
+		} else {
+			makeInitialRequest(out, t, serializer, req)
+		}
```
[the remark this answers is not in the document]
jronak: Spent some time on moving it into explicit handlers which avoids cluttering main.go by moving it to a separate file to handle unary and stream request

handler.go
```diff
@@ -0,0 +1,318 @@
...
+
+			err = sendStreamMessage(ctx, stream, reqBody)
+		}
+
+		if err != nil {
+			sendErr = err
+		}
+	}()
+
+	var err error
```
Dogild, line 261: nit: rename this to `responseErr`, so we have `sendErr` and `responseErr`, easier to know what err match to what ;)

handler.go
```diff
@@ -0,0 +1,318 @@
...
+		var err error
+		for err == nil {
+			var reqBody []byte
+			reqBody, err = reqSupplier()
+			if err == io.EOF {
+				err = closeSendStream(ctx, stream)
+				break
+			}
+			if err != nil {
+				cancel()
```
Dogild, line 249: can you add a comment why this cancel is needed?

handler.go
```diff
@@ -0,0 +1,318 @@
...
+
+		var err error
+		for err == nil {
+			var reqBody []byte
+			reqBody, err = reqSupplier()
+			if err == io.EOF {
+				err = closeSendStream(ctx, stream)
+				break
+			}
+			if err != nil {
```
Dogild, line 248: nit: missing break line

handler.go
```diff
@@ -0,0 +1,318 @@
...
+	res, err := receiveStreamMessage(ctx, stream)
+	if err != nil {
+		return err
+	}
+	return resHandler(res)
+}
+
+// makeBidiStream starts bi-directional streaming rpc
+func makeBidiStream(ctx context.Context, stream *yarpctransport.ClientStream,
+	reqSupplier streamRequestSupplier, resHandler streamResponseHandler) error {
```
Dogild, line 227: nit: `reqSupplier` to `nextBodyFn`, `resHandler` to `responseHandler`

handler.go
```diff
@@ -0,0 +1,318 @@
...
+				err = closeSendStream(ctx, stream)
+				break
+			}
+			if err != nil {
+				cancel()
+				break
+			}
+
+			err = sendStreamMessage(ctx, stream, reqBody)
+		}
```
Dogild, line 254: Is there any chance to reuse the code from `makeClientStream` ? Also is there any reason those code are not the same since they have the same behavior?

You should also add a TODO that the message sent will be controlled by a timer once the option --stream-interval is available

handler.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+// sendStreamMessage sends the stream message using message body provided
+func sendStreamMessage(ctx context.Context, stream *yarpctransport.ClientStream, msgBody []byte) error {
+	req := &yarpctransport.StreamMessage{Body: ioutil.NopCloser(bytes.NewReader(msgBody))}
+	if err := stream.SendMessage(ctx, req); err != nil {
+		return fmt.Errorf("Failed while sending stream request: %v", err)
+	}
+	return nil
+}
```
Dogild, line 291: Should we also print the message we are sending? This can be very helpful in a streaming mode IMO, without it it will be hard to match responses with requests.

Dogild (commented):
Lot of comments but this is good work! we are almost there ;)

Can you update the commit summary and add example of yab request that work with streaming? From what I understand, this PR should enable this feature already

handler.go
```diff
@@ -0,0 +1,318 @@
...
+	if err != nil {
+		r.out.Fatalf("Failed while preparing the request: %v\n", err)
+	}
+
+	requestSupplier := func() ([]byte, error) {
+		msg, err := streamMsgReader.NextBody()
+		if err == io.EOF {
+			return nil, err
+		}
+		if err != nil {
```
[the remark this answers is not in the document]
Dogild: never mind this one, same context for the error

handler.go
```diff
@@ -0,0 +1,318 @@
...
+// isStreamingMethod returns true if RPC is streaming type
+func isStreamingMethod(serializer encoding.Serializer) bool {
+	streamSerializer, ok := serializer.(encoding.StreamSerializer)
+	if !ok {
+		return false
+	}
+	return streamSerializer.MethodType() != encoding.Unary
+}
+
+// makeStreamRequest creates a stream from the given transport and handles
```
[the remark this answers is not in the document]
jronak: Can you elaborate? this fn just creates the stream and delegates it to the correct handler based on method type

```
// makeStreamRequest creates a stream from the given transport and delegates
// it the handler based on the method's stream type
```

handler.go
```diff
@@ -0,0 +1,318 @@
...
+
+// makeServerStream starts server-side streaming rpc
+func makeServerStream(ctx context.Context, stream *yarpctransport.ClientStream,
+	reqSupplier streamRequestSupplier, resHandler streamResponseHandler) error {
+
+	req, err := reqSupplier()
+	if err != nil && err != io.EOF {
+		return err
+	}
+	if err == nil {
```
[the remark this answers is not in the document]
jronak: it's the same context for error handling

handler.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+	return makeServerStream(ctx, stream, reqSupplier, resHandler)
+}
+
+// makeServerStream starts server-side streaming rpc
+func makeServerStream(ctx context.Context, stream *yarpctransport.ClientStream,
+	reqSupplier streamRequestSupplier, resHandler streamResponseHandler) error {
+
+	req, err := reqSupplier()
+	if err != nil && err != io.EOF {
```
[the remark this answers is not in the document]
jronak: Yes, this is a server-side streaming method where client sends the initial mandatory request and server streams the response. In case user does not provide any input we send nil request, this is similar to the unary request body handling in Yab

handler.go
```diff
@@ -0,0 +1,318 @@
...
+		if err != nil {
+			break
+		}
+		err = resHandler(resBody)
+	}
+
+	if err == io.EOF {
+		return nil
+	}
+	return err
```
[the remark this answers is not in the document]
jronak: I prefer the original handling since it avoids multiple err checks within the loop, unless you insist on the refactor 🙂

handler.go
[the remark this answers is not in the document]
Dogild: Add a comment please

handler.go
jronak: same context of err handling

handler.go
jronak: Yes, both are different behaviour to input handling, io.EOF on response, etc.

| Client stream      | Server stream |
| ----------- | ----------- |
| 0 or more requests      | mandatory one request       |
| -  | error if more than one request body in input        |
| mandatory one response | 0 or more response |
| error if receive response is EOF, since one response is mandatory |  - |

Dogild: Yes the confusion was about the 0 request you explained in another comment

handler.go
jronak: We shouldn't do it, it would look confusing when streaming is used with interactive mode (STDIN) as user will see the request twice

jronak: Adding a TODO for this feature

handler.go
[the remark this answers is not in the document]
jronak: Updated the fn doc
```
// makeStreamRequest opens a stream rpc from the given transport and stream request
// it then delegates to handler based on rpc type to handle request and response of the stream
// nextBodyFn is called to get the next stream message body
// responseHandlerFn is called with the response of the stream
```

jronak:
@Dogild  Addressed the comments, going to update the PR summary with the Yab streaming request example once the sample server (lottery/proberFx) is ready to accept streaming RPCs.

Will include the example in the commit summary when I squash and merge

handler.go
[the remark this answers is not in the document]
Dogild: Sounds good to me

handler.go
```diff
@@ -0,0 +1,339 @@
...
+		if err != nil {
+			return fmt.Errorf("Failed to convert map to JSON: %v\nMap: %+v", err, res)
+		}
+
+		r.out.Printf("%s\n\n", bs)
+		return nil
+	}
+
+	if r.shouldMakeInitialRequest() {
+		if err = makeStreamRequest(r.transport, streamReq, streamSerializer, nextBodyFn, responseHandlerFn); err != nil {
```
Dogild, line 118: Why did you choose to have `func makeStreamRequest` over `func(r requestHandler) makeStreamRequest`? (you would need to pass less argument to the functions and `responseHandlerFn` and `nextBodyFn` could be attached to the receiver. 

Thoughts?

(I guess you did this because it is like this for the unary request, but this does not mean we should follow the same pattern here - or we can change it)

handler.go
```diff
@@ -0,0 +1,339 @@
...
+		resBody, receiveErr = receiveStreamMessage(ctx, stream)
+		if receiveErr != nil {
+			break
+		}
+
+		receiveErr = responseHandlerFn(resBody)
+	}
+
+	cancel()
+	wg.Wait()
```
Dogild, line 290: I dont understand this cancel management - we have 3 cancels in this function, one after the send, one after the receive and one in a defer. Can you elaborate why you added this?

Dogild: Ok got it now :)

handler.go
jronak: Kept `makeStreamRequest` separately as it is invoked in the benchmark (next PR), handler struct could be sparsely constructed but that would be confusing since we wouldn't know what all fields this method would use
https://github.com/yarpc/yab/blob/c1572df9f116de9812e4b0cd13c9c5f9fa740c53/bench_method.go#L105

Dogild: I see, we could refactor this I believe. We could give `m` to `benchmarkMethod` and `m` could implement a specific interface. 

Let's discuss offline about this if we want to do it in another diff

handler.go
```diff
@@ -0,0 +1,339 @@
...
+	if err != nil {
+		r.out.Fatalf("Failed while serializing the input: %v\n", err)
+	}
+
+	req, err = prepareRequest(req, r.headers, r.opts)
+	if err != nil {
+		r.out.Fatalf("Failed while preparing the request: %v\n", err)
+	}
+
+	if r.shouldMakeInitialRequest() {
```
Dogild, line 63: nit: can we add a comment that is needed for the warmup option

integration_test.go
```diff
@@ -368,31 +372,273 @@ func (s *simpleService) Baz(c context.Context, in *simple.Foo) (*simple.Foo, err
...
-func (s *simpleService) BidiStream(stream simple.Bar_BidiStreamServer) error {
-	return nil
+func (s *simpleService) ClientStream(stream simple.Bar_ClientStreamServer) error {
+	if s.exitOnStart {
+		return nil
+	}
+	if s.errOnStart != nil {
+		return s.errOnStart
+	}
+	counter := 0
```
Dogild, line 382: nit: missing break line

integration_test.go
```diff
@@ -368,31 +372,273 @@ func (s *simpleService) Baz(c context.Context, in *simple.Foo) (*simple.Foo, err
 	return nil, fmt.Errorf("negative input")
 }
 
-func (s *simpleService) BidiStream(stream simple.Bar_BidiStreamServer) error {
-	return nil
+func (s *simpleService) ClientStream(stream simple.Bar_ClientStreamServer) error {
```
Dogild, line 375: Let's add a comment on what is the behavior of this method. (same comment for ServerStream and BidiStream).
explaining that the return value will be equal to the number of messages received will help for maintaining the test ;)

integration_test.go
```diff
@@ -368,31 +372,273 @@ func (s *simpleService) Baz(c context.Context, in *simple.Foo) (*simple.Foo, err
...
+	}
+	return stream.SendAndClose(&simple.Foo{Test: int32(counter)})
 }
 
-func (s *simpleService) ClientStream(stream simple.Bar_ClientStreamServer) error {
+func (s *simpleService) ServerStream(req *simple.Foo, stream simple.Bar_ServerStreamServer) error {
+	if s.errOnStart != nil {
+		return s.errOnStart
+	}
+	for i := 0; i < int(req.Test); i++ {
```
Dogild, line 397: BidiStream

integration_test.go
```diff
@@ -368,31 +372,273 @@ func (s *simpleService) Baz(c context.Context, in *simple.Foo) (*simple.Foo, err
 	return nil, fmt.Errorf("negative input")
 }
 
-func (s *simpleService) BidiStream(stream simple.Bar_BidiStreamServer) error {
-	return nil
+func (s *simpleService) ClientStream(stream simple.Bar_ClientStreamServer) error {
+	if s.exitOnStart {
+		return nil
+	}
+	if s.errOnStart != nil {
```
Dogild, line 379: Missing breakline

integration_test.go
```diff
@@ -368,31 +372,273 @@ func (s *simpleService) Baz(c context.Context, in *simple.Foo) (*simple.Foo, err
...
+	}
 	return nil
 }
 
-func (s *simpleService) ServerStream(req *simple.Foo, stream simple.Bar_ServerStreamServer) error {
+func (s *simpleService) BidiStream(stream simple.Bar_BidiStreamServer) error {
+	if s.exitOnStart {
+		return nil
+	}
+	if s.errOnStart != nil {
```
Dogild, line 409: Missing breakline

integration_test.go
```diff
@@ -368,31 +372,273 @@ func (s *simpleService) Baz(c context.Context, in *simple.Foo) (*simple.Foo, err
...
+			break
+		}
+		if err != nil {
+			return err
+		}
+		msg.Test += 100
+		if err = stream.Send(msg); err != nil {
+			return err
+		}
+	}
```
Dogild, line 424: What do you think of simplifying all of this handler by having in each request message a field `expectedReturnValue`? The big benefits of having this is that for maintaining the test we will not need to understand the logic of the handler, but everything will be driven by the test itself.

This way of doing could also apply for the field `errOnStart` and `exitOnStart`.

Thoughts?

integration_test.go
```diff
@@ -368,31 +372,273 @@ func (s *simpleService) Baz(c context.Context, in *simple.Foo) (*simple.Foo, err
...
+				},
+				TOpts: TransportOptions{
+					ServiceName: "foo",
+					Peers:       []string{"grpc://" + addr.String()},
+				},
+			},
+			wantRes: ``,
+			wantErr: "Failed while reading stream input: unexpected EOF\n",
+		},
+	}
```
Dogild, line 627: Few missing unittests:
- Can we have unittests with yaml format?
- Can we add a test with bidi where the first values return is in a good json but not the second one? (like `{"test":250}{"test_wi": [1]}`

Dogild (commented):
Very good work! we are almost there \o/

integration_test.go
jronak: Replaced test service behaviour with simple expected input and output

integration_test.go
jronak: +1

Added the simple expected input and output in this test

Dogild (approved):

handler.go
```diff
@@ -0,0 +1,343 @@
...
+			return nil, fmt.Errorf("Failed while reading stream input: %v", err)
+		}
+
+		return msg, nil
+	}
+	return nextBodyFn
+}
+
+// isStreamingMethod returns true if RPC is streaming type
+func isStreamingMethod(serializer encoding.Serializer) bool {
```
Dogild, line 141: This method should have a receiver `func(r requestHandler) isStreamingRequest() bool`

handler.go
```diff
@@ -0,0 +1,343 @@
...
+
+// closeSendStream closes the stream from the client side while
+// stream can continue to receive messages from server
+func closeSendStream(ctx context.Context, stream *yarpctransport.ClientStream) error {
+	// YARPC stream.Close method internally invokes closeSend on gRPC clientStream.
+	if err := stream.Close(ctx); err != nil {
+		return fmt.Errorf("Failed to close send stream: %v", err)
+	}
+	return nil
+}
```
Dogild, line 343: Ideally we could have a files `handler.go`, `handler_unary.go` and `handler_stream.go` here, this will simplify code maintenance

handler.go
```diff
@@ -0,0 +1,343 @@
...
+		return fmt.Errorf("Failed to convert map to JSON: %v\nMap: %+v", err, res)
+	}
+
+	r.out.Printf("%s\n\n", bs)
+	return nil
+}
+
+// streamRequestSupplier returns a stream request supplier function which uses
+// provided stream request reader
+func streamRequestSupplier(streamMsgReader encoding.StreamRequestReader) streamRequestSupplierFn {
```
Dogild, line 125: Should we add a receiver to this function. I agree this is not needed yet, but it might be in the future (with option-interval). It will also make the code more consistent with the response handler

handler.go
```diff
@@ -0,0 +1,343 @@
...
+
+	streamReq.Request, err = prepareRequest(streamReq.Request, r.headers, r.opts)
+	if err != nil {
+		r.out.Fatalf("Failed while preparing the request: %v\n", err)
+	}
+
+	nextBodyFn := streamRequestSupplier(streamMsgReader)
+
+	if r.shouldMakeInitialRequest() {
+		if err = makeStreamRequest(r.transport, streamReq, r.serializer, nextBodyFn, r.responseHandler); err != nil {
```
Dogild, line 94: nit: `if err = makeStreamRequest(r.transport, streamReq, r.serializer, streamRequestSupplier(streamMsgReader), r.responseHandler); err != nil {`

handler.go
jronak: Since the methods of streaming and unary belong to `requestHandler`, I prefer keeping them in the same file. And I think handler.go isn't very huge either

Dogild (approved):

main_test.go
```diff
@@ -467,7 +467,7 @@ func TestBenchmarkLowRPSDuration(t *testing.T) {
 	start := time.Now()
 	main()
 	duration := time.Since(start)
-	assert.True(t, duration < 200*time.Millisecond, "Expected 100ms benchmark to complete within 200ms, took %v", duration)
+	assert.True(t, duration < 400*time.Millisecond, "Expected 100ms benchmark to complete within 400ms, took %v", duration)
```
AllenLuUber, line 470: can you add a comment on why do we put 400 millisecond as a threshold here?

integration_test.go
```diff
@@ -359,40 +361,386 @@ func setupYARPCServer(t *testing.T, inbound ytransport.Inbound, opts ...ythrift.
...
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			addr, server, simpleSvc := setupGRPCServer(t)
+			defer server.Stop()
+
+			simpleSvc.returnError = tt.returnError
+			simpleSvc.returnOutput = tt.returnOutput
+			simpleSvc.expectedInput = tt.expectedInput
```
AllenLuUber, line 731: optional/can refactor in another PR:
this assignment looks a bit weird, like it is has side effects on return values  in `setupGRPCServer`. Do you think the following might be more clear?

```go
func setupGRPCServer(t *testing.T, ss *simpleService) {
  if ss == nil {
    ss = &simpleService{}
  }
  ...
}
```
and here we can just do

```go
addr, server := setupGRPCServer(t, &simpleService{returnError: tt.returnError, // other 2 assignments})
```

integration_test.go
```diff
@@ -359,40 +361,386 @@ func setupYARPCServer(t *testing.T, inbound ytransport.Inbound, opts ...ythrift.
...
+					FileDescriptorSet: []string{"testdata/protobuf/simple/simple.proto.bin"},
+					Procedure:         "Bar/ClientStream",
+					Timeout:           timeMillisFlag(time.Second),
+					RequestJSON:       `{`,
+				},
+				TOpts: TransportOptions{
+					ServiceName: "foo",
+				},
+			},
+			wantRes: ``,
```
AllenLuUber, line 703: can leave as zero value (don't specify) or ""

AllenLuUber (approved):

main_test.go
jronak: This wasn't required, purged this line

PR #320, opened 2021-02-24
[gRPC Streams] Add stream-io interface 

jronak:
* Refactor existing request-supplier and response-handler functions into a separate interface for fetching stream requests and handling stream responses.

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -16,12 +16,16 @@ import (
 	"go.uber.org/zap"
 )
 
-// streamRequestSupplierFn is a function that returns the stream message body
-// if supplier has no more messages it must return io.EOF
-type streamRequestSupplierFn func() (requestBody []byte, err error)
+// StreamIO defines the interface for fetching stream requests and handling
+// stream responses.
+type StreamIO interface {
+	// NextRequest returns the next stream message body
+	// If there are no more messages it will return EOF.
+	NextRequest() (request []byte, err error)
 
-// streamResponseHandlerFn is a function that receives the stream response body
-type streamResponseHandlerFn func(responseBody []byte) error
+	// HandleResponse handles the received stream response message.
+	HandleResponse(responseBody []byte) error
+}
 
 type requestHandler struct {
 	out        output
@@ -88,8 +92,10 @@ func (r requestHandler) handleStreamRequest() {
 		r.out.Fatalf("Failed while preparing the request: %v\n", err)
 	}
 
+	streamIO := newStreamIOInitializer(r.out, r.serializer, streamMsgReader)
+
 	if r.shouldMakeInitialRequest() {
-		if err = makeStreamRequest(r.transport, streamReq, r.serializer, streamRequestSupplier(streamMsgReader), r.responseHandler); err != nil {
+		if err = makeStreamRequest(r.transport, streamReq, r.serializer, streamIO); err != nil {
 			r.out.Fatalf("%v\n", err)
 		}
 	}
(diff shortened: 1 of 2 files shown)


handler.go
```diff
@@ -342,3 +310,51 @@ func closeSendStream(ctx context.Context, stream *yarpctransport.ClientStream) e
...
+// IO methods to read requests one by one and handle responses.
+type streamIOHandler struct {
+	out             output
+	serializer      encoding.Serializer
+	streamMsgReader encoding.StreamRequestReader
+}
+
+// NextRequest returns the next stream request body from the given stream
+// message reader.
+// Note: this method must not be called once EOF has been returned.
```
Dogild, line 324: What happens if you call this method once EOF has been returned? would it panic? Please add this information.
You can also update the comment and mention the method return EOF once there is not message to read from the provided streamMsgReader

nit: you can use the style `It should be noted that calling this method once EOF was returned will....`

handler.go
```diff
@@ -342,3 +310,51 @@ func closeSendStream(ctx context.Context, stream *yarpctransport.ClientStream) e
 	}
 	return nil
 }
+
+// streamIOHandler uses provided stream message reader to provide stream
+// IO methods to read requests one by one and handle responses.
+type streamIOHandler struct {
```
Dogild, line 316: As talked offline, let's call this struct `streamIOInitializer`. This would match with other APIs using this struct (like `shouldMakeInitialRequest` for instance).

handler.go
```diff
@@ -342,3 +310,51 @@ func closeSendStream(ctx context.Context, stream *yarpctransport.ClientStream) e
 	}
 	return nil
 }
+
+// streamIOHandler uses provided stream message reader to provide stream
+// IO methods to read requests one by one and handle responses.
```
Dogild, line 315: nit: `to read` to `to get next requests and handle responses`

handler.go
```diff
@@ -342,3 +310,51 @@ func closeSendStream(ctx context.Context, stream *yarpctransport.ClientStream) e
...
+		return nil, err
+	}
+	if err != nil {
+		return nil, fmt.Errorf("Failed while reading stream input: %v", err)
+	}
+
+	return msg, nil
+}
+
+// HandleResponse validates the response bytes and prints indented JSON body.
```
Dogild, line 337: Is there any reason you have used the word `validates`? What does it mean in this context? From what I see, we get the response, we deserialize it and we print it.

Dogild (commented):
Love it!

handler.go
jronak: You are right `validates` isn't the right word, it parses the response bytes and prints it

handler.go
jronak: It actually depends on the underlying reader of streamMsgReader, if it is STDIN then trying to read again after EOF would wait for data or EOF from STDIN. To make this much simpler, we can always return EOF from `NextRequest` if streamMsgReader ever returned EOF earlier.

handler_test.go
```diff
@@ -0,0 +1,107 @@
...
+		var outBuf bytes.Buffer
+		out := testOutput{
+			Buffer: &outBuf,
+		}
+
+		body := `{
+  "test": "1"
+}
+
+`
```
Dogild, line 91: nit: remove extra break line

handler_test.go
```diff
@@ -0,0 +1,107 @@
...
+		streamIO := streamIOInitializer{out: out, serializer: encoding.NewJSON("test")}
+		require.NoError(t, streamIO.HandleResponse([]byte(body)))
+
```
Dogild, line 94: nit: remove extra break line

handler_test.go
```diff
@@ -0,0 +1,107 @@
...
+		for i, expectedReq := range requests {
+			req, err := streamIO.NextRequest()
+			require.NoError(t, err)
+			assert.Equal(t, expectedReq, req, "requests[%d] is not equal", i)
+		}
+
+		_, err := streamIO.NextRequest()
+		assert.EqualError(t, err, io.EOF.Error())
+		assert.True(t, streamIO.eofReached)
```
Dogild, line 71: Let's add one more step when you call `_, err := streamIO.NextRequest()` again so it can test the eof feature

handler_test.go
jronak: Newline is required here since it's the exact response printed by `HandleResponse` method

Dogild: Oh true, I did not pay attention to the `

Dogild (approved):
Very nice!

AllenLuUber (approved):
💯

PR #321, opened 2021-02-25
Change input-decoder interface to return YAML request body

jronak:
Prior to this change input decoder exposed method which returns JSON body, but existing `Serializer` interface already exposes methods with YAML body. This change upholds the consistency between all the interfaces and also avoids the lossy conversion from YAML to JSON type.

Fixes https://github.com/yarpc/yab/pull/314#discussion_r565766802


encoding/inputdecoder/decoder_test.go
```diff
@@ -62,13 +62,13 @@ func TestRequestReader(t *testing.T) {
 		reader, err := New(buf)
 		assert.NotNil(t, reader, "expected non-nil reader")
 		assert.NoError(t, err)
-		body, err := reader.NextJSONBytes()
+		body, err := reader.NextYAMLBytes()
```
AllenLuUber, line 65: can we change the test input here? The output `{"test":2}` looks more like JSON than YAML here. Would not expect to see `{}` in the YAML bytes.

AllenLuUber (approved):
overall LGTM. ~~Just need to update the test input to reflect the change~~

encoding/inputdecoder/decoder_test.go
jronak: JSON output is actually valid YAML (since JSON is subset of YAML).
This test has JSON requests as input, which uses JSON decoder to parse input and return back JSON bytes (which is valid YAML)

AllenLuUber: TIL! https://en.wikipedia.org/wiki/YAML: 
> ...  so that JSON files are valid YAML 1.2.

prashantv:
Does this support multiple YAML documents? E.g.,
```
---
bar: 1
---
bar: 2
```

https://www.ribbybibby.me/posts/multiple-yaml-documents/

jronak:
> Does this support multiple YAML documents? E.g.,
> 
> ```
> ---
> bar: 1
> ---
> bar: 2
> ```
> 
> https://www.ribbybibby.me/posts/multiple-yaml-documents/

Yup, `---` is used as a delimiter by yaml decoder and it can parse multiple yaml docs from the same input reader.

Linking to a gRPC stream integration test which uses multiple YAML documents -
https://github.com/yarpc/yab/blob/dace80fcc1439a7a161713a4f7b801ec738d702d/integration_test.go#L493-L500

PR #319, opened 2021-01-13
New: added gRPC streams support for benchmark - 5

jronak:
## Aim
Support gRPC streams in Yab

## Changes
* Add benchmark handling for stream requests which works with the existing benchmarking parameters
* Reuses the initial stream request dispatcher for benchmarking
* Refactor benchmark handling into different files `benchmark_unary.go` and `benchmark_stream.go` for handlings it's respective benchmark type. 
* Stream benchmark works alongside warmup request (ad-hoc request) by recording the stream requests (via STDIN or passed on CLI) sent during the warmup and reusing the same during the Benchmarks

## How to test?
Download [grpc-stream-test-server.zip](https://github.com/yarpc/yab/files/5826809/grpc-stream-test-server.zip), follow the `readme.md` to run the test server and dispatch streaming yab queries

### Bidi Stream
```
yab  --max-requests=10000 --max-duration=1s --timeout=1m -F testproto --service "test-server" --method uber.yarpc.encoding.protobuf.Test/Stream -p grpc://localhost:55555   -r '{"value": "1"}{"value": "2"}'
// Warmup response
{
  "value": "1"
}

{
  "value": "12"
}

Benchmark parameters:
  CPUs:            12
  Connections:     24
  Concurrency:     1
  Max requests:    10000
  Max duration:    1s
  Max RPS:         0
Latencies:
  0.5000: 922.41µs
  0.9000: 2.495403ms
  0.9500: 3.514855ms
  0.9900: 22.468952ms
  0.9990: 40.554458ms
  0.9995: 40.882234ms
  1.0000: 42.377369ms
Elapsed time (seconds):   0.74
Total requests:           10000
RPS:                      13558.80
```
#### STDIN mode
```
yab  --max-requests=10000 --max-duration=1s --timeout=1m -F testproto --service "test-server" --method uber.yarpc.encoding.protobuf.Test/Stream -p grpc://localhost:55555   -r '-'
{"value":"1"} // STDIN request-1
{
  "value": "1"
}

{"value":"2"} // STDIN request-2
{
  "value": "12"
}

Benchmark parameters:
  CPUs:            12
  Connections:     24
  Concurrency:     1
  Max requests:    10000
  Max duration:    1s
  Max RPS:         0
Latencies:
  0.5000: 625.964µs
  0.9000: 1.353507ms
  0.9500: 1.899145ms
  0.9900: 16.519147ms
  0.9990: 23.410829ms
  0.9995: 23.508979ms
  1.0000: 23.947096ms
[...]

PR #322, opened 2021-02-26
Refactor benchmark to accept benchmarker interface 

jronak:
Add `benchmarker` interface in benchmark and warmup methods. This change sets up clean interface to add streaming benchmarker from #319. Decouples transport warmup methods from unary benchmark which enables them to be used by streaming benchmarks as well.

### DIFF
--- a/bench_method.go
+++ b/bench_method.go
@@ -25,7 +25,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/yarpc/yab/encoding"
 	"github.com/yarpc/yab/transport"
 
 	"github.com/opentracing/opentracing-go"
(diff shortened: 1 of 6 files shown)


Dogild (approved):

AllenLuUber (approved):

bench_method.go
```diff
@@ -36,21 +35,22 @@ type peerTransport struct {
 	peerID int
 }
 
-type benchmarkMethod struct {
-	serializer encoding.Serializer
-	req        *transport.Request
+// benchmarker exposes method to dispatch requests for benchmark.
+type benchmarker interface {
+	// Call dispatches a request using the provided transport.
+	Call(transport.Transport) (time.Duration, error)
```
AllenLuUber, line 41: do you want to call method `benchmark` or `benchmarkUnary` ?

jronak: Renamed it to `benchmarkCaller`

PR #324, opened 2021-03-01
Add benchmark caller result interface

jronak:
This change intends to enable `Call` to return richer call result details apart from Latency for streaming benchmark calls such as stream messages sent and received during the benchmark call.

Follow-up PR: #325

### DIFF
--- a/bench_method.go
+++ b/bench_method.go
@@ -38,7 +38,24 @@ type peerTransport struct {
 // benchmarkCaller exposes method to dispatch requests for benchmark.
 type benchmarkCaller interface {
 	// Call dispatches a request using the provided transport.
-	Call(transport.Transport) (time.Duration, error)
+	Call(transport.Transport) (benchmarkCallResult, error)
+}
+
+type benchmarkCallResult interface {
+	// Latency returns the time taken to send request and receive response.
+	Latency() time.Duration
+}
+
+type benchmarkCallLatencyResult struct {
+	latency time.Duration
+}
+
+func (r benchmarkCallLatencyResult) Latency() time.Duration {
+	return r.latency
+}
+
+func newBenchmarkCallLatencyResult(latency time.Duration) benchmarkCallLatencyResult {
+	return benchmarkCallLatencyResult{latency}
 }
 
 // warmTransport warms up a transport and returns it. The transport is warmed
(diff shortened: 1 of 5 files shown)


bench_method.go
```diff
@@ -38,7 +38,24 @@ type peerTransport struct {
 // benchmarkCaller exposes method to dispatch requests for benchmark.
 type benchmarkCaller interface {
 	// Call dispatches a request using the provided transport.
-	Call(transport.Transport) (time.Duration, error)
+	Call(transport.Transport) (benchmarkCallResult, error)
+}
+
+type benchmarkCallResult interface {
```
Dogild, line 44: We could call that `benchmarkResulter` ? What do you think? No need of putting `Call` in the name IMO.

Also, is there a purpose to have an interface ?

Dogild, line 44: missing comment on the interface

jronak: It's the result of a single benchmark call, IMO`benchmarkResulter` wouldn't be the right name since this is not the result of entire benchmark rather of a single benchmark call. 

Currently, benchmark call returns just the Latency which is quite limiting for streaming benchmarks. #325 adds interface to return number of stream messages sent and received in streaming benchmarks

bench_method.go
```diff
@@ -38,7 +38,24 @@ type peerTransport struct {
...
+
+type benchmarkCallLatencyResult struct {
+	latency time.Duration
+}
+
+func (r benchmarkCallLatencyResult) Latency() time.Duration {
+	return r.latency
+}
+
+func newBenchmarkCallLatencyResult(latency time.Duration) benchmarkCallLatencyResult {
```
Dogild, line 57: This function should be above the `Latency` one - see https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering for function and group ordering

Dogild (approved):
I like it, please fix the nit

bench_method.go
AllenLuUber: Maybe there is use case in the future, but at least from this diff's scope, I don't see the value of having this interface. `benchmarkCallLatencyResult` seems to be enough for the use case of this change.

AllenLuUber (approved):
Thanks for making the PR self contained!

PR #325, opened 2021-03-01
Add stream messages sent and received summary in the benchmark output - gRPC Streams 6

jronak:
Follow-up from #324 
Stream benchmark now outputs the total number of stream messages send and received during the benchmark for successful calls, it is supported JSON and plaintext output modes.

Note: This must be merged after #319 has been approved

### Preview
Stream benchmark plaintext output:
```
Benchmark parameters:
  CPUs:            12
  Connections:     24
  Concurrency:     1
  Max requests:    10000
  Max duration:    1s
  Max RPS:         0
Latencies:
  0.5000: 779.01µs
  0.9000: 2.034852ms
  0.9500: 2.932846ms
  0.9900: 11.698821ms
  0.9990: 15.839751ms
  0.9995: 16.651223ms
  1.0000: 17.198644ms
Elapsed time (seconds):         0.49
Total requests:                 10000
RPS:                            20190.25
Total stream messages sent:     10000
Total stream messages received: 30000
```

Stream benchmark JSON output:
```
{
  "benchmarkParameters": {
    "cpus": 12,
    "connections": 24,
    "concurrency": 1,
    "maxRequests": 10000,
    "maxDuration": "1s",
    "maxRPS": 0
  },
  "latencies": {
    "0.5000": "748.123µs",
    "0.9000": "2.014507ms",
    "0.9500": "3.066871ms",
    "0.9900": "9.474091ms",
    "0.9990": "14.943381ms",
    "0.9995": "15.304219ms",
    "1.0000": "17.019023ms"
  },
  "summary": {
    "elapsedTimeSeconds": 0.488,
    "totalRequests": 10000,
    "rps": 20487.52
  },
  "streamSummary": {
    "totalStreamMessagesSent": 10000,
    "totalStreamMessagesReceived": 30000
  }
}
```


benchmark.go
```diff
@@ -65,11 +65,18 @@ type Summary struct {
...
+
 // BenchmarkOutput stores benchmark settings and results for JSON output
 type BenchmarkOutput struct {
-	Parameters Parameters        `json:"benchmarkParameters"`
-	Latencies  map[string]string `json:"latencies"`
-	Summary    Summary           `json:"summary"`
+	Parameters    Parameters        `json:"benchmarkParameters"`
+	Latencies     map[string]string `json:"latencies"`
+	Summary       Summary           `json:"summary"`
+	StreamSummary *StreamSummary    `json:"streamSummary,omitempty"`
```
Dogild, line 79: Can you add a comment why you use a pointer here please? I guess it is for the omit empty and to not print anything for unary request

benchmark.go
```diff
@@ -65,11 +65,18 @@ type Summary struct {
 	RPS                float64 `json:"rps"`
 }
 
+// StreamSummary stores the stream messages send and received summary
+type StreamSummary struct {
+	TotalStreamMessagesSent     int `json:"totalStreamMessagesSent,omitempty"`
+	TotalStreamMessagesReceived int `json:"totalStreamMessagesReceived,omitempty"`
```
Dogild, line 71: Why do we have omitempty for `TotalStreamMessagesSent` and `TotalStreamMessagesReceived`? when `StreamSummary` I do not see why both variables will not be used.

benchmark.go
```diff
@@ -246,23 +257,33 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
 		RPS:                rps,
 	}
 
+	var streamSummary *StreamSummary
+
+	if overall.totalStreamMessagesReceived > 0 || overall.totalStreamMessagesSent > 0 {
```
Dogild, line 262: Is there another way of checking that we are in a streaming mode?

benchmark_stream.go
```diff
@@ -85,11 +92,34 @@ func (b *streamIOBenchmark) NextRequest() ([]byte, error) {
 // HandleResponse records stream response.
 func (b *streamIOBenchmark) HandleResponse(res []byte) error {
 	b.streamResponses = append(b.streamResponses, res)
+
+	b.streamCallResult.streamMsgsReceived++
+
```
Dogild, line 97: nit: you can remove the breaklines

benchmark_stream.go
```diff
@@ -77,6 +82,8 @@ func (b *streamIOBenchmark) NextRequest() ([]byte, error) {
 	req := b.streamRequests[b.streamRequestsIdx]
 	b.streamRequestsIdx++
 
+	b.streamCallResult.streamMsgsSent++
```
Dogild, line 85: nit: you can remove the breakline

benchmark_stream.go
```diff
@@ -37,11 +39,12 @@ type benchmarkStreamMethod struct {
 
 // Call dispatches stream request on the provided transport.
 func (m benchmarkStreamMethod) Call(t transport.Transport) (benchmarkCallResult, error) {
-	streamIO := newStreamIOBenchmark(m.streamRequestMessages)
+	callResult := &streamCallResult{}
+	streamIO := newStreamIOBenchmark(m.streamRequestMessages, callResult)
```
Dogild, line 43: Is there another solution than giving callResult to `newStreamIOBenchmark`? The way callResult is updated is not consistent, one part is done in this method `Call` (with the latency) and the other part in the struct `StreamIOBenchmark`.

benchmark_stream_test.go
```diff
@@ -244,12 +250,48 @@ func TestBenchmarkStreamIO(t *testing.T) {
...
+			[]byte("1"),
+			[]byte("2"),
+			[]byte("3"),
+		}
+		expectedMsgsSent := 2
+
+		streamResult := &streamCallResult{}
+		streamIO := newStreamIOBenchmark(requests, streamResult)
+
+		for i := 0; i < expectedMsgsSent; i++ {
```
Dogild, line 273: Did you do this test on purpose? This test looks so unnatural since you loop on expectedMsgsSent and not on the length of requests. What is the value of doing this?

benchmark_stream.go
```diff
@@ -85,11 +92,34 @@ func (b *streamIOBenchmark) NextRequest() ([]byte, error) {
...
+		streamCallResult: r,
 	}
 }
+
+type streamCallResult struct {
+	latency time.Duration
+
+	streamMsgsSent     int
+	streamMsgsReceived int
+}
```
Dogild, line 113: Please be consistent, everywhere else in the code we use the entire word `messages` and not `msgs` ;)

benchmark.go
jronak: There are few cases when client or server might not send/receive any message, more like a one way communication. In such cases `TotalStreamMessagesSent` or `TotalStreamMessagesReceived` could be zero. 

Example: Client sends multiple messages and closes send stream, server processes them and can choose to end the stream without response (this is a valid scenario)

```
yab --max-duration=1s --timeout=1m -F testproto --service "test-server" --method uber.yarpc.encoding.protobuf.Test/ClientStream -p grpc://localhost:55555   -r '{}{}' --format=JSON
{}
{}

{
  "benchmarkParameters": {
    "cpus": 12,
    "connections": 24,
    "concurrency": 1,
    "maxRequests": 0,
    "maxDuration": "1s",
    "maxRPS": 0
  },
  "latencies": {
    "0.5000": "649.711µs",
    "0.9000": "1.066174ms",
    "0.9500": "1.661274ms",
    "0.9900": "16.081985ms",
    "0.9990": "23.005123ms",
    "0.9995": "24.909958ms",
    "1.0000": "26.037134ms"
  },
  "summary": {
    "elapsedTimeSeconds": 1,
    "totalRequests": 25088,
    "rps": 25077.35
  },
  "streamSummary": {
    "totalStreamMessagesSent": 50176
  }
}
```

IMO zero value fields such as send and received can be printed as well, wdyt?

benchmark.go
jronak: Yes, it is omitted in unary benchmark. Added a comment on the field

benchmark_stream.go
jronak: My bad :-)

benchmark_stream_test.go
jronak: I agree, fixed the test to use messages length

benchmark_stream.go
jronak: good catch, I fixed this by adding additional methods in streamIO to get send and received message count and callResult is created in `Call` method. This seems much more cleaner.

benchmark.go
jronak: ATM no, we have to expose `encoding` from benchmarkCaller to check if we are benchmarking stream requests.

Dogild: I am fine with this, we can add  a method `Serializer` in the interface `benchmarkCaller`. What do you think?

benchmark_stream.go
```diff
@@ -28,6 +28,8 @@ import (
 	"github.com/yarpc/yab/transport"
 )
 
+var _ benchmarkStreamCallResult = (*streamCallResult)(nil)
```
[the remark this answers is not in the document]
jronak: (Discussed offline) It is good to have an interface compliance check, but similar checks are missing for the existing interfaces. To keep it consistent, I will create a follow-up PR that adds this check for other interfaces.

bench_method.go
```diff
@@ -38,7 +38,33 @@ type peerTransport struct {
 // benchmarkCaller exposes method to dispatch requests for benchmark.
 type benchmarkCaller interface {
 	// Call dispatches a request using the provided transport.
-	Call(transport.Transport) (time.Duration, error)
+	Call(transport.Transport) (benchmarkCallResult, error)
+}
+
+// benchmarkCallResult exposes method to access benchmark call latency.
+type benchmarkCallResult interface {
```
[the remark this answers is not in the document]
jronak: I like this naming better, I have renamed the interfaces and structs. Thanks!

benchmark.go
jronak: I added a method in `benchmarkCaller` to expose the type of underlying call (unary or streaming)

benchmark.go
```diff
@@ -111,15 +122,19 @@ func (o BenchmarkOptions) enabled() bool {
...
 			s.recordError(err)
 			// TODO: Add information about which peer specifically failed.
 			logger.Info("Failed while making call.", zap.Error(err))
 			continue
 		}
 
-		s.recordLatency(latency)
+		s.recordLatency(callReport.Latency())
+
+		if streamCallReport, ok := callReport.(benchmarkStreamCallReporter); ok {
```
Dogild, line 135: You can use `b.CallMethodType()` here now

Dogild (approved):
lgtm

bench_method.go
```diff
@@ -38,18 +39,37 @@ type peerTransport struct {
...
 	// Call dispatches a request using the provided transport.
-	Call(transport.Transport) (benchmarkCallLatencyResult, error)
+	Call(transport.Transport) (benchmarkCallReporter, error)
+
+	// CallMethodType returns the type of the RPC method invoked by `Call`.
+	CallMethodType() encoding.MethodType
+}
+
+// benchmarkCallReporter exposes method to access benchmark call report like latency.
+type benchmarkCallReporter interface {
```
AllenLuUber, line 49: love the naming!

AllenLuUber (approved):

PR #327, opened 2021-03-09
Add `stream-interval` stream request option

jronak:
Stream interval option ensures consecutive stream message sends always have a gap of at-least `stream-interval` specified time duration, basically stream requests are dispatched slowly. This is useful in stream benchmarking/testing where users want to observe stream behaviour under real world conditions where each message send would have some gap between them. This works with ad-hoc and benchmark streaming requests.

```
yab --timeout=1m --service "test-server" --method uber.yarpc.encoding.protobuf.Test/Stream -p grpc://localhost:55555 -r '{"value":"req-1"}{"value":"req-2"}' --stream-interval=500ms

{
  "value": "req-1"
}

// Wait for `stream-interval`=500ms

{
  "value": "req-1req-2"
}
```

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -436,5 +436,4 @@ func (s *intervalWaiter) wait(ctx context.Context) {
 	}
 
 	s.lastAllowed = time.Now()
-	return
 }
--- a/handler.go
+++ b/handler.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"io/ioutil"
 	"sync"
+	"time"
 
 	"github.com/uber/tchannel-go"
 	"github.com/yarpc/yab/encoding"
@@ -95,7 +96,7 @@ func (r requestHandler) handleStreamRequest() {
 	streamIO := newStreamIOInitializer(r.out, r.serializer, streamMsgReader)
 
 	if r.shouldMakeInitialRequest() {
-		if err = makeStreamRequest(r.transport, streamReq, r.serializer, streamIO); err != nil {
+		if err = makeStreamRequest(r.transport, streamReq, r.serializer, streamIO, r.opts.ROpts.StreamRequestOptions); err != nil {
 			r.out.Fatalf("%v\n", err)
 		}
 	}
@@ -113,6 +114,7 @@ func (r requestHandler) handleStreamRequest() {
 		serializer:            r.serializer,
 		streamRequest:         streamReq,
 		streamRequestMessages: streamRequests,
+		opts:                  r.opts.ROpts.StreamRequestOptions,
 	})
 }
 
@@ -130,7 +132,7 @@ func (r requestHandler) isStreamingMethod() bool {
 // it then delegates to handler based on rpc type to handle request and response of the stream
 // nextBodyFn is called to get the next stream message body
 // responseHandlerFn is called with the response of the stream
-func makeStreamRequest(t transport.Transport, streamReq *transport.StreamRequest, serializer encoding.Serializer, streamIO StreamIO) error {
+func makeStreamRequest(t transport.Transport, streamReq *transport.StreamRequest, serializer encoding.Serializer, streamIO StreamIO, opts StreamRequestOptions) error {
 	streamTransport, ok := t.(transport.StreamTransport)
 	if !ok {
 		return fmt.Errorf("Transport does not support stream calls: %q", t.Protocol())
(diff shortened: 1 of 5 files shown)


handler_test.go
```diff
@@ -127,3 +129,29 @@ func TestStreamRequestRecorder(t *testing.T) {
 		assert.Equal(t, requests, gotRequests)
 	})
 }
+
+func TestIntervalWaiter(t *testing.T) {
+	t.Run("initial call must not wait", func(t *testing.T) {
+		waiter := newIntervalWaiter(time.Millisecond * 100)
```
AllenLuUber, line 135: can you move this `time.Millisecond * 100` to a top level var?
```
var duration = time.Millisecond * 100
t.Run("initiial..
```

handler_test.go
```diff
@@ -127,3 +129,29 @@ func TestStreamRequestRecorder(t *testing.T) {
...
+		now := time.Now()
+		waiter.wait(context.Background())
+		assert.True(t, time.Now().Sub(now) < time.Millisecond)
```
AllenLuUber, line 138: assert.Less is better and add a comment:
`assert.Less(t, time.Now().Sub(now), duration, "takes more than an interval for initial wait to finish")`

handler.go
```diff
@@ -395,3 +403,38 @@ func newStreamIOInitializer(out output, serializer encoding.Serializer, streamMs
...
+	}
+
+	// wait for the remaining duration or context completion.
+	select {
+	case <-ctx.Done():
+	case <-time.After(s.interval - diff):
+	}
+
+	s.lastAllowed = time.Now()
+	return
```
AllenLuUber, line 439: no need for this empty return

handler_test.go
jronak: Tried both `Less`, but it does not work well with `Time.duration` (might require type cast to int64)
```
 --- FAIL: TestIntervalWaiter/initial_call_must_not_wait (0.00s)
	panic: interface conversion: interface {} is time.Duration, not int64
```

handler_test.go
```diff
@@ -127,3 +129,29 @@ func TestStreamRequestRecorder(t *testing.T) {
...
+		duration := time.Now().Sub(now)
+		assert.True(t, duration > (time.Millisecond*100), "Unexpected time taken on wait: %v", duration)
+		assert.True(t, duration < (time.Millisecond*200), "Unexpected time taken on wait: %v", duration)
+
+		now = time.Now()
+		waiter.wait(context.Background())
+		duration = time.Now().Sub(now)
+		assert.True(t, duration > (time.Millisecond*100), "Unexpected time taken on wait: %v", duration)
+		assert.True(t, duration < (time.Millisecond*200), "Unexpected time taken on wait: %v", duration)
+	})
```
AllenLuUber: should this be in a loop?

handler_test.go
```diff
@@ -127,3 +129,31 @@ func TestStreamRequestRecorder(t *testing.T) {
...
+		waiter.wait(context.Background())
+		duration := time.Now().Sub(now)
+		assert.True(t, duration > (duration), "Unexpected time taken on wait: %v", duration)
+		assert.True(t, duration < (duration), "Unexpected time taken on wait: %v", duration)
+
+		now = time.Now()
+		waiter.wait(context.Background())
+		duration = time.Now().Sub(now)
+		assert.True(t, duration > (duration), "Unexpected time taken on wait: %v", duration)
+		assert.True(t, duration < (duration), "Unexpected time taken on wait: %v", duration)
```
AllenLuUber, line 157: should this be in a loop?

handler_test.go
AllenLuUber: your call then. Duration is an int64 underneath anyway. It seem to be fine to cast it to int64 and use Less to compare.

handler_test.go
jronak: Yup yup, this was recent commit mistake. And I agree this makes more sense under a loop

handler_test.go
jronak: After adding test under the loop, repeated usage of time.Millisecond*100 is much lower. Adding a variable actually makes this test less readable.

options.go
```diff
@@ -70,6 +70,13 @@ type RequestOptions struct {
 		JSON     bool        `long:"json" hidden:"true"`
 		Raw      bool        `long:"raw" hidden:"true"`
 	}
+
+	StreamRequestOptions StreamRequestOptions
+}
+
+// StreamRequestOptions are stream request related options
+type StreamRequestOptions struct {
+	Interval timeMillisFlag `long:"stream-interval" description:"Interval between consecutive stream request message sends."`
```
Dogild, line 79: Can you elaborate this comment and explain how it is working with concurrency and number of connections please

jronak: updated the description
```
Interval between consecutive stream message sends, applicable separately to every stream request opened on a connection.
```

PR #328, opened 2021-03-09
Bugfix: gRPC bidirectional stream stuck on invalid input

jronak:
Bidirectional stream handler has two go-routines, one for sending requests and another one for receiving the response from the server. In case any error occurs in the `send` go-routine, the context is canceled to stop the receiving routine as both send & receive routines share the same context. But turns out context is not respected by [yarpc-go/SendMessage](https://github.com/yarpc/yarpc-go/blob/b7ba06c0dfa8658b705622e338b0ad35d14bfb78/transport/grpc/stream.go#L130) and [yarpc-go/ReceiveMessage](https://github.com/yarpc/yarpc-go/blob/b7ba06c0dfa8658b705622e338b0ad35d14bfb78/transport/grpc/stream.go#L147-L154) methods from YARPC (gRPC stream methods also do not respect context in sendMsg and recvMsg API). Instead the send and receive gRPC calls can be unblocked by canceling the context originally passed while initiating stream requests.

Fix: when an unexpected error occurs in send or receive handling, cancel the original context used during stream initiation. This closes the underlying http/2 stream, causing send and receive routines blocked on stream message APIs to unblock.

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -149,7 +149,7 @@ func makeStreamRequest(t transport.Transport, streamReq *transport.StreamRequest
 
 	switch serializer.MethodType() {
 	case encoding.BidirectionalStream:
-		return makeBidiStream(ctx, stream, streamIO)
+		return makeBidiStream(ctx, cancel, stream, streamIO)
 	case encoding.ClientStream:
 		return makeClientStream(ctx, stream, streamIO)
 	default:
@@ -220,12 +220,9 @@ func makeClientStream(ctx context.Context, stream *yarpctransport.ClientStream, 
 }
 
 // makeBidiStream starts bi-directional streaming rpc
-func makeBidiStream(ctx context.Context, stream *yarpctransport.ClientStream, streamIO StreamIO) error {
+func makeBidiStream(ctx context.Context, cancel context.CancelFunc, stream *yarpctransport.ClientStream, streamIO StreamIO) error {
 	var wg sync.WaitGroup
 	var sendErr error
-
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
 
 	wg.Add(1)
 	// Start go routine to concurrently send stream messages.
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ install:
 
 .PHONY: test
 test:
-	go test -cover -race ./...
+	go test -cover -timeout 30s -race ./...
 
 
 .PHONY: docs
@@ -28,6 +28,6 @@ docs:
 
 .PHONY: test_ci
 test_ci: install build
-	go test -race -coverprofile=cover_temp.out -coverpkg=./... ./...
+	go test -timeout 30s -race -coverprofile=cover_temp.out -coverpkg=./... ./...
 	grep -v "^github.com/yarpc/yab/testdata/*" cover_temp.out > cover.out
 	go tool cover -html=cover.out -o cover.html
--- a/integration_test.go
+++ b/integration_test.go
@@ -970,8 +970,10 @@ test: 1
 				ROpts: RequestOptions{
 					FileDescriptorSet: []string{"testdata/protobuf/simple/simple.proto.bin"},
 					Procedure:         "Bar/BidiStream",
-					Timeout:           timeMillisFlag(time.Minute),
-					RequestJSON:       `{"test_err": 1}`,
+					// Ensure request timeout is longer than go test timeout (30s)
+					// See MakeFile(test_ci,test) for go test timeout.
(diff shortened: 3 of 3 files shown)


handler.go
```diff
@@ -226,12 +226,9 @@ func makeClientStream(ctx context.Context, stream *yarpctransport.ClientStream,
 }
 
 // makeBidiStream starts bi-directional streaming rpc
-func makeBidiStream(ctx context.Context, stream *yarpctransport.ClientStream, streamIO StreamIO, opts StreamRequestOptions) error {
+func makeBidiStream(ctx context.Context, cancel context.CancelFunc, stream *yarpctransport.ClientStream, streamIO StreamIO, opts StreamRequestOptions) error {
```
Dogild, line 229: (off this commit) Thoughts; do we need to play with cancel in this function? It is very unfamiliar to control execution of grouting through cancel context calls.

I believe that cancel should only be controlled by the caller of this method (as we have for makeServerStream and makeClientStream actually). 

Here, it looks like you are using cancel to stop a for loop from a goroutine. Can we use a channel for this instead?
Maybe something like this:

```
// Start go routine to concurrently send stream messages.
sendErr := make(chan error)
receiveErr := make(chan error)
done := make(chan bool)

go func() {
	var err error
	for err == nil {
		var reqBody []byte
		reqBody, err = streamIO.NextRequest()
		if err == io.EOF {
			err = closeSendStream(ctx, stream)
			break
		}
		if err != nil {
			break
		}
		reqWaiter.wait(ctx)
		err = sendStreamMessage(ctx, stream, reqBody)
	}
	if err != nil {
		sendErr <- err
		close(sendErr)
	}
	close(done)
}()

go func () {
	var err error
	for err == nil {
		var resBody []byte
		resBody, err = receiveStreamMessage(ctx, stream)
		if err != nil {
			break
		}
		err = streamIO.HandleResponse(resBody)
	}
	
	receiveErr <- err
	close(receiveErr)
}()

select {
case err := <-receiveErr:
	if err != io.EOF {
		return err
	}
case err := <-sendErr:
	if err != io.EOF {
		return err
	}
case <- done:
	return nil
}
```
Here, if one fo the goroutine have an error, `makeBidiStream` will then return the error. If sending messages is over,  `makeBidiStream` return nil.

Let me know what you think, I might miss some context here as well.

Dogild (commented):
```
But turns out context done is not respected in sendMessage and ReceiveMessage stream methods from YARPC (gRPC stream methods also do not respect context in sendMsg and recvMsg API).
```

I don't understand this statement. Can you elaborate please? I don't understand 100% th problem here to be able to review.

handler.go
jronak: Send and Receive stream APIs both use the same underlying connection/(http-2 stream) created during the stream request initiation. Both send and receive are blocking calls which return when: 
* data has been sent or received successfully 
* error (could be stream closure as well)

The idea in this PR is to close the stream by cancelling the context sent during request initiation, which causes send/receive API call to return immediately as the underlying stream ends on context cancel.

>do we need to play with cancel in this function? It is very unfamiliar to control execution of grouting through cancel context calls.

When blocking operations in two different routines are dependent on an underlying object which is context aware, then cancelling context is the cleanest way of unblocking them. 

> If sending messages is over, makeBidiStream return nil.

Good idea, but I think this might be a bit complicated as we have to wait for both routines to gracefully exit.
We cannot return once sending is over,  number of messages to be sent and received can vary so it's not safe to assume stream is over when the all the messages have been sent. Calling `close(done)` in sender routine would cause early exit of handler and it will cause some messages to be lost from the server as caller of this handler cancels the context once returned.  Another potential issue is routine leak when either of the routine exit early as handler does not wait for both to complete. (there might be deadlock issues when one of the routine finishes it causes select to return err/nil while the other routine also will try to send on a channel which does not have a consumer).

Dogild: Agree, there are some flaws in my proposal. But nothing insurmountable. This is more an off topic of this diff (since it is a refactoring and not a bug fix). Let's sync offline on how we want to change this piece of logic (or if we finally want).

handler.go
```diff
@@ -226,12 +226,9 @@ func makeClientStream(ctx context.Context, stream *yarpctransport.ClientStream,
 }
 
 // makeBidiStream starts bi-directional streaming rpc
```
Dogild, line 228: Please add a small comment explaining why cancel must big given

Dogild (approved):
lgtm, please add a comment on why the cancel must be given

AllenLuUber (approved):
thanks for adding the comments!

PR #329, opened 2021-03-09
Accept multiple requests from yaml templates

jronak:
Yaml template can be used to send gRPC stream requests now, `requests` field accepts multiple requests.

Example template: 
```
service: foo
procedure: Bar/BidiStream
caller: bar
timeout: 4.5s
requests:
    - test: 1
    - test: 2
```

Template command:
```
yab -y test/stream.yab -p grpc://localhost:55555
{
  "value": "1"
}

{
  "value": "12"
}
```

### DIFF
--- a/template.go
+++ b/template.go
@@ -21,6 +21,7 @@
 package main
 
 import (
+	"errors"
 	"io/ioutil"
 	"net/url"
 	"path/filepath"
@@ -32,6 +33,8 @@ import (
 
 	"gopkg.in/yaml.v2"
 )
+
+var errIncorrectRequestFields = errors.New("do not set `request` and `requests` fields in the template together")
 
 type template struct {
 	Peers                 []string `yaml:"peers"`
@@ -81,6 +84,11 @@ func readYAMLFile(yamlTemplate string, templateArgs map[string]string, opts *Opt
 }
 
 func getYAMLRequestBody(t template, templateArgs map[string]string) ([]byte, error) {
+	// request and requests must not be set together
+	if t.Request != nil && t.Requests != nil {
+		return nil, errIncorrectRequestFields
+	}
+
 	if t.Request != nil {
 		req, err := templateargs.ProcessMap(t.Request, templateArgs)
 		if err != nil {
@@ -90,7 +98,7 @@ func getYAMLRequestBody(t template, templateArgs map[string]string) ([]byte, err
 		return yaml.Marshal(req)
 	}
 
-	var body string
+	var body []byte
 	for _, req := range t.Requests {
 		processedReq, err := templateargs.ProcessMap(req, templateArgs)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


template_test.go
```diff
@@ -81,6 +81,13 @@ func TestTemplate(t *testing.T) {
 	assert.True(t, opts.ROpts.ThriftDisableEnvelopes)
 }
 
+func TestTemplateRequestsField(t *testing.T) {
+	opts := newOptions()
+	mustReadYAMLFile(t, "testdata/templates/stream.yab", opts)
+
+	assert.Equal(t, "---\ntest: 1\n---\ntest: 2\n", opts.ROpts.RequestJSON)
```
Dogild, line 88: nit: Let's have the following format (easier to read) for the test

```
want = `---
test: 1
---
test: 2
```

assert.Equal(t, want, opts.ROpts.RequestJSON)

(off commit) having `opts.ROpts.RequestJSON` and not json as value is odd

template.go
```diff
@@ -79,19 +80,42 @@ func readYAMLFile(yamlTemplate string, templateArgs map[string]string, opts *Opt
...
+	if t.Request != nil {
+		req, err := templateargs.ProcessMap(t.Request, templateArgs)
+		if err != nil {
+			return nil, err
+		}
+
+		return yaml.Marshal(req)
+	}
+
+	var body string
```
Dogild, line 93: nit: you could use a string builder here (faster for string concatenation). This code is not in a hot path, so only nit

Dogild (approved):

template.go
```diff
@@ -47,11 +47,12 @@ type template struct {
...
-	Headers map[string]string           `yaml:"headers"`
-	Baggage map[string]string           `yaml:"baggage"`
-	Jaeger  bool                        `yaml:"jaeger"`
-	Request map[interface{}]interface{} `yaml:"request"`
-	Timeout time.Duration               `yaml:"timeout"`
+	Headers  map[string]string             `yaml:"headers"`
+	Baggage  map[string]string             `yaml:"baggage"`
+	Jaeger   bool                          `yaml:"jaeger"`
+	Request  map[interface{}]interface{}   `yaml:"request"`
+	Requests []map[interface{}]interface{} `yaml:"requests"`
```
prashantv, line 54: what happens if both are set? should we return an error (I think we ignore `Request` silently right now)

template.go
jronak: good one, this didn't require string use actually so changed body to `[]byte` which now uses `append`

template_test.go
jronak: Yup `RequestJSON` weirdly carries YAML as well. This needs some cleanup for sure

template.go
jronak: I missed adding this check earlier, now it returns error if both fields are set

prashantv (approved):

PR #332, opened 2021-04-01
[yab] handle byte correctly based on thrift spec

AllenLuUber:
Currently the check range for the byte is 0-256 and it is based on a Golang's byte. 
But [thrift spec](https://thrift.apache.org/docs/types) says it is a signed 8-bit integer between -128 and 127. 

Update the logic to reflect this.
```go
func main() {
	v := -8
	fmt.Printf("casted is %+v\n", byte(v))
}
// Output
casted is 248
```
Verified with a Java server that they will handle the overflown 248 correctly in Java (credits to @abhishekparwal)
```java
public class Main
{
	public static void main(String[] args) {
	    byte x = (byte)248;
		System.out.println(x);
	}
}
// Output
-8
```

### DIFF
--- a/thrift/types.go
+++ b/thrift/types.go
@@ -118,8 +118,11 @@ func parseBinaryList(vl []interface{}) ([]byte, error) {
 		case int:
 			// YAML uses int for all small values. For large values, it may use int64 or uint64
 			// but those are not valid bool values anyway, so we don't need to check them.
-			if v < 0 || v >= (1<<8) {
-				return nil, fmt.Errorf("failed to parse list of bytes: %v is not a byte", v)
+			// from https://thrift.apache.org/docs/types:
+			// byte: An 8-bit signed integer
+			// 8-bit signed ranges from -128 to 127
+			if boundary := 1 << 7; v < -boundary || v >= boundary {
+				return nil, fmt.Errorf("failed to parse list of bytes: %v is not a valid thrift byte(8-bit signed integer) between -128 and 127 inclusive", v)
 			}
 			bs = append(bs, byte(v))
 		case string:
--- a/thrift/types.go
+++ b/thrift/types.go
@@ -120,8 +120,8 @@ func parseBinaryList(vl []interface{}) ([]byte, error) {
 		// but those are not valid bool values anyway, so we don't need to check them.
 		case int:
 			// from https://thrift.apache.org/docs/types, byte is an 8-bit signed integer
-			// For Go, thriftrw uses a Go byte, which is an uint8 so it is possible to have 0 to 255 over the wire.
-			// Apache Java clients respect the signed int8 and send [-128,127].
+			// In Go, byte is an unsigned int8 (uint8) which has a range [0, 255].
+			// In Java, byte is a signed int8 which has a range of [-128,127].
 			// To make both sides happy, check between [-128,255]
 			if v < math.MinInt8 || v > math.MaxUint8 {
 				return nil, fmt.Errorf("failed to parse list of bytes: %v is not a valid thrift byte", v)
--- a/thrift/types.go
+++ b/thrift/types.go
@@ -119,9 +119,9 @@ func parseBinaryList(vl []interface{}) ([]byte, error) {
 		// but those are not valid bool values anyway, so we don't need to check them.
 		case int:
 			// from https://thrift.apache.org/docs/types, byte is an 8-bit signed integer
-			// For Go, thriftrw uses a Go byte, which is an uint8 so it is possible to have 0 to 225 over the wire.
-			// Apache Java clients respects the signed int8 and send -128 to 127.
-			// To make both sides happy, check between -128 and 225
+			// For Go, thriftrw uses a Go byte, which is an uint8 so it is possible to have 0 to 255 over the wire.
+			// Apache Java clients respect the signed int8 and send -128 to 127.
+			// To make both sides happy, check between -128 and 255
 			if v < -(1<<7) || v >= (1<<8) {
 				return nil, fmt.Errorf("failed to parse list of bytes: %v is not a valid thrift byte", v)
 			}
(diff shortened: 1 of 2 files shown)


thrift/types.go
```diff
@@ -118,8 +118,11 @@ func parseBinaryList(vl []interface{}) ([]byte, error) {
 		case int:
 			// YAML uses int for all small values. For large values, it may use int64 or uint64
 			// but those are not valid bool values anyway, so we don't need to check them.
-			if v < 0 || v >= (1<<8) {
-				return nil, fmt.Errorf("failed to parse list of bytes: %v is not a byte", v)
+			// from https://thrift.apache.org/docs/types:
+			// byte: An 8-bit signed integer
+			// 8-bit signed ranges from -128 to 127
+			if boundary := 1 << 7; v < -boundary || v >= boundary {
```
abhishekparwal, line 124: though this is strictly the right thing to do for the original thrift specs which are implemented by apache thrift lib (in java, and other langs): https://github.com/apache/thrift. 
however since the issue lies with thriftrw, so if anyone uses bytes serialized by thriftrw which is in the range 0 to 255, it would fail there. Thus, recommend having this check between -128 to 255 to maintain backward compatibility, and since we are not planning to fix it in thriftrw-go. This will keep both apache thrift and thriftrw-go clients happy!

AllenLuUber: yup this makes sense

thrift/types.go
```diff
@@ -115,11 +115,15 @@ func parseBinaryList(vl []interface{}) ([]byte, error) {
...
+		// but those are not valid bool values anyway, so we don't need to check them.
 		case int:
-			// YAML uses int for all small values. For large values, it may use int64 or uint64
-			// but those are not valid bool values anyway, so we don't need to check them.
-			if v < 0 || v >= (1<<8) {
-				return nil, fmt.Errorf("failed to parse list of bytes: %v is not a byte", v)
+			// from https://thrift.apache.org/docs/types, byte is an 8-bit signed integer
+			// For Go, thriftrw uses a Go byte, which is an uint8 so it is possible to have 0 to 225 over the wire.
+			// Apache Java clients respects the signed int8 and send -128 to 127.
+			// To make both sides happy, check between -128 and 225
```
abhishekparwal, line 124: typo: 255

thrift/types.go
```diff
@@ -115,11 +115,15 @@ func parseBinaryList(vl []interface{}) ([]byte, error) {
...
+			if v < -(1<<7) || v >= (1<<8) {
```
gopalchouhan, line 125: can we use math.MinInt8 and math.MaxUint8 for min and max values? I believe that is more cleaner

abhishekparwal (approved):

thrift/types.go
abhishekparwal: given we are talking in explicit number range, so I think we should avoid indirection by using math util. being more explicit here IMO is cleaner.

thrift/types.go
```diff
@@ -115,11 +115,15 @@ func parseBinaryList(vl []interface{}) ([]byte, error) {
...
+		// but those are not valid bool values anyway, so we don't need to check them.
 		case int:
-			// YAML uses int for all small values. For large values, it may use int64 or uint64
-			// but those are not valid bool values anyway, so we don't need to check them.
-			if v < 0 || v >= (1<<8) {
-				return nil, fmt.Errorf("failed to parse list of bytes: %v is not a byte", v)
+			// from https://thrift.apache.org/docs/types, byte is an 8-bit signed integer
+			// For Go, thriftrw uses a Go byte, which is an uint8 so it is possible to have 0 to 255 over the wire.
+			// Apache Java clients respect the signed int8 and send -128 to 127.
+			// To make both sides happy, check between -128 and 255
```
prashantv, line 124:
```suggestion
			// To make both sides happy, allow [-128, 255]
```
nit: use the inclusive brackets to make it clear that the border values are allowed

thrift/types.go
prashantv: +1

thrift/types_test.go
```diff
@@ -292,6 +292,10 @@ func TestParseBinary(t *testing.T) {
 			value:  []interface{}{256},
 			errMsg: "failed to parse list of bytes",
 		},
+		{
+			value:  []interface{}{-129},
+			errMsg: "failed to parse list of bytes",
+		},
```
prashantv, line 298: Can we add a couple of test cases:
 * values in the < 0 range that are allowed, including the borderline value of 128
 * values in the 128-255 range that are allowed, including the border line value of 255

Ideally, the values in the 128-255 range overlap with the negative values used, so we can show that -1 and 255 are the same in the test

thrift/types.go
```diff
@@ -115,11 +116,15 @@ func parseBinaryList(vl []interface{}) ([]byte, error) {
...
+		// YAML uses int for all small values. For large values, it may use int64 or uint64
+		// but those are not valid bool values anyway, so we don't need to check them.
 		case int:
-			// YAML uses int for all small values. For large values, it may use int64 or uint64
-			// but those are not valid bool values anyway, so we don't need to check them.
-			if v < 0 || v >= (1<<8) {
-				return nil, fmt.Errorf("failed to parse list of bytes: %v is not a byte", v)
+			// from https://thrift.apache.org/docs/types, byte is an 8-bit signed integer
+			// For Go, thriftrw uses a Go byte, which is an uint8 so it is possible to have 0 to 255 over the wire.
+			// Apache Java clients respect the signed int8 and send [-128,127].
```
prashantv, line 124: let's remove "over the wire" since it's not that the numbers sent on the wire are different. over the wire, a byte is always the same -- it's just semantic meaning on the byte that languages put on top of the byte that results in negative values vs values > 128.
```suggestion
			// In Go, byte is an unsigned int8 (uint8) which has a range [0, 255].
			// In Java, byte is a signed int8 which has a range of [-128,127].
```

prashantv (approved):

gopalchouhan (approved):

PR #336, opened 2021-04-21
Delay client send stream closure

jronak:
Adds option `delay-close-send-stream` which accepts time duration to delay the closure of the send stream in the client streaming, and bidirectional streaming methods. 

Currently, Yab closes the send stream once all the client requests have been dispatched which causes certain bidirectional streaming servers to stop the stream entirely. With this option, once all the client requests have been dispatched we can delay the closure of send stream. 

This is particularly useful for Adhoc requests where servers stop sending the messages once send stream is closed. 

For example, the client opens a bidirectional stream to watch the user activity of two users. But the server closes the stream if the client-side send stream is stopped.

Earlier:
```zsh
yab --timeout=1m  --service "test-service" --method uber.yarpc.encoding.protobuf.Test/Stream -p grpc://localhost:5555   -r '{"watch": "user-1"} {"watch": "user-2"}'

// server doesn't send any message as the client has ended the send stream after sending the above two messages.
```

Now:
```zsh
yab --timeout=1m  --service "test-service" --method uber.yarpc.encoding.protobuf.Test/Stream -p grpc://localhost:5555   -r '{"watch": "user-1"} {"watch": "user-2"}' --delay-close-send-stream=1m

{"user": "user-1", "activity": "LOGIN"}
.
.
.
// server continues to stream responses until send stream is open
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Request Options:
       --stream-interval=         Interval between consecutive stream message sends,
                                  applicable separately to every stream request
                                  opened on a connection.
-      --delay-close-send-stream= Delay the closure of send stream once all the
+      --stream-delay-close-send= Delay the closure of send stream once all the
                                  request messages have been sent.
 
 Transport Options:
--- a/README.md
+++ b/README.md
@@ -73,7 +73,11 @@ Request Options:
       --multiplexed-thrift       Enables the Thrift TMultiplexedProtocol used
                                  by services that host multiple Thrift services
                                  on a single endpoint.
-      --stream-interval=         Interval between consecutive stream message sends, applicable separately to every stream request opened on a connection.
+      --stream-interval=         Interval between consecutive stream message sends,
+                                 applicable separately to every stream request
+                                 opened on a connection.
+      --delay-close-send-stream= Delay the closure of send stream once all the
+                                 request messages have been sent.
 
 Transport Options:
   -s, --service=                 The TChannel/Hyperbahn service name
--- a/handler.go
+++ b/handler.go
@@ -212,7 +212,7 @@ func makeClientStream(ctx context.Context, stream *yarpctransport.ClientStream, 
 	}
 
 	if err == io.EOF {
-		err = closeSendStream(ctx, stream)
+		err = closeSendStream(ctx, stream, opts.DelayCloseSendStream.Duration())
 	}
 	if err != nil {
 		return err
@@ -241,7 +241,7 @@ func makeBidiStream(ctx context.Context, cancel context.CancelFunc, stream *yarp
 			var reqBody []byte
 			reqBody, err = streamIO.NextRequest()
 			if err == io.EOF {
-				err = closeSendStream(ctx, stream)
+				err = closeSendStream(ctx, stream, opts.DelayCloseSendStream.Duration())
 				break
 			}
 			if err != nil {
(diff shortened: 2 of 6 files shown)


integration_test.go
```diff
@@ -1001,72 +1001,113 @@ test: 1
...
-			ROpts: RequestOptions{
-				FileDescriptorSet: []string{"testdata/protobuf/simple/simple.proto.bin"},
-				Procedure:         "Bar/ClientStream",
-				Timeout:           timeMillisFlag(time.Second * 5),
-				RequestJSON:       `{"test":1}{"test":2}{"test":-1}`,
-				StreamRequestOptions: StreamRequestOptions{
-					Interval: timeMillisFlag(time.Millisecond * 1000),
+			// Test must run for more than 2 seconds as there are three requests and
+			// it must take 2 seconds totally between consecutive messages.
+			wantExecTime: time.Second * 2,
```
prashantv, line 1030: here and below, since tests will wait for the timeout, should we use shorter values?

(ideally we wouldn't have tests sleeping at all, but I think that'll need a larger refactor, so something to note for future)

handler.go
```diff
@@ -317,8 +317,16 @@ func receiveStreamMessage(ctx context.Context, stream *yarpctransport.ClientStre
...
 // closeSendStream closes the stream from the client side while
-// stream can continue to receive messages from server
-func closeSendStream(ctx context.Context, stream *yarpctransport.ClientStream) error {
+// stream can continue to receive messages from server. If non-zero delay is
+// passed, stream is closed after the delay.
+func closeSendStream(ctx context.Context, stream *yarpctransport.ClientStream, delayCloseSendStream time.Duration) error {
+	if delayCloseSendStream != 0 {
+		select {
+		case <-time.After(delayCloseSendStream):
+		case <-ctx.Done():
```
prashantv, line 326: can we cover this case in a test?

prashantv (approved):

README.md
```diff
@@ -73,7 +73,11 @@ Request Options:
       --multiplexed-thrift       Enables the Thrift TMultiplexedProtocol used
                                  by services that host multiple Thrift services
                                  on a single endpoint.
-      --stream-interval=         Interval between consecutive stream message sends, applicable separately to every stream request opened on a connection.
+      --stream-interval=         Interval between consecutive stream message sends,
+                                 applicable separately to every stream request
+                                 opened on a connection.
+      --delay-close-send-stream= Delay the closure of send stream once all the
```
Dogild, line 79: Can we call this `--stream-delay-close-send` ? I would prefer to keep having the prefix --stream for all options regarding streaming.

options.go
```diff
@@ -76,7 +76,8 @@ type RequestOptions struct {
 
 // StreamRequestOptions are stream request related options
 type StreamRequestOptions struct {
-	Interval timeMillisFlag `long:"stream-interval" description:"Interval between consecutive stream message sends, applicable separately to every stream request opened on a connection."`
+	Interval             timeMillisFlag `long:"stream-interval" description:"Interval between consecutive stream message sends, applicable separately to every stream request opened on a connection."`
+	DelayCloseSendStream timeMillisFlag `long:"delay-close-send-stream" description:"Delay the closure of send stream once all the request messages have been sent."`
```
Dogild, line 80: We should add a validation with this option that it is not used with the stdin option (they are not compatible right?)

jronak: It's good to keep this option available for interactive mode as well. This enables users to pipe a file to the `stdin` where the file reads EOF in the end, stream closure due to stdin EOF can be delayed with this option.

README.md
jronak: makes sense, done.

integration_test.go
jronak: I have reduced the time to ~100ms for all the test cases. I agree, I'll try to refactor this with a mock clock setup.

handler.go
jronak: Added a unit test, also covered in an integration test.

integration_test.go
```diff
@@ -1001,72 +1001,113 @@ test: 1
...
-		opts := Options{
-			ROpts: RequestOptions{
-				FileDescriptorSet: []string{"testdata/protobuf/simple/simple.proto.bin"},
-				Procedure:         "Bar/ClientStream",
-				Timeout:           timeMillisFlag(time.Second * 5),
-				RequestJSON:       `{"test":1}{"test":2}{"test":-1}`,
-				StreamRequestOptions: StreamRequestOptions{
-					Interval: timeMillisFlag(time.Millisecond * 1000),
+			// Test must run for more than 2 seconds as there are three requests and
+			// it must take 2 seconds totally between consecutive messages.
```
Dogild, line 1029: nit: this comment is not up to date anymore

Dogild (approved):

Issue #335, opened 2021-04-20
Stalled gRPC server can result in reflection calls hanging (past specified timeout)


prashantv:
Goroutine when yab is hung making a gRPC call to a "stuck" Java server,
```
goroutine 1 [select]:
google.golang.org/grpc/internal/transport.(*Stream).waitOnHeader(0xc000040400)
        external/org_golang_google_grpc/internal/transport/transport.go:318 +0xd5
google.golang.org/grpc/internal/transport.(*Stream).RecvCompress(...)
        external/org_golang_google_grpc/internal/transport/transport.go:333
google.golang.org/grpc.(*csAttempt).recvMsg(0xc000160280, 0xe9aca0, 0xc000150320, 0x0, 0xc000038200, 0x5)
        external/org_golang_google_grpc/stream.go:870 +0x725
google.golang.org/grpc.(*clientStream).RecvMsg.func1(0xc000160280, 0x203000, 0x203000)
        external/org_golang_google_grpc/stream.go:735 +0x46
google.golang.org/grpc.(*clientStream).withRetry(0xc0003bc7e0, 0xc0003a8bd8, 0xc0003a8ba8, 0xc000038200, 0x7fb3abd921a0)
        external/org_golang_google_grpc/stream.go:593 +0x9c
google.golang.org/grpc.(*clientStream).RecvMsg(0xc0003bc7e0, 0xe9aca0, 0xc000150320, 0xc0000e80c0, 0x0)
        external/org_golang_google_grpc/stream.go:734 +0x105
google.golang.org/grpc/reflection/grpc_reflection_v1alpha.(*serverReflectionServerReflectionInfoClient).Recv(0xc0000ac190, 0xc0000e80c0, 0x0, 0x0)
        external/org_golang_google_grpc/reflection/grpc_reflection_v1alpha/reflection.pb.go:686 +0x62
github.com/jhump/protoreflect/grpcreflect.(*Client).doSendLocked(0xc000160200, 0x1, 0xc0000e80c0, 0xc0003a8d28, 0x4149e5, 0xc0000f6418)
        external/com_github_jhump_protoreflect/grpcreflect/client.go:469 +0x1a2
github.com/jhump/protoreflect/grpcreflect.(*Client).doSend(0xc000160200, 0xea3801, 0xc0000e80c0, 0x0, 0x0, 0x0)
        external/com_github_jhump_protoreflect/grpcreflect/client.go:449 +0x90
[...]

PR #337, opened 2021-04-21
Fix protobuf reflection hung calls

jronak:
I have added a test to confirm the issue of the hung call in protobuf reflection methods. This change fixes it by adding a context deadline around the Protobuf reflection gRPC stream call. 

Fixes: #335

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -1136,15 +1136,15 @@ func TestGRPCReflectionSource(t *testing.T) {
 }
 
 type protoReflectService struct {
-	sleepDuration time.Duration // If non-zero, handler sleeps for the given duration.
+	waitChan chan struct{} // If non-nil, handler waits on the channel
 }
 
 func (p protoReflectService) ServerReflectionInfo(s rpb.ServerReflection_ServerReflectionInfoServer) error {
-	if p.sleepDuration != 0 {
-		// Sleep until the given duration or until stream context completes.
+	if p.waitChan != nil {
+		// Wait on the given channel or until stream context completes.
 		select {
 		case <-s.Context().Done():
-		case <-time.After(p.sleepDuration):
+		case <-p.waitChan:
 		}
 	}
 	return nil
@@ -1154,14 +1154,14 @@ func TestGPCReflectionTimeout(t *testing.T) {
 	ln, err := net.Listen("tcp", "127.0.0.1:0")
 	require.NoError(t, err)
 
-	// Sleep duration must be longer than go test timeout (30s) to
-	// catch client which does not respect timeouts in reflection calls.
-	// See MakeFile(test_ci,test) for go test timeout.
-	reflectSvc := protoReflectService{sleepDuration: time.Second * 31}
-	s := grpc.NewServer()
-	rpb.RegisterServerReflectionServer(s, reflectSvc)
-	go s.Serve(ln)
-	defer s.Stop()
+	waitChan := make(chan struct{})
+	reflectSvc := protoReflectService{waitChan: waitChan}
+	server := grpc.NewServer()
+	rpb.RegisterServerReflectionServer(server, reflectSvc)
+	go server.Serve(ln)
+
+	defer server.Stop()
+	defer close(waitChan)
 
 	gotOut, gotErr := runTestWithOpts(Options{
 		ROpts: RequestOptions{
--- a/integration_test.go
+++ b/integration_test.go
@@ -55,6 +55,7 @@ import (
 	ytchan "go.uber.org/yarpc/transport/tchannel"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/reflection"
+	rpb "google.golang.org/grpc/reflection/grpc_reflection_v1alpha"
 )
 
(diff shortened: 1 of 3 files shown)


protobuf/source_reflection.go
```diff
@@ -95,7 +98,7 @@ func (s *grpcreflectSource) FindService(fullyQualifiedName string) (*desc.Servic
 		}
 
 		if !grpcreflect.IsElementNotFoundError(err) {
-			return nil, err
+			return nil, fmt.Errorf("error in protobuf reflection: %v", err)
```
prashantv, line 101: Couple of things to check:
 * Should we add this same wrapper for all errors that are unexpected (e.g., in `FindMessage` as well)
 * slightly related to this change, but it's an existing issue: should we check `IsElementNotFound` before we even do `ListServices`?

integration_test.go
```diff
@@ -1133,3 +1134,46 @@ func TestGRPCReflectionSource(t *testing.T) {
...
+}
+
+func TestGPCReflectionTimeout(t *testing.T) {
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+
+	// Sleep duration must be longer than go test timeout (30s) to
+	// catch client which does not respect timeouts in reflection calls.
+	// See MakeFile(test_ci,test) for go test timeout.
+	reflectSvc := protoReflectService{sleepDuration: time.Second * 31}
```
prashantv, line 1160: Instead of relying on the test timeout (which may change), we can instead have the `protoReflectService` block on a context until the test is complete. So instead of a `sleepDuration`, we can pass a channel or a context.

prashantv (commented):
Thanks for fixing this so quickly! Left a couple of small comments

integration_test.go
jronak: Thanks! moved this to use a channel 👍

protobuf/source_reflection.go
jronak: Wrapped the errors in `FindMessage` and `ListServices` with the same message. Ordering of `ListServices` was above to avoid checking `IsElementNotFound` twice, I have reordered them with an additional `IsElementNotFound` check for `ListServices`

prashantv (approved):
Thanks for fixing this!

protobuf/source_reflection.go
```diff
@@ -64,14 +64,17 @@ func NewDescriptorProviderReflection(args ReflectionArgs) (DescriptorProvider, e
...
-		client: grpcreflect.NewClient(metadataContext, pbClient),
+		client:    grpcreflect.NewClient(metadataContext, pbClient),
+		ctxCancel: cancel,
 	}, nil
 }
 
 type grpcreflectSource struct {
-	client *grpcreflect.Client
+	client    *grpcreflect.Client
+	ctxCancel context.CancelFunc
```
Dogild, line 77: nit: you can call this `cancelFunc`

protobuf/source_reflection.go
```diff
@@ -83,19 +86,23 @@ func (s *grpcreflectSource) FindMessage(messageType string) (*desc.MessageDescri
 		return nil, nil
 	}
 
+	if err != nil {
+		return nil, wrapReflectionError(err)
```
Dogild, line 90: missing unitests

protobuf/source_reflection.go
```diff
@@ -83,19 +86,23 @@ func (s *grpcreflectSource) FindMessage(messageType string) (*desc.MessageDescri
...
-			err = availableErr
+		if !grpcreflect.IsElementNotFoundError(err) {
+			return nil, wrapReflectionError(err)
 		}
 
-		if !grpcreflect.IsElementNotFoundError(err) {
-			return nil, err
+		available, availableErr := s.client.ListServices()
+		if availableErr != nil && !grpcreflect.IsElementNotFoundError(availableErr) {
+			return nil, wrapReflectionError(availableErr)
```
Dogild, line 105: missing unitests

protobuf/source_reflection.go
```diff
@@ -111,5 +118,10 @@ func (s *grpcreflectSource) FindService(fullyQualifiedName string) (*desc.Servic
 }
 
 func (s *grpcreflectSource) Close() {
+	s.ctxCancel()
 	s.client.Reset()
 }
+
+func wrapReflectionError(err error) error {
+	return fmt.Errorf("error in protobuf reflection: %v", err)
```
Dogild, line 126: nit: you can use `fmt.Errorf("error in protobuf reflection: %s", err.Error())` (faster, less alloc)

Dogild (approved):

protobuf/source_reflection.go
jronak: For consistency in the codebase, going to stick to `%v` (this method won't impact benchmark or call performance)

protobuf/source_reflection.go
jronak: Catching this one is a bit tricky as the grpcReflect client doesn't expose an interface. Also, this a [minor rejig](https://github.com/yarpc/yab/pull/337#discussion_r617695948) of statements from before and this change doesn't reduce the coverage around here. Needs more refactor, which can be done outside of this PR.

Dogild (approved):

PR #338, opened 2021-04-29
Add Encoding & File to YAML template

ajwans:


Dogild:
Thanks for the contribution, is there any chance to update the summary of the PR to get more context on this please?

ajwans:
I updated the top comment, is that what you need?

template.go
```diff
@@ -160,6 +170,17 @@ func readYAMLRequest(base string, contents []byte, templateArgs map[string]strin
 		opts.ROpts.ThriftFile = thriftFileURL.Path
 	}
 
+	switch strings.ToLower(t.Encoding) {
+	case "thrift":
+		opts.ROpts.Encoding = encoding.Thrift
+	case "json":
+		opts.ROpts.Encoding = encoding.JSON
+	case "protobuf":
+		opts.ROpts.Encoding = encoding.Protobuf
```
Dogild, line 179: When using the option -e, --encoding, encoding.Protobuf will match with `proto` and not `protobuf`.

template.go
```diff
@@ -129,6 +132,13 @@ func readYAMLRequest(base string, contents []byte, templateArgs map[string]strin
 		return err
 	}
 
+	if t.File != "" {
+		body, err = ioutil.ReadFile(t.File)
+		if err != nil {
+			return err
+		}
+	}
+
```
Dogild, line 141: We should prevent users of using `file` and `request` together. Right now, File have priority over request, only one should be available

template.go
```diff
@@ -160,6 +170,17 @@ func readYAMLRequest(base string, contents []byte, templateArgs map[string]strin
...
+	case "thrift":
+		opts.ROpts.Encoding = encoding.Thrift
+	case "json":
+		opts.ROpts.Encoding = encoding.JSON
+	case "protobuf":
+		opts.ROpts.Encoding = encoding.Protobuf
+	case "raw":
+		opts.ROpts.Encoding = encoding.Raw
+	}
+
```
Dogild, line 183: Line `169`, there is the following line of code `overrideParam(&opts.ROpts.RequestJSON, string(body))`.  How would it work when the body given in file has another format than JSON? Right now, body could be with format proto, thrift or raw.

template_test.go
```diff
@@ -391,3 +391,9 @@ func TestReadYAMLRequestFails(t *testing.T) {
 		})
 	}
 }
+
+func TestReadYAMLSucceeds(t *testing.T) {
+	opts := newOptions()
+	err := readYAMLFile("testdata/valid.yab", nil, opts)
+	assert.NoError(t, err)
+}
```
Dogild, line 399: Can we add more test please? We should provide the following unit tests:
- Yab template with file + JSON format
- Yab template with file + YAML format
- Yab template with file + RAW format
- Yab template with file + Proto format
- Yab template with file + Thrift format
- Negative test with bad combinaison of options (like File and request together)

Dogild (commented):
Please update the summary of the diff with examples of this new feature. Can you also please update the Changelog with this new feature

PR #345, opened 2021-07-21
Fixed: Maintain protobuf reflection connection until end of request

jronak:
This change avoids early closure of protobuf reflection connection, allowing correct parsing of protobuf messages containing `Anytype` fields in the response or error details.

Before:
```
> yab test-service test_service.Test/Unary -r '{}'
Failed while parsing response: unknown message type "test_service.Test.FooAny"
```

After:
```
> yab test-service test_service.Test/Unary -r '{}'
{
  "body": {
    "any" : {
      "value": 123
    }
  }
}
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -329,6 +329,9 @@ func runWithOptions(opts Options, out output, logger *zap.Logger) {
 	if err != nil {
 		out.Fatalf("Failed while parsing input: %v\n", err)
 	}
+	if serializerWithClose, ok := serializer.(io.Closer); ok {
+		defer serializerWithClose.Close()
+	}
 
 	tracer, closer := getTracer(opts, out)
 	if closer != nil {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@ Changelog
 =========
 
 # Unreleased (0.21.0)
-* Nothing yet
+* Fix parsing of protobuf responses/error-details containing Any type fields, by maintaining
+  reflection server connection until end of the request.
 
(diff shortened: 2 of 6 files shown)


main.go
```diff
@@ -329,6 +329,9 @@ func runWithOptions(opts Options, out output, logger *zap.Logger) {
 	if err != nil {
 		out.Fatalf("Failed while parsing input: %v\n", err)
 	}
+	if serializerWithClose, ok := serializer.(io.Closer); ok {
```
Dogild, line 332: Did you consider using the interface `protobuf.DescriptorProvider` instead of `io.Closer` ?

Dogild (approved):

main.go
```diff
@@ -329,6 +329,9 @@ func runWithOptions(opts Options, out output, logger *zap.Logger) {
...
+		defer serializerWithClose.Close()
+	}
```
prashantv, line 334: It looks a little odd to create a serializer, and close it without doing anything with it. Should we rename the method so it's not `Close()`?

jronak: Serializer is (deferred) closed after the benchmark, or ad-hoc request handling is complete, `io.Closer` looked cleaner to me as serializer held protobuf descriptor fd (net/file) resource which must be closed after usage. Do you think it must be named something else like a `stop`, `stopDescriptor`, `closeDescriptor`, etc.?

prashantv: Misread the `defer`, so it looked odd. Given the defer, this makes sense to me.

prashantv (approved):

PR #346, opened 2021-07-21
Fixed: patch gRPC server handler to close send stream post request dispatch

jronak:
This change is a compatibility fix with the Java gRPC server specifically for server-side streaming methods.

In server-side streaming requests, Java servers wait until the client sends the first request and closes the send stream before forwarding the request to the service handler. This behavior is different in Go gRPC servers, where it receives the request and immediately invokes the handler without waiting for the client's send stream to close.

With this change for a server-side streaming method, we dispatch the request and immediately close the send stream. (does not have any side-effect on the Go side).

Earlier:
```
> yab --peer grpc://localhost:8087 --timeout 10s --service stream-java-server -H  --procedure example.java.service/GetMulti -r '{"topic_name": "test"}'
Failed while receiving stream response: code:deadline-exceeded message:context deadline exceeded
```

After:
```
> yab --peer grpc://localhost:8087 --timeout 10s --service stream-java-server -H  --procedure example.java.service/GetMulti -r '{"topic_name": "test"}'

{
  "topic_name": "test_1"
}
{
  "topic_name": "test_2",
  "status": true
}
```

Go server stream with this change works as expected:
```
yab -F hellostream --procedure routeguide.RouteGuide/ListFeatures --service helloworld -p grpc://localhost:10000 -r '{"lo":{"latitude": 400000000, "longitude": -750000000},"hi":{"latitude": 410000000, "longitude": -749800000}}'
{
  "name": "100-122 Locktown Road, Frenchtown, NJ 08825, USA",
  "location": {
    "latitude": 405047245,
    "longitude": -749800722
  }
}

{
  "location": {
    "latitude": 405314270,
    "longitude": -749836354
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,7 @@ Changelog
 =========
 
 # Unreleased (0.21.0)
-* Fix gRPC server stream handling to be compatible with Java gRPC server by closing
-  the send stream post dispatching the client request.
+* Fix gRPC server stream handling to be compatible with Java gRPC server.
 
 # 0.20.0 (2021-05-18)
 * Add `stream-delay-close-send` option which delays client send stream closure.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@ Changelog
 =========
 
 # Unreleased (0.21.0)
-* Nothing yet
+* Fix gRPC server stream handling to be compatible with Java gRPC server by closing
+  the send stream post dispatching the client request.
 
 # 0.20.0 (2021-05-18)
 * Add `stream-delay-close-send` option which delays client send stream closure.
--- a/handler.go
+++ b/handler.go
@@ -155,12 +155,12 @@ func makeStreamRequest(t transport.Transport, streamReq *transport.StreamRequest
 	case encoding.ClientStream:
 		return makeClientStream(ctx, stream, streamIO, opts)
 	default:
-		return makeServerStream(ctx, stream, streamIO, opts)
+		return makeServerStream(ctx, stream, streamIO)
 	}
 }
 
 // makeServerStream starts server-side streaming rpc
-func makeServerStream(ctx context.Context, stream *yarpctransport.ClientStream, streamIO StreamIO, opts StreamRequestOptions) error {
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,8 @@ Changelog
 =========
 
 # Unreleased (0.21.0)
-* Nothing yet
+* Fix gRPC server stream handling to be compatible with Java gRPC server by closing
+  the send stream post dispatching the client request.
```
Dogild, line 6: nit: no need of mentioning `by closing the send stream post dispatching the client request.`

handler.go
```diff
@@ -180,6 +180,10 @@ func makeServerStream(ctx context.Context, stream *yarpctransport.ClientStream,
 		return err
 	}
 
+	if err = closeSendStream(ctx, stream, opts.DelayCloseSendStream.Duration()); err != nil {
```
Dogild, line 183: nit: `if err :=`

Dogild (commented):
Can you please add example in the summary that Golang gRPC is still working?
Also, it would be great to show that benchmarking is working with Java and golang as well

handler.go
prashantv: if the request is unary, dos a close-send delay make sense?

prashantv (commented):
> In server-side streaming requests, Java servers wait until the client sends the first request and closes the send stream before forwarding the request to the service handler. This behavior is different in Go gRPC servers, where it receives the request and immediately invokes the handler without waiting for the client's send stream to close.

Do you think it's worth filing an issue on gRPC to call out this behaviour difference?

handler.go
jronak: Dropped close-send delay, initially thought users could leverage it to notice request behavior through interceptors/middleware before reaching the handler.

jronak:
> Do you think it's worth filing an issue on gRPC to call out this behavior difference?

@prashantv, thanks for the idea. I have filed an issue on [grpc-go](https://github.com/grpc/grpc-go/issues/4608) to clarify server behavioral differences.

prashantv (approved):
Looks good to me, but get a review from @Dogild too.

Dogild (approved):

PR #347, opened 2021-07-23
Set up dependabot

abhinav:


prashantv:
Not opposed to this since releasing a new version is still manual, but I do wish there was some way to use releases older than X days, basically:
https://github.com/dependabot/dependabot-core/issues/2208

And context for why is: https://twitter.com/garybernhardt/status/1093326386465361920?s=21

abhinav:
Fair. I don't think we have that level of customization right now.

PR #351, opened 2022-03-03
Copy GenerateAndRegisterManualResolver from grpc.

zmt:
The function was removed from upstream grpc after v1.29.1. The
fucntionality still works without changes with newer grpc. Copying this
locally will facilitate consumers upgrading grpc.


zmt:
This is no longer relevant or necessary.

PR #353, opened 2022-04-03
Fix Go module importing

sdzyba:
Currently `github.com/yarpc/yab` cannot be imported as Go module dependency due to an error while trying to execute
`go get github.com/yarpc/yab`:

```
go get: github.com/yarpc/yab@v1.19.1 requires
	github.com/uber-go/atomic@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
```

In order to fix that the necessary to upgrade dependencies were upgraded, particularly:
- `go.uber.org/atomic`
- `github.com/opentracing/opentracing-go`
- `github.com/uber/tchannel-go`

And some related modules as well.

Also tried to keep `glide.yaml` declarations in sync with what declared in the `go.mod`.

There is also an issue with `github.com/streadway/quantile` (which is required as an indirect dependency) and Go 1.18.
The corresponding [merge request](https://github.com/streadway/quantile/pull/7) is not merged yet so I had to `replace` the issue in the `go.mod`.


abhinav:
@sdzyba Thanks for highlighting this, and the PR.
Looks like the project was outdated on several fronts.
We've updated the go.mod directive and a number of dependencies,
including all three you mentioned above.
Is this still an issue for you with the latest dev?

prashantv:
One thing I wanted to mention -- `yab` is not intended to be consumed as a library, but as a binary. This means:
 * The API may change at any time (no backwards compatibility guarantees)
 * All the subpackages could be moved into an internal package

sdzyba:
@abhinav
Yes, I can confirm the module can be imported with the current latest commit.
Thanks!

@prashantv 
Of course, I understand, but in it's current state it contains some very useful utility package.
Hope they remain non-internal. :)

PR #354, opened 2022-04-04
Build and test with Go 1.18

abhinav:
Build and test yab with Go 1.18 instead of 1.14 which is completely
unsupported at this point.

Steps taken to do this:

- Update GitHub workflow file
- Run `go mod tidy -go=1.18`
- Fix failing test (the error message change between 1.14 and 1.18)


go.mod
```diff
@@ -1,51 +1,68 @@
 module github.com/yarpc/yab
 
-go 1.14
+go 1.18
```
rabbbit, line 3: why are we doing this?

I thought you previously recommended we don't update unless necessary? 

I was reading about the unused module pruning in go 1.17 recently though - I was wondering if that means we can drop the `tools` sub-package.

abhinav: I might have been misinformed when I advised against bumping this.

I thought that for libraries, the `go` directive specifies the minimum required Go version but that's not true. A library with a 1.18 directive will still build on 1.17 as long as 1.18-based features are gated behind build tags.

For binaries, it matters even less.
So it makes sense to make use of the graph pruning introduced in 1.17.
(go.sum tries to remain compatible with N-1 version, so graph pruning doesn't clean it up until 1.18.)

PR #403, opened 2023-02-23
Output errors in JSON format

jakobgt:
So far, the errors encountered during a YAB benchmark run would be printed as text and not json, when format is json. This PR fixes that.

I'm rearranging the error printing a bit to align it with the other info printed. Lastly, I've added a test that verifies the output when errors are encountered (something that was not tested before).

### DIFF
--- a/benchmark.go
+++ b/benchmark.go
@@ -69,9 +69,9 @@ type Summary struct {
 
 // ErrorSummary stores the summary of the errors encountered
 type ErrorSummary struct {
-	TotalErrors int     `json:"totalErrors"`
-	ErrorRate   float64 `json:"errorRate"`
-	ErrorsCount map[string]int
+	TotalErrors int            `json:"totalErrors"`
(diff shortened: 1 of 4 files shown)


benchmark.go
```diff
@@ -66,6 +67,13 @@ type Summary struct {
 	RPS                float64 `json:"rps"`
 }
 
+// ErrorSummary stores the summary of the errors encountered
+type ErrorSummary struct {
+	TotalErrors int     `json:"totalErrors"`
+	ErrorRate   float64 `json:"errorRate"`
+	ErrorsCount map[string]int
```
jronak, line 74: nit: can you add the JSON tag here as well?

jronak (approved):
LGTM! - added a minor nit

jeppeuber (approved):

benchmark.go
jakobgt: Good point! Fixed.

PR #414, opened 2024-05-31
Add an option to force capture of Jaeger traces

jquirke:
Currently, the existing --jaeger flag adds trace, span and parent span information to outgoing RPCs.

However it does not set the sampling bit, which means analysing the trace is impossible if the backend infrastructure is not recording the distributed trace.

This new option --force-jaeger-sample ensures that every originating yab RPC is captured, most useful in the single yab (non benchmarking) case.

It uses constSampler with a decision value of true to ensure the sampling bit in the Jaeger headers. See:
https://www.jaegertracing.io/docs/1.22/client-libraries/#tracespan-identity

### DIFF
--- a/options.go
+++ b/options.go
@@ -93,7 +93,7 @@ type TransportOptions struct {
 	TransportHeaders    map[string]string `short:"T" long:"topt" description:"Transport options for TChannel, protocol headers for HTTP"`
 	HTTPMethod          string            `long:"http-method" description:"The HTTP method to use"`
 	GRPCMaxResponseSize int               `long:"grpc-max-response-size" description:"Maximum response size for gRPC requests. Default value is 4MB"`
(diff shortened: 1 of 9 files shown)


options.go
```diff
@@ -93,7 +93,7 @@ type TransportOptions struct {
 	TransportHeaders    map[string]string `short:"T" long:"topt" description:"Transport options for TChannel, protocol headers for HTTP"`
 	HTTPMethod          string            `long:"http-method" description:"The HTTP method to use"`
 	GRPCMaxResponseSize int               `long:"grpc-max-response-size" description:"Maximum response size for gRPC requests. Default value is 4MB"`
-
+	ForceJaegerSample   bool              `long:"force-jaeger-sample" description:"Force Jaeger to sample every request"`
```
jronak, line 96: nit: description does not match with the one seen in readme

jronak (approved):
lgtm, thanks for adding this option

PR #416, opened 2024-06-19
Fix bug with unicode handling in proto map keys

jquirke:
In a map where the key is a string, containing unicode characters, yab may fail to properly decode the protobuf message.
```
 // User Tags, key is tag name while value is the actual tag
    map<string, common.CustomerTag> tags = 11
```

This is because the JSON string written by jump/protoreflect's MarshalJSONPB method doesn't correctly escape unicode characters.

Inspecting the intermediate JSON we see invalid JSON:

```
 "AU_P\342\200\231NutAsianKitchen_2024feb05-T": { ... }
```

This issue was fixed in https://github.com/jhump/protoreflect/pull/481 in v1.10.2, however we upgrade the dependency to 1.15.6, the latest usable before a breaking change to protobuf and a requirement to use go1.19.

Before
```
$ yab .... | jq keys
Failed while parsing response: invalid character '3' in string escape code
Process 97171 has exited with status 1
```

After

```
$ yab .... | jq keys
[
  "body",
  "headers"
]
```


jronak (approved):
Let's wait for the internal repository to remove version clamp on protoreflect, otherwise LGTM. Thanks!

jquirke:
> Let's wait for the internal repository to remove version clamp on protoreflect, otherwise LGTM. Thanks!

Hey, what's the latest on this, I can't see Uber's internal repository removing the version clamp, and yab is a tool we publish OSS, so I forget why we are constrained by Uber's internal issues.

biosvs:
> so I forget why we are constrained by Uber's internal issues

We won't be able to apply any following fix internally. So let's make it in a right order: prepare internals first, then yab.

biosvs (changes requested):
(I'll leave a "Request changes", though it's just a sign that we're currently blocked with this PR.)

PR #417, opened 2024-09-24
remove resolver.UnregisterForTesting usage

bananacocodrilo:
`resolver.UnregisterForTesting` has been removed from grpc-go https://github.com/grpc/grpc-go/pull/6471, this change is needed in order to update to grpc 1.59.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+# 0.26.0 (2025-09-01)
+* Add http2 support to yab
+* Fix flaky unit test TestBenchmarkStateGetQuantileSuccess
+
 # 0.25.0 (2025-08-12)
 * Fix invalid reference to syscall.recvmsg
 * Optionally encode binary in thrift responses in the same way as requests
--- a/version.go
+++ b/version.go
@@ -22,4 +22,4 @@ package main
 
 // versionString is the sem-ver version string for yab.
 // It will be bumped explicitly on releases.
-var versionString = "0.25.0"
+var versionString = "0.26.0"


jronak (approved):

protobuf/source_reflection.go
```diff
@@ -32,8 +32,7 @@ type ReflectionArgs struct {
 // NewDescriptorProviderReflection returns a DescriptorProvider that reaches
 // out to a reflection server to access file descriptors.
 func NewDescriptorProviderReflection(args ReflectionArgs) (DescriptorProvider, error) {
-	r, deregisterScheme := GenerateAndRegisterManualResolver()
-	defer deregisterScheme()
+	r := GenerateAndRegisterManualResolver()
```
biosvs: We're creating potential memory leak here.
Let's discuss potential solution with a custom resolver, i.e. to register only one resolver, and then reuse it for different services.

biosvs (changes requested):

protobuf/source_reflection.go
bananacocodrilo: I've looked into two options to address it:
 
 ### Keep the manual resolver.

We can create the scheme deterministically (I'm thinking `scheme = "dest-" + args.Service`).
This allows us to use `resolver.Get(scheme)` to avoid creating unnecessary resolvers, and we also can keep an updated the list of peers in the resolver.
 
 ### Use passthrough

This results in simpler code, but only allows to specify one peer.
If the peer-list file used contains outdated addresses this could make the reflection fail where otherwise it would have worked.

protobuf/source_reflection.go
```diff
@@ -126,9 +124,21 @@ func wrapReflectionError(err error) error {
...
 
-func GenerateAndRegisterManualResolver() *manual.Resolver {
-	scheme := strconv.FormatInt(time.Now().UnixNano(), 36)
+func GetOrGenerateAndRegisterManualResolver(service string, peers []resolver.Address) *manual.Resolver {
+	scheme := "dest-" + service
+	newState := resolver.State{Addresses: peers}
+
+	rb := resolver.Get(scheme)
+	if rb != nil {
+		if r, ok := rb.(*manual.Resolver); ok {
```
bananacocodrilo: Manual resolvers implement both, Resolver and Builder.
r.InitialState(newState) will set the new list of peers the next time that Build is called

protobuf/source_reflection_test.go
```diff
@@ -196,6 +196,44 @@ func TestReflectionRoutingHeaders(t *testing.T) {
 	}
 }
 
+func TestResolverAlreadyExists(t *testing.T) {
```
bananacocodrilo: Test creating a new resolver for the same service with a different list of peers

protobuf/source_reflection.go
```diff
@@ -127,9 +124,21 @@ func wrapReflectionError(err error) error {
...
-func GenerateAndRegisterManualResolver() (*manual.Resolver, func()) {
-	scheme := strconv.FormatInt(time.Now().UnixNano(), 36)
+func GetOrGenerateAndRegisterManualResolver(service string, peers []resolver.Address) *manual.Resolver {
+	scheme := "dest-" + service
+	newState := resolver.State{Addresses: peers}
+
+	rb := resolver.Get(scheme)
+	if rb != nil {
+		if r, ok := rb.(*manual.Resolver); ok {
+			r.InitialState(newState)
```
biosvs: InitialState doesn't use mutex inside, so it's a potential data race.

PR #419, opened 2024-12-20
Fix: add default authority for reflection call

pulkit4tech:
Tested in Mono-repo - https://code.uberinternal.com/D16017325
Tested with librelease - https://librelease.uberinternal.com/diff_branch/librelease_oss-libraries_go-code_D16017325


pulkit4tech:
Probably, we can provide a request option flag to override this (if needed)

biosvs (approved):

PR #423, opened 2025-06-11
Fixes invalid reference to syscall.recvmsg

jakobgt:
I can't build the newest version of yab, due to the error `link: golang.org/x/net/internal/socket: invalid reference to syscall.recvmsg`, which is caused by an old version of `golang.org/x/net` (see error below). I've tried both on a Mac Arm machine and a Linux one.

In this PR, I've updated the `golang.org/x/net` to the newest version 0.41.0, and all tests passes. With that go.mod has also been updated to use `go 1.23.0`, and if that is a problem I can easily revert back to `golang.org/x/net` version 0.9.0, that also works. I opted to use version 0.41.0, as it is only 6 days old (while 0.9.0 is ~2 years old). See versions in https://go.googlesource.com/net/+refs

I was inspired by a [similar issue](https://github.com/chipmk/docker-mac-net-connect/issues/39#issuecomment-2334939751), that was also fixed by a version bump.

```
$ make test
go test -cover -timeout 30s -race ./...
# github.com/yarpc/yab/limiter.test
link: golang.org/x/net/internal/socket: invalid reference to syscall.recvmsg
# github.com/yarpc/yab/statsd.test
link: golang.org/x/net/internal/socket: invalid reference to syscall.recvmsg
# github.com/yarpc/yab/statsd/statsdtest.test
link: golang.org/x/net/internal/socket: invalid reference to syscall.recvmsg
# github.com/yarpc/yab/protobuf.test
link: golang.org/x/net/internal/socket: invalid reference to syscall.recvmsg
# github.com/yarpc/yab/encoding.test
link: golang.org/x/net/internal/socket: invalid reference to syscall.recvmsg
# github.com/yarpc/yab.test
link: golang.org/x/net/internal/socket: invalid reference to syscall.recvmsg
FAIL	github.com/yarpc/yab [build failed]
```

PR #424, opened 2025-06-11
Fix GH actions cache version

jakobgt:
It seems that the Github actions v1 is deprecated, so upgrading the GH action cache.


jakobgt:
Fixed the GH actions, so that https://github.com/yarpc/yab/pull/423 can be landed.

rabbbit (approved):

jakobgt:
Thanks!

PR #425, opened 2025-06-12
Upgrade x/net to v0.40

Groxx:
Addresses issues like
```
❯ make test
go test -cover -timeout 30s -race ./...
# github.com/yarpc/yab/limiter.test
link: golang.org/x/net/internal/socket: invalid reference to syscall.recvmsg
# github.com/yarpc/yab/encoding.test
link: golang.org/x/net/internal/socket: invalid reference to syscall.recvmsg
# github.com/yarpc/yab.test
link: golang.org/x/net/internal/socket: invalid reference to syscall.recvmsg
```
which have occurred since go 1.23.

---

The full set of steps to create this PR are:
- use go 1.24
- `go get golang.org/x/net@v0.40.0`
- `go mod tidy`

The resulting changes are more... opinionated... than I would like, but it's what Go tooling produces all on its own.  So I'm just leaving it in for discussion purposes.
The toolchain can, at the very least, be downgraded if desired.  For the rest, I'm not sure.


jakobgt:
Note, I'm in the process of merging https://github.com/yarpc/yab/pull/423, which should also fix this one.

Groxx:
Closing in favor of #423, no need to duplicate :)

PR #426, opened 2025-06-12
Optionally encode binary in thrift responses in the same way as requests

Groxx:
Yab allows requests with binary fields to be populated in two ways:
```json
{"arg":"a regular string"}
```
for stuff that's JSON-safe, probably for simplicity, and:
```json
{"arg":{"base64":"YSByZWd1bGFyIHN0cmluZwo="}}
```
for arbitrary bytes which cannot otherwise be encoded in JSON (or simply as an opt-in preference).

Unfortunately it currently encodes a response with ^ that exact same value echoed back as:
```json
{"arg":"YSByZWd1bGFyIHN0cmluZwo="}
```
which is inconsistent, doesn't round-trip (if you send this as a request, you'll receive that string, not the decoded bytes), and confuses users (we have a few cases where people have thought "I got base64 here" means "I should send base64 here" and didn't understand why their service didn't receive the data they encoded).

So that's now changed.

To avoid making `{"arg":"string"}` ambiguous, it's currently _always_ returning `{"arg":{"base64":"YSByZWd1bGFyIHN0cmluZwo="}}`.
This will break many things that depended on the output, but it's at least clear how to handle the new format, and doesn't risk being incorrectly base64-decoded.

Later maybe we can make it prefer the simple format for JSON-safe data, but I think that should probably be after a bit, to let people update their scripts away from inline-base64 to keep things unambiguous.


Groxx:
As a concrete example, this thrift spec:
```
service Echo {
    binary echo(1: binary data)
}
```

with this impl in yarpc:
```go
func (e echohandler) Echo(ctx context.Context, data []byte) ([]byte, error) {
	return data, nil
}
```

leads to this behavior:
```
❯ yab -p 127.0.0.1:8888 --service demo --method Echo::echo -r '{"data":"asdf"}' -t echo.thrift
{                                                                       ^^^^
  "body": {                                                             ||||
    "result": "YXNkZg=="                                                ||||
  },           ^^^^^^^^---------------------------------------------- different
  "headers": {
    "$rpc$-service": "demo",
    "load": "q=0"
  },
  "ok": true,
  "trace": "0"
}
```
which seems rather unambiguously bad / wrong.

pulkit4tech:
Hi @Groxx , overall change LGTM, but this will most likely break services using Yab to make call. I think we need to atleast test this in Go, Java & Python mono-repo before we can release this.

pulkit4tech (approved):

PR #430, opened 2025-08-18
Add http2 support to yab

mark200:
Yab now supports HTTP/2 for HTTP transport benchmarking, providing improved performance through:
- Multiplexing multiple requests over a single connection
- Header compression
- Server push capabilities
- Better flow control

The following command-line options are available for HTTP/2 configuration:

## --http2
- **Type**: Boolean
- **Default**: `true`
- **Description**: Enable HTTP/2 for HTTP transport (default is HTTP/1.1)

# Basic HTTP/2 Benchmarking
yab -s my-service -p localhost:8080 -m MyMethod -r '{"key": "value"}' --http2

### DIFF
--- a/options.go
+++ b/options.go
@@ -95,6 +95,10 @@ type TransportOptions struct {
 	HTTPMethod          string            `long:"http-method" description:"The HTTP method to use"`
 	GRPCMaxResponseSize int               `long:"grpc-max-response-size" description:"Maximum response size for gRPC requests. Default value is 4MB"`
 	ForceJaegerSample   bool              `long:"force-jaeger-sample" description:"Force all requests to be sampled for Jaeger tracing (use with --jaeger)"`
+
+	// Enables HTTP2 transport
+	UseHTTP2 bool `long:"http2" description:"Enable HTTP/2 for HTTP transport"`
+
 	// This is a hack to work around go-flags not allowing disabling flags:
 	// https://github.com/jessevdk/go-flags/issues/191
 	// Do not specify this value in a defaults.ini file as it is not possible
--- a/transport/http.go
+++ b/transport/http.go
@@ -22,13 +22,17 @@ package transport
 
 import (
 	"bytes"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"io/ioutil"
 	"math/rand"
+	"net"
 	"net/http"
 	"strconv"
 	"time"
+
+	"golang.org/x/net/http2"
 
 	"github.com/opentracing/opentracing-go"
 	"golang.org/x/net/context"
@@ -51,6 +55,9 @@ type HTTPOptions struct {
 	ShardKey        string
 	Encoding        string
 	Tracer          opentracing.Tracer
+
+	// HTTP/2 specific options
+	UseHTTP2 bool
 }
 
 var (
@@ -70,11 +77,25 @@ func NewHTTP(opts HTTPOptions) (Transport, error) {
 		opts.Method = "POST"
 	}
 
+	var transport http.RoundTripper
+
+	if opts.UseHTTP2 {
+		transport = &http2.Transport{
+			AllowHTTP: true,
+			DialTLSContext: func(ctx context.Context, network, addr string, _ *tls.Config) (net.Conn, error) {
+				var d net.Dialer
+				return d.DialContext(ctx, network, addr)
+			},
+		}
+	} else {
+		transport = &http.Transport{}
+	}
+
 	return &httpTransport{
 		opts: opts,
 		// Use independent HTTP clients for each transport.
 		client: &http.Client{
-			Transport: &http.Transport{},
+			Transport: transport,
 		},
 		tracer: opts.Tracer,
 	}, nil
--- a/transport/http_test.go
+++ b/transport/http_test.go
@@ -22,7 +22,6 @@ package transport
 
 import (
 	"io"
-	"io/ioutil"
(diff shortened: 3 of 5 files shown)


options.go
```diff
@@ -95,6 +95,10 @@ type TransportOptions struct {
 	HTTPMethod          string            `long:"http-method" description:"The HTTP method to use"`
 	GRPCMaxResponseSize int               `long:"grpc-max-response-size" description:"Maximum response size for gRPC requests. Default value is 4MB"`
 	ForceJaegerSample   bool              `long:"force-jaeger-sample" description:"Force all requests to be sampled for Jaeger tracing (use with --jaeger)"`
+
+	// HTTP/2 specific options
```
pulkit4tech: fix comment

transport/http.go
```diff
@@ -70,11 +77,25 @@ func NewHTTP(opts HTTPOptions) (Transport, error) {
 		opts.Method = "POST"
 	}
 
+	var transport http.RoundTripper
+
+	if opts.UseHTTP2 {
+		transport = &http2.Transport{
+			AllowHTTP: true,
+			DialTLSContext: func(ctx context.Context, network, addr string, cfg *tls.Config) (net.Conn, error) {
```
pulkit4tech:
```suggestion
			DialTLSContext: func(ctx context.Context, network, addr string, _ *tls.Config) (net.Conn, error) {
```

since you are not using tls config

transport/http.go
```diff
@@ -148,6 +169,10 @@ func (h *httpTransport) Call(ctx context.Context, r *Request) (*Response, error)
 		return nil, err
 	}
 
+	req.GetBody = func() (io.ReadCloser, error) {
```
pulkit4tech: why do we need this?

transport/http_test.go
```diff
@@ -169,6 +108,11 @@ func TestHTTPCall(t *testing.T) {
 		case "server_err":
 			w.WriteHeader(http.StatusInternalServerError)
 			return
+		case "timeout_hook":
+			deadline, _ := r.Context().Deadline()
+			newDeadline := deadline.Add(3 * time.Millisecond)
```
pulkit4tech: why 3ms?

transport/http_test.go
```diff
@@ -183,7 +127,16 @@ func TestHTTPCall(t *testing.T) {
 
 		w.Header().Set("Custom-Header", "ok")
 		io.WriteString(w, "ok")
-	}))
+	})
+
+	h2s := &http2.Server{
+		NewWriteScheduler: func() http2.WriteScheduler {
+			return http2.NewPriorityWriteScheduler(nil)
+		},
```
pulkit4tech: you might not need this, can you verify?

transport/http_test.go
```diff
@@ -254,3 +208,105 @@ func TestHTTPCall(t *testing.T) {
...
+	ctxOverride context.Context
+	hook        string
+	method      string
+	errMsg      string
+	ttlMin      time.Duration
+	ttlMax      time.Duration
+	wantCode    int
+	wantBody    []byte // If nil, uses the request body
+
+	useHTTP2 bool
```
pulkit4tech: let's not pass this flag here, instead let's have 2 test function - 

- TestHTTPCall(...)
- TestHTTP2Call(...)

both should get same set of test cases but http2 one will just pass flag when making call

pulkit4tech (changes requested):

transport/http.go
pulkit4tech: https://github.com/golang/go/issues/69412

pulkit4tech (approved):

options.go
```diff
@@ -95,6 +95,10 @@ type TransportOptions struct {
 	HTTPMethod          string            `long:"http-method" description:"The HTTP method to use"`
 	GRPCMaxResponseSize int               `long:"grpc-max-response-size" description:"Maximum response size for gRPC requests. Default value is 4MB"`
 	ForceJaegerSample   bool              `long:"force-jaeger-sample" description:"Force all requests to be sampled for Jaeger tracing (use with --jaeger)"`
+
+	// Enables HTTP2 transport
```
pulkit4tech, line 99: update comment that this cli option only works for thrift case at the moment

main.go
```diff
@@ -432,6 +432,9 @@ func resolveProtocolEncoding(protocolScheme string, rOpts RequestOptions) resolv
 	// Try to determine a transport based on the guessed encoding.
 	switch enc {
 	case encoding.Thrift:
+		if opts.TOpts.UseHTTP2 {
```
pulkit4tech: move this condition check to above protocolScheme switch case as well as we want to cover cases where cli tools provide `peer http://{host}:port`

transport/http_test.go
```diff
@@ -183,7 +186,179 @@ func TestHTTPCall(t *testing.T) {
...
+		})
+	}
+}
+
+func TestHTTP2Call(t *testing.T) {
+	timeoutCtx, _ := context.WithTimeout(context.Background(), 3*time.Second)
+
+	immediateTimeout, _ := context.WithTimeout(context.Background(), clientTimeout)
+
+	tests := []TestBody{
```
pulkit4tech: let's use common http1 test cases and if needed append additional here

pulkit4tech (changes requested):

pulkit4tech (approved):

PR #436, opened 2026-02-11
Add support to set the value of rpc-encoding header

bananacocodrilo:
Yarpc supports custom codecs defined by the user. It uses the rpc-encoding header to decide which codec to use.

This flag allows developers to overwrite the content of the rpc-encoding header so the destination service selects the desired codec.

### DIFF
--- a/options.go
+++ b/options.go
@@ -90,6 +90,7 @@ type TransportOptions struct {
 	RoutingKey          string            `long:"rk" description:"The routing key overrides the service name traffic group for proxies."`
 	RoutingDelegate     string            `long:"rd" description:"The routing delegate overrides the routing key traffic group for proxies."`
 	ShardKey            string            `long:"sk" description:"The shard key is a transport header that clues where to send a request within a clustered traffic group."`
+	RPCEncoding         string            `long:"rpc-encoding" description:"Override the rpc-encoding header/metadata value for gRPC and HTTP transports. This does not re-encode the request body and is intended for development."`
 	Jaeger              bool              `long:"jaeger" description:"Use the Jaeger tracing client to send Uber style traces and baggage headers"`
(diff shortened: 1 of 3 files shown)


rpc_encoding_flag_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	_, err = tp.Call(context.Background(), &yabtransport.Request{TargetService: "svc", Method: "svc::echo", Body: []byte("hello")})
+	require.NoError(t, err)
+
+	select {
+	case got := <-router.capturedEncoding:
+		assert.Equal(t, want, got)
+	default:
+		t.Fatal("did not capture inbound encoding")
+	}
```
mark200, line 87: Can you maybe add a test for tchannel? Just to stay consistent with all transports.

bananacocodrilo: Removed it from tChannel altogether. 
tChannel uses the encoding internally for the `format` field that controls how the payload is serialized.

mark200 (approved):
LGTM

options.go
```diff
@@ -90,6 +90,7 @@ type TransportOptions struct {
 	RoutingKey          string            `long:"rk" description:"The routing key overrides the service name traffic group for proxies."`
 	RoutingDelegate     string            `long:"rd" description:"The routing delegate overrides the routing key traffic group for proxies."`
 	ShardKey            string            `long:"sk" description:"The shard key is a transport header that clues where to send a request within a clustered traffic group."`
+	RPCEncoding         string            `long:"rpc-encoding" description:"Override the rpc-encoding header/metadata value for gRPC and HTTP transports. This does not re-encode the request body and is intended for development."`
```
pulkit4tech: nit: can we add short option for this as well

pulkit4tech (approved):

PR #439, opened 2026-07-02
Make concurrency parameter more fair between transports

nshipilov:
We are transitioning to Envoy from a golang based forwarding proxy. We were loadtesting with yab and getting unexpected results when testing cases with O(worker thread) connections + high concurrency per worker thread.

golang's green threads allow the requests to be forwarded on any available green thread, meanwhile on envoy, the requests are handled on the thread that has the connection. 

In turn, the fact that the requests per transport are not distributed fairly, shows up as high latency on envoy.

What does the unfairness come from?

We start the runWorker goroutines in the order of [TP: 0, Conc: 1], [TP: 0, Conc: 2], ... [TP: 0, Conc: M], ... [TP: N, Conc: 0], ... [TP: N, Conc: M]

They all block on the timePeriod ::take() mutex: https://github.com/yarpc/yab/blob/dev/ratelimit/time_period.go#L61-L63, which, as I understand is likely fair in the case of large amounts of contention.

Thus, the requests largely execute in the order they goroutines hit the take() mutex, and re-queue themselves up again in the same order (basically a "circular buffer" like strategy).

This changes it so that the runWorker goroutines are instead started in the order of [TP: 0, Conc: 1], [TP: 1, Conc: 1], ... [TP: N, Conc: 1], ... [TP: 0, Conc: M], [TP: N, Conc: M], such that the requests are distributed in more of a round-robin fashion between the transports as opposed to the above.

This was tested and verified in an ad-hoc manner, against our envoy deployment:

Previous yab behavior:

```
[redacted]@[redacted]:/var/tmp$ yab -p grpc://localhost:6435 --health -s [redacted] --connections=12 --concurrency=250 --rps=5000 -d 5s --warmup=100
{
  "body": {
    "status": "SERVING"
  },
  "headers": {
    "load": "q=0"
  }
}

Benchmark parameters:
  CPUs:            48
  Connections:     12
  Concurrency:     250
  Max requests:    75000
  Max duration:    15s
  Max RPS:         5000
Latencies:
  0.5000: 2.861853ms
  0.9000: 8.080991ms
  0.9500: 10.963272ms
  0.9900: 18.975153ms
  0.9990: 30.026031ms
  0.9995: 32.678249ms
  1.0000: 40.899147ms
Elapsed time (seconds):         5.00
Total requests:                 74996
RPS:                            4997.64
``` 

vs:

```
[redacted]@[redacted]:/var/tmp$ ./test-yab -p grpc://localhost:6435 --health -s [redacted] --connections=12 --concurrency=250 --rps=5000 -d 5s --warmup=100
{
  "body": {
    "status": "SERVING"
  },
  "headers": {
    "load": "q=0"
  }
}

Benchmark parameters:
  CPUs:            48
  Connections:     12
  Concurrency:     250
  Max requests:    25000
  Max duration:    5s
  Max RPS:         5000
Latencies:
  0.5000: 1.528597ms
  0.9000: 2.341612ms
  0.9500: 2.671046ms
  0.9900: 4.918146ms
  0.9990: 6.633406ms
  0.9995: 7.462587ms
  1.0000: 21.999739ms
Elapsed time (seconds):         5.00
Total requests:                 25000
RPS:                            4996.71
```

### DIFF
--- a/benchmark.go
+++ b/benchmark.go
@@ -229,8 +229,8 @@ func runBenchmark(out output, logger *zap.Logger, allOpts Options, resolved reso
 
 	logger.Info("Benchmark starting.", zap.Any("options", opts))
 	start := time.Now()
-	for i, c := range connections {
-		for j := 0; j < opts.Concurrency; j++ {
+	for j := 0; j < opts.Concurrency; j++ {
+		for i, c := range connections {
 			state := states[i*opts.Concurrency+j]
 
 			wg.Add(1)


DheerendraRathor (approved):

cinchurge (approved):
w00t!

bananacocodrilo (approved):

